PR #5, opened 2016-05-31
DaemonEvent unit test: use temp file location if testing, ignore non-…

HenryRawas:
3 fixes for DaemonEvent unit tests
use temp file location if testing
ignore non-container objects in response
handle read socket timeout (returns -1)

@jeffaco

### DIFF
--- a/source/code/providers/Container_DaemonEvent_Class_Provider.cpp
+++ b/source/code/providers/Container_DaemonEvent_Class_Provider.cpp
@@ -18,6 +18,7 @@
 #include "Container_ContainerLogFileReader.h"
 
 #define LASTQUERYTIMEFILE "/var/opt/microsoft/docker-cimprov/state/LastEventQueryTime.txt"
+#define TEST_LASTQUERYTIMEFILE "./LastEventQueryTime.txt"
 
 using namespace std;
 
@@ -43,6 +44,20 @@ class EventQuery
 {
 private:
 	///
+	/// Utility to get file path for LastEventQueryTime
+	/// 
+	static char* GetEventQueryTimeFilePath()
+	{
+		char* lastQueryFile = LASTQUERYTIMEFILE;
+		const char *cTestRun = getenv("CONTAINER_TESTRUN_ACTIVE");
+		if (cTestRun != NULL)
+		{
+			lastQueryFile = TEST_LASTQUERYTIMEFILE;
+		}
+		return lastQueryFile;
+	}
+
+	///
 	/// Get the previous time if it was stored, otherwise get the current time
 	///
 	/// \returns Time of previous query if available
@@ -51,7 +66,8 @@ private:
 	{
 		int fileTime = time(NULL);
 		int currentTime = fileTime;
-		FILE* file = fopen(LASTQUERYTIMEFILE, "r");
+		char* lastQueryFile = GetEventQueryTimeFilePath();
+		FILE* file = fopen(lastQueryFile, "r");
 
 		if (file)
 		{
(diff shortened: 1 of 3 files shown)


source/code/providers/Container_DaemonEvent_Class_Provider.cpp
```diff
@@ -18,6 +18,7 @@
 #include "Container_ContainerLogFileReader.h"
 
 #define LASTQUERYTIMEFILE "/var/opt/microsoft/docker-cimprov/state/LastEventQueryTime.txt"
+#define TEST_LASTQUERYTIMEFILE "/tmp/LastEventQueryTime.txt"
```
jeffaco: So this is a problem if multiple people work on the unit tests. Why not make this something like ./LastEventQueryTime.txt, which would allow two people (working on the same machine) to not clobber each other, since they'd be in different directory trees?

source/code/providers/Container_DaemonEvent_Class_Provider.cpp
```diff
@@ -79,7 +86,13 @@ class EventQuery
...
 	static void SetPreviousTime(int t)
 	{
-		FILE* file = fopen(LASTQUERYTIMEFILE, "w");
+		char* lastQueryFile = LASTQUERYTIMEFILE;
+		const char *cTestRun = getenv("CONTAINER_TESTRUN_ACTIVE");
+		if (cTestRun != NULL)
+		{
+			lastQueryFile = TEST_LASTQUERYTIMEFILE;
+		}
+		FILE* file = fopen(lastQueryFile, "w");
```
jeffaco: Code duplication. Consider creating a function that returns a FILE* (and, I guess, a lastQueryFile pointer by parameter), and then calling that here and above. Less work if we need to change something.

source/code/providers/Container_DaemonEvent_Class_Provider.cpp
```diff
@@ -179,7 +192,7 @@ class EventQuery
 				{
 					cJSON* entry = cJSON_GetArrayItem(response[0], i);
 
-					if (entry)
+					if (entry && !strcmp(cJSON_GetObjectItem(entry, "type")->valuestring, "container"))
```
jeffaco: A comment please, on what's coming back and why a change was needed, thanks.

jeffaco:
Your commit message is too long, keep it under 70 bytes or so. If it needs to be longer, then make it multiple lines, like this:

```
Resolve DaemonEvent unit test problems

Changes include:
  . use temp file location if testing,
  . ignore non-container objects,
  . handle read socket timeout.
```

Assuming this really resolves the test, you have sign-off with the changes from my comments. In particular, I want a meaningful comment from the strcmp change. Clearly, before, we were only getting containers before. What changed?

source/code/providers/Container_DaemonEvent_Class_Provider.cpp
HenryRawas: I can do that, but I saw other places where /tmp/ was used as a folder location for the unit tests. I decided to follow their example.

________________________________

HenryRawas:
I don't know if this test code ever ran before, since the other bug prevented the test from getting this far.

When did it start to fail?


________________________________

source/code/providers/Container_DaemonEvent_Class_Provider.cpp
jeffaco: Perhaps so. But I wouldn't have done this, and I think it's a poor example. What else did you see that did that? Was it in the docker provider or elsewhere?

Just because you see some yucky code doesn't mean it should be duplicated! :smile:

jeffaco:
I had failures as soon as I ported to GitHub, but a number of things changed (port to Git, tried on a different system, etc). I didn't personally run the tests earlier, and it could be that they never ran (I can't say for sure since I never personally ran them, and since the runs were not automated at this point).

By the way, it's better to respond via the GitHub WWW interface, so there's not so much crud attached to the comments here.

jeffaco:
Is this ready to be merged? Please let me know, thanks!

HenryRawas:
Yes it is ready to be merged.

Regarding why the test may have worked at one time but then failed later, the unit tests do not use mock objects, but actual docker instances. So the host environment affects the test. It is possible that the person who wrote the tests did not see these failures on their environment.

jeffaco:
Ultimately, the test needs to set up the environment properly so the test will reliably pass, assuming no changes are made to docker while the test is running. That's part of the point of this test.

Thanks. I'll merge this in for now.

jeffaco:
Merged to master:

````
commit 16f3406 (HEAD, origin/master, origin/HEAD, master)
Author: Jeff Coffler <jeff@taltos.com>
Date:   Wed Jun 1 14:49:31 2016

    Resolve DaemonEvent unit test problems
    
    Changes include:
      . use temp file location if testing,
      . ignore non-container objects,
      . handle read socket timeout.
    
    These changes came from contractor; I'm just doing the merge to master.
```

PR #8, opened 2016-06-03
Unittestimgenum

HenryRawas:
Fix for unit test image inventory enum. Ignore extra images if repo:tag is <none>:<none>

@jeffaco

### DIFF
--- a/test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
+++ b/test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
@@ -92,31 +92,61 @@ protected:
 		// Enumerate provider
 		StandardTestEnumerateInstances<mi::Container_ImageInventory_Class_Provider>(m_keyNames, context, CALL_LOCATION(errMsg));
 
-		// Get images using command line
-		CPPUNIT_ASSERT(!system("docker images -q --no-trunc > /tmp/docker_image_ids.txt"));
+		// Get images info using command line
+		wstring dockerimages = StrFromMultibyte(RunCommand("docker images --no-trunc"));
+		// split into lines
+		vector<wstring> lines;
+		StrTokenize(dockerimages, lines, L"\n");
 
-		FILE* idFile = fopen("/tmp/docker_image_ids.txt", "r");
-		CPPUNIT_ASSERT(idFile);
-
-		wchar_t id[65];
 		set<wstring> allIds;
 
-		// Full image IDs (one per line)
-		while (fwscanf(idFile, L"%ls", id) != EOF)
+		int matched = 0;
+
+		// Full image info (header plus one per line)
+		if (lines.size() > 0)
 		{
-			allIds.insert(wstring(id));
+			// parse header to get column offsets
+			vector<wstring>::iterator it = lines.begin();
+			string hdrstr = StrToMultibyte(*it, true);
+			it++;
+
+			string::size_type repoix = hdrstr.find("REPOSITORY", 0);
+			CPPUNIT_ASSERT_MESSAGE(string("Can't find 'REPOSITORY' keyword in docker output line: ").append(hdrstr), repoix != string::npos);
+			string::size_type tagix = hdrstr.find("TAG", 0);
+			CPPUNIT_ASSERT_MESSAGE(string("Can't find 'TAG' keyword in docker output line: ").append(hdrstr), tagix != string::npos);
+			string::size_type idix = hdrstr.find("IMAGE ID", 0);
+			CPPUNIT_ASSERT_MESSAGE(string("Can't find 'IMAGE ID' keyword in docker output line: ").append(hdrstr), idix != string::npos);
+			string::size_type createdix = hdrstr.find("CREATED", 0);
+			CPPUNIT_ASSERT_MESSAGE(string("Can't find 'CREATED' keyword in docker output line: ").append(hdrstr), createdix != string::npos);
+
+			// loop over lines after the header
+			for (; it != lines.end(); ++it)
+			{
+				wstring id = StrTrimR((*it).substr(idix, createdix - idix));
+				bool match = false;
+				for (unsigned i = 0; i < context.Size(); ++i)
+				{
+					// Verify the InstanceID
+					if (id.compare(context[i].GetKey(L"InstanceID", CALL_LOCATION(errMsg))) == 0)
+					{
+						match = true;
+						matched++;
+						break;
+					}
+				}
+
+				if (!match)
+				{
+					// not found. If the repo & tag are both "<none>", then this is ok since enumerate may filter these
+					wstring repo = StrTrimR((*it).substr(repoix, tagix - repoix));
+					CPPUNIT_ASSERT_EQUAL(wstring(L"<none>"), repo);
+					wstring tag = StrTrimR((*it).substr(tagix, idix - tagix));
+					CPPUNIT_ASSERT_EQUAL(wstring(L"<none>"), tag);
+				}
+			}
 		}
 
-		fclose(idFile);
-		remove("/tmp/docker_image_ids.txt");
-
-		CPPUNIT_ASSERT_EQUAL(allIds.size(), context.Size());
-
-		for (unsigned i = 0; i < context.Size(); ++i)
-		{
-			// Verify the InstanceID
-			CPPUNIT_ASSERT(allIds.count(context[i].GetKey(L"InstanceID", CALL_LOCATION(errMsg))));
-		}
+		CPPUNIT_ASSERT_EQUAL(matched, context.Size());
 	}
 
 	void TestEnumerateVerifyAllValues()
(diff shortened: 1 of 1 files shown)


test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -53,6 +53,18 @@ class ImageInventoryTest : public CppUnit::TestFixture
...
+	{
+		wstring sub = line.substr(begin, end - begin);
+		string::size_type trim = sub.find_last_not_of(L" \t");
+		if (trim != string::npos)
+		{
+			return sub.substr(0, trim + 1);
+		}
+		return sub;
+	}
+
```
jeffaco: There's a stringutils thing for that, see SCXCoreLIb::StrTrim in file `pal/source/code/include/scxcorelib/stringaid.h`. To avoid code duplication needlessly, considering that this project already requires the PAL, I'm going to ask you to make that change before accepting this.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -93,30 +105,62 @@ class ImageInventoryTest : public CppUnit::TestFixture
 		StandardTestEnumerateInstances<mi::Container_ImageInventory_Class_Provider>(m_keyNames, context, CALL_LOCATION(errMsg));
 
 		// Get images using command line
-		CPPUNIT_ASSERT(!system("docker images -q --no-trunc > /tmp/docker_image_ids.txt"));
+		CPPUNIT_ASSERT(!system("docker images --no-trunc > /tmp/docker_image_ids.txt"));
```
jeffaco: Blech. Is this what all the test code is doing? Nothing is using SCXProcess::Run() from the PAL, which gives you much greater control about looking at stdout/stderr?

I noted that providers/Container_DaemonEvent_Class_Provider_UnitTest.cpp is using SCXProcess::Run. Please do that here, it lets you look at stderr and make sure that nothing came in on that channel.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -93,30 +105,62 @@ class ImageInventoryTest : public CppUnit::TestFixture
...
+		int matched = 0;
+
+		// Full image info (header plus one per line)
+		if (fgetws(line, 200, idFile) != NULL)
 		{
-			allIds.insert(wstring(id));
+			// parse header to get column offsets
+			wstring hdrstr(line);
+			string::size_type repoix = hdrstr.find(L"REPOSITORY", 0);
+			CPPUNIT_ASSERT(repoix != wstring::npos);
```
jeffaco: This type of ASSERT gives very little information if it fires. Can you at least use CPPUNIT_ASSERT_MESSAGE to get a meaningful message of the problem if the unit test fails? Same for the lines below ...

On failure, you should print the line itself plus the fact that you couldn't find "REPOSITORY" within the line. Perhaps something like:

Can't find "REPOSITORY" keyword in docker output line: <docker output line>

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -93,30 +105,62 @@ class ImageInventoryTest : public CppUnit::TestFixture
...
+						break;
+					}
+				}
+
+				if (!match)
+				{
+					// not found. If the repo & tag are both "<none>", then this is ok since enumerate may filter these
+					wstring none(L"<none>");
+					wstring repo = substrTrim(linestr, repoix, tagix);
+					CPPUNIT_ASSERT_EQUAL(none, repo);
```
jeffaco: Why create "repo" here, why not do something like:

`CPPUNIT_ASSERT_EQUAL(wstring(L"<none>"), SCXCore::StrTrim(...))`

Same with tag below.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -93,30 +105,62 @@ class ImageInventoryTest : public CppUnit::TestFixture
 		StandardTestEnumerateInstances<mi::Container_ImageInventory_Class_Provider>(m_keyNames, context, CALL_LOCATION(errMsg));
 
 		// Get images using command line
-		CPPUNIT_ASSERT(!system("docker images -q --no-trunc > /tmp/docker_image_ids.txt"));
+		CPPUNIT_ASSERT(!system("docker images --no-trunc > /tmp/docker_image_ids.txt"));
 
 		FILE* idFile = fopen("/tmp/docker_image_ids.txt", "r");
```
jeffaco: Why not create an actual temporary file (with a temporary name) in the current directory, so other users on the same machine won't stomp on this? And why not use our existing framework to auto-delete the file (even if the test aborts early)?

For auto-delete of the file, see SelfDeletingFilePath in PAL: providers/Container_DaemonEvent_Class_Provider_UnitTest.cpp.

For creation of temp files, search for CreateTempFile in pal/test. There's a bunch of occurrences there.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -93,30 +105,62 @@ class ImageInventoryTest : public CppUnit::TestFixture
...
-		wchar_t id[65];
+		wchar_t line[200];
 		set<wstring> allIds;
 
-		// Full image IDs (one per line)
-		while (fwscanf(idFile, L"%ls", id) != EOF)
+		int matched = 0;
+
+		// Full image info (header plus one per line)
+		if (fgetws(line, 200, idFile) != NULL)
```
jeffaco: 200 is a magic number, you know what I want from prior ode reviews. Better to use C++ String class, at which point this doesn't matter at all. We have infrastructure to make that easy.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -93,30 +105,62 @@ class ImageInventoryTest : public CppUnit::TestFixture
...
+					wstring repo = substrTrim(linestr, repoix, tagix);
+					CPPUNIT_ASSERT_EQUAL(none, repo);
+					wstring tag = substrTrim(linestr, tagix, idix);
+					CPPUNIT_ASSERT_EQUAL(none, tag);
+				}
+			}
 		}
 
 		fclose(idFile);
 		remove("/tmp/docker_image_ids.txt");
```
jeffaco: This isn't the way to do that, make this an auto-deleting file that happens regardless of test success. See comments above.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -162,8 +206,8 @@ class ImageInventoryTest : public CppUnit::TestFixture
 		for (unsigned i = 0; i < images.size(); i++)
 		{
 			bool flag = false;
-			int rc = mbstowcs(currentId, cJSON_GetObjectItem(images[i], "InstanceID")->valuestring, (sizeof(currentId)/sizeof(currentId[0])) - 1);
-			if (rc > 0 && rc < sizeof(currentId)/sizeof(currentId[0])) currentId[rc] = 0;
+			int rc = (int)mbstowcs(currentId, cJSON_GetObjectItem(images[i], "InstanceID")->valuestring, (sizeof(currentId)/sizeof(currentId[0])) - 1);
```
jeffaco: Why not just type rc properly so you don't need to cast?

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
```diff
@@ -162,8 +206,8 @@ class ImageInventoryTest : public CppUnit::TestFixture
...
+			if (rc > 0 && rc < (int)(sizeof(currentId)/sizeof(currentId[0]))) currentId[rc] = 0;
```
jeffaco: I imagine typing rc properly would avoid this cast too?

jeffaco:
Please see my comments above. Resubmit when ready, thanks.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
HenryRawas: It returns an unsigned value, but uses a negative value to indicate an error. I cast to int so that the rc > 0 is meaningful.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
HenryRawas: mainly to avoid duplication, as there are 2 tests for "<none>"

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
jeffaco: But it probably doesn't use some "random negative value", it probably uses some specific value (like 0xffffffff) or something. Up, it says it returns a (size_t) -1, test for that, and type RC properly, thanks.

test/code/providers/Container_ImageInventory_Class_Provider_UnitTest.cpp
jeffaco: But this also makes the CPPUNIT_ASSERT_EQUAL less meaningful. I'd rather have one duplicated string (which, by the way, would be optimized away) and make the CPPUNIT_ASSERT_EQUAL test output more meaningful.

HenryRawas:
Resubmitted

jeffaco:
Merged to master branch. All tests appear to pass successfully now:

```
---- All tests run ----


OK (9 tests)
```

PR #9, opened 2016-06-08
Eventtimeinit

HenryRawas:
Unit test for Daemon event enum needs to init previous time to before now if file not exist.

@jeffaco


jeffaco:
Merged to master branch

PR #10, opened 2016-07-29
Status for paused container is being reported as ‘Running’.

amitsara:
@Microsoft/omsagent-devs @keikhara 

OMI provider is constructing status information from ‘Running’ & ‘Paused’ flag, both of these flags are true when container is paused.

Fixed the OMI provider for getting this information directly from State->Status node.


keikhara:
I checked the code. I think it looks good.

PR #13, opened 2016-08-08
fix image paused container count

HenryRawas:
adjust container counts in image inventory for paused
@amitsara


amitsara:
Looks good.

PR #17, opened 2016-10-11
Support json-file log driver in docker-provider

amitsara:
@Microsoft/omsagent-devs 
@keikhara 

Docker-provider will collect container-logs if the log driver is set to fluentd/json-file.

### DIFF
--- a/source/code/dockerapi/DockerRemoteApi.cpp
+++ b/source/code/dockerapi/DockerRemoteApi.cpp
@@ -217,6 +217,153 @@ vector<cJSON*> getResponse(vector<string>& request, bool isMultiJson, bool ignor
     return response;
 }
 
+void readLine(char* str, int length, int& readPtr, string& line)
+{
+    while (str && (readPtr < length))
+    {
+        if ((readPtr + 1) < length &&
+            str[readPtr] == '\r' && str[readPtr + 1] == '\n')
+        {
+            readPtr = readPtr + 2;
+            break;
+        }
+        line.append(string(1, str[readPtr]));
+        readPtr++;
+    }
+}
+
+void parseLogs(char* str, int length, vector<string>& logs)
+{
+    int readPtr = 0;
+   
+    if(!str)
+        return;
+
+    // skip header
+    while (readPtr < length)
+    {
+        if ((readPtr + 3) < length &&
+            str[readPtr] == '\r' && str[readPtr + 1] == '\n' && str[readPtr + 2] == '\r' && str[readPtr + 3] == '\n')
+        {
+            readPtr = readPtr + 4;
+            break;
+        }
+        readPtr++;
+    }
+
+    // read the logs
+    while (readPtr < length)
+    {
+        // skip data length
+        string dataLength;
+        readLine(str, length, readPtr, dataLength);
+
+        if (readPtr + 8 > length)
+        {
+            return;
+        }
+
+        // stream type
+        string parsedLog;
+        if(str[readPtr] == 1)
+        {
+            parsedLog = "stdout;";
+        }
+        else
+        {
+            parsedLog = "stderr;";
+        }
+        readPtr = readPtr + 8;
+
+        string message;
+        readLine(str, length, readPtr, message);
+        parsedLog.append(message);
+        logs.push_back(parsedLog);
+    }
+}
+
+vector<string> getContainerLogs(string& request)
+{
+    vector<string> response;
+
+    try
+    {
+        const int bufferSize = 4096;
+        const int timeoutSecond = 2;
+        vector<int> sockfd;
+        int n = 1;
+        createConnection(n, sockfd);
+
+        size_t r = write(sockfd[0], request.c_str(), request.length());
+        if (r != request.length())
+        {
+            throw string("Write to socket in getContainerLogs failed");
+        }
+
+        char readBuf[bufferSize + 1];
+        int read_n = 0;
+        char* raw_response = 0;
+        int raw_reponse_size = 0;
+
+        while ((read_n = readSocket(sockfd[0], readBuf, bufferSize, timeoutSecond)) > 0)
+        {
+            char* tempBuf = new char[raw_reponse_size + read_n];
+            if(raw_reponse_size > 0)
+            {
+                memcpy(tempBuf, raw_response, raw_reponse_size);
+                delete raw_response;
+                raw_response = 0;
+            }
+
+            memcpy(tempBuf + raw_reponse_size, readBuf, read_n);
+            raw_response = tempBuf;
+            raw_reponse_size = raw_reponse_size + read_n;
+        }
+
+        if(raw_reponse_size != 0 && raw_response)
+        {
+            parseLogs(raw_response, raw_reponse_size, response);
+            delete raw_response;
+        }
+
+        close(sockfd[0]);
+    }
+    catch (string str)
+    {
+        openlog("DockerRemoteApi", LOG_PID | LOG_NDELAY, LOG_LOCAL1);
+        syslog(LOG_ERR, "%s", str.c_str());
+    }
+
+    return response;
+}
+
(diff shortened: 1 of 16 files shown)


source/code/dockerapi/DockerRemoteApi.cpp
```diff
@@ -217,6 +217,153 @@ vector<cJSON*> getResponse(vector<string>& request, bool isMultiJson, bool ignor
...
+    catch (string str)
+    {
+        openlog("DockerRemoteApi", LOG_PID | LOG_NDELAY, LOG_LOCAL1);
+        syslog(LOG_ERR, "%s", str.c_str());
+    }
+
+    return response;
+}
+
+string getDockerHostName()
```
jeffaco: So you replaced gethostname() (a local Linux call) with a call to the docker API, which is considerably more expensive.

Was gethostname() failing in some way that made it unacceptable?

source/code/mof/docker.mof
```diff
@@ -157,4 +157,36 @@ class Container_ContainerInventory : CIM_ManagedElement
...
-};
\ No newline at end of file
+};
+
+// Container_ContainerLog
+// -------------------------------------------------------------------
+[   Version ( "1.0.0" ),
+	Description ( "Docker Container Logs Information" )
+]
+class Container_ContainerLog : CIM_ManagedElement
```
jeffaco: Generally, new classes have to be "blessed" from PM to make sure they abide by the CIM guidelines. Was this done? If @agup006 isn't comfortable with that, you might want to reach out to @KrisBash to be sure that this is "right".

source/code/providers/Container_ContainerInventory_Class_Provider.cpp
```diff
@@ -312,14 +312,7 @@ class ContainerQuery
...
         // Get computer name
-        char name[256];
-        string hostname = gethostname(name, 256) ? "Unknown" : string(name);
-        // in case get full name, extract up to '.'
-        size_t dotpos = hostname.find('.');
-        if (dotpos != string::npos)
-        {
-            hostname = hostname.substr(0, dotpos);
-        }
+        string hostname = getDockerHostName();
```
jeffaco: As mentioned above: Really needed? getDockerHostName is a lot more expensive than a local gethostname() call ...

source/code/providers/Container_ContainerLog.h
```diff
@@ -0,0 +1,898 @@
+/* @migen@ */
```
jeffaco: I presume this was completely generated from the MOF and not otherwise modified?

source/code/providers/Container_ContainerLog_Class_Provider.cpp
```diff
@@ -0,0 +1,288 @@
...
+#include <time.h>
+#include <unistd.h>
+#include <uuid/uuid.h>
+#include <vector>
+
+#include "../cjson/cJSON.h"
+#include "../dockerapi/DockerRemoteApi.h"
+#include "../dockerapi/DockerRestHelper.h"
+
+#define LASTLOGQUERYTIMEFILE "/var/opt/microsoft/docker-cimprov/state/LastLogQueryTime.txt"
```
jeffaco: Blech, a new one of these (we had another state file for time). I wonder if you should write a general state class (like SCX has for the logfilereader) that stores state in a proper way and manages what to do if said state information doesn't exist?

I thought the other time file was a bit of a cludge, having two seems twice as bad. It can stay, I guess, but you should think about doing something more formal.

source/code/providers/Container_ContainerLog_Class_Provider.cpp
```diff
@@ -0,0 +1,288 @@
...
+            fclose(file);
+
+            if (fileTime > currentTime)
+            {
+                syslog(LOG_WARNING, "The time stored in %s is more recent than the current time", lastQueryFile);
+            }
+        }
+        else
+        {
+            syslog(LOG_ERR, "Attempt in GetPreviousTime to open %s for reading failed", lastQueryFile);
```
jeffaco: This is part of the cludge. Shouldn't you be able to recover from this? If not, do you do something in the installer to add this initially? If so, I didn't see that!

Also, what's going to be returned here? Should we return some sort of error? Is just syslog() enough?

source/code/providers/Container_ContainerLog_Class_Provider.cpp
```diff
@@ -0,0 +1,288 @@
...
+        FILE* file = fopen(lastQueryFile, "w");
+
+        if (file)
+        {
+            fprintf(file, "%d", t);
+            fclose(file);
+        }
+        else
+        {
+            syslog(LOG_ERR, "Attempt in SetPreviousTime to open %s for writing failed", lastQueryFile);
```
jeffaco: Is just syslog() enough? Do we throw? Abort the operation? Make it more "obvious" that we're sick?

source/code/providers/Container_DaemonEvent_Class_Provider.cpp
```diff
@@ -181,14 +181,7 @@ class EventQuery
...
             // Get computer name
-            char name[256];
-            string hostname = gethostname(name, 256) ? "" : string(name);
-            // in case get full name, extract up to '.'
-            size_t dotpos = hostname.find('.');
-            if (dotpos != string::npos)
-            {
-                hostname = hostname.substr(0, dotpos);
-            }
+            string hostname = getDockerHostName();
```
jeffaco: See above, I won't comment on this anymore.

source/code/dockerapi/DockerRemoteApi.cpp
amitsara: When omsagent runs as a docker container, gethostname returns the hostname of the container (which is usually some random name) not the computer name.
Since omsagent container can talk to docker server, now it is getting computer name from the docker server.
Agree rest api is expensive than gethostname, that’s why I am keeping the name in static var so it will be one time cost only.

source/code/providers/Container_ContainerLog.h
amitsara: yes, it is generated code, nothing modified.

source/code/providers/Container_ContainerLog_Class_Provider.cpp
amitsara: Purpose of this file is to get the previous time when the container logs were collected. First time when the file is not present, provider will collect the latest logs and will write the current time in this file.
Next time file will be present and will have the previous time stored.

Any kind of error is not terminating here, syslog warning is sufficient. And in case of error again provider collects the latest logs and sends them to oms.

source/code/providers/Container_ContainerLog_Class_Provider.cpp
amitsara: syslog message is sufficient here, it is not terminating error. operation doesn't need to be aborted.

source/code/providers/Container_ContainerLog_Class_Provider.cpp
amitsara: Sure, I will create a state class in a separate pull request.

source/code/mof/docker.mof
agup006: CIM class Looks good to me

source/code/providers/Container_ContainerLog_Class_Provider.cpp
jeffaco: Note that there is a class in the PAL that may be of assistance for that, see either the older, more complicated (but perhaps more robust) [persist](https://github.com/Microsoft/pal/tree/master/source/code/scxcorelib/util/persist) stuff, or for something much smaller and simpler, look at the [config](https://github.com/Microsoft/pal/blob/master/source/code/scxcorelib/pal/scxconfigfile.cpp) stuff, which would work for this as well.

PR #19, opened 2016-10-20
Set 'containername' instead of 'containerId' to Perf->InstanceName field

amitsara:


amitsara:
@keikhara 
@Microsoft/omsagent-devs

keikhara:
sign-off.

PR #20, opened 2016-10-21
Fixed: 'opendir() returned null: Too many open files' issue.

amitsara:
docker-provider starts throwing 'opendir() returned null: Too many open files' error once opendir reaches to MAX_OPEN limit. 
An open directory must always be closed with the closedir to ensure that the next attempt to open that directory is successful.


amitsara:
@keikhara 
@Microsoft/omi-devs

keikhara:
sign-off

PR #21, opened 2016-12-14
Docker-Provider updates for robustness

raghushantha:
This PR addresses two issues:

1) Make the JSON parser more reliable to handle parsing requests

2) CIM Provider needs to handle cases when container stats do not contain 'Network' information in the REST response. We were segfaulting here earlier.


keikhara:
@Microsoft/omsagent-devs  please review this and sign-off.

PR #22, opened 2016-12-15
Remove non-applicable installation message

raghushantha:
We no longer use this global setting to configure the Docker log driver. This message is confusing and not required and has prevented our customers from performing Docker tasks.


raghushantha:
Got previous commits here. Will submit a new one

PR #26, opened 2017-03-23
Add separate buffer for container data types

samisms:
Separating the buffer for container data types. Container log volumes are quite high and they were filling up the common buffer leading to queues getting backed up. Separated thew buffers to alleviate the problem.
@Microsoft/omsagent-devs @keikhara

### DIFF
--- a/installer/conf/container.conf
+++ b/installer/conf/container.conf
@@ -64,3 +64,17 @@
 <filter oms.container.**>
 	type filter_container
 </filter>
+
+#separate buffer for containers
+<match oms.container.**>
+  type out_oms
+  log_level info
+  buffer_chunk_limit 20m
+  buffer_type file
+  buffer_path /var/opt/microsoft/omsagent/state/state/out_oms_docker*.buffer
+  buffer_queue_limit 20
+  flush_interval 20s
+  retry_limit 10
+  retry_wait 15s
+  max_retry_wait 9m
+</match>
--- a/installer/conf/container.conf
+++ b/installer/conf/container.conf
@@ -71,7 +71,7 @@
   log_level info
   buffer_chunk_limit 20m
   buffer_type file
-  buffer_path /var/opt/microsoft/omsagent/state/state/out_oms_docker*.buffer
+  buffer_path %STATE_DIR_WS%/out_oms_docker*.buffer
   buffer_queue_limit 20
   flush_interval 20s
   retry_limit 10


installer/conf/container.conf
```diff
@@ -64,3 +64,17 @@
...
 	type filter_container
 </filter>
+
+#separate buffer for containers
+<match oms.container.**>
+  type out_oms
+  log_level info
+  buffer_chunk_limit 20m
+  buffer_type file
+  buffer_path /var/opt/microsoft/omsagent/state/state/out_oms_docker*.buffer
```
lagalbra: You say "state" twice. Have you tested this?

samisms: image
 Yes, based on whatever path the omsagent uses right now.

lagalbra: Will this conf file ever get onto a VM without a Container? Perhaps that machine's setup is different, but with release 1.3.1-15, I have no ...state/state folder:

`ls: cannot access /var/opt/microsoft/omsagent/state/state/: No such file or directory`

samisms: Removed the hardcoded path , Tested by modifying omsadmin.sh to update the buffer path during install. Image shows container.conf contents after onboarding to a workspace
image
Will send out a separate PR for the omsadmin.sh

installer/conf/container.conf
```diff
@@ -64,3 +64,17 @@
...
 	type filter_container
 </filter>
+
+#separate buffer for containers
+<match oms.container.**>
+  type out_oms
+  log_level info
+  buffer_chunk_limit 20m
+  buffer_type file
+  buffer_path %STATE_DIR_WS%/out_oms_docker*.buffer
```
lagalbra: Where are you replacing this string? [omsadmin.sh](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/omsadmin.sh#L820) currently only replaces this string in a hardcoded set of files (search page for `update_path`).

samisms: if [ -s $OMSAGENTD_DIR/container.conf ] ; then
      update_path $OMSAGENTD_DIR/container.conf
fi
This is the change in copy_omsagent_d_conf() in omsadmin.sh I tested.

lagalbra: For reference, that is this PR: https://github.com/Microsoft/OMS-Agent-for-Linux/pull/420

PR #27, opened 2017-03-29
Fix for segfault in LogFileReader

samisms:
This PR contains two fixes.
1. Added proper null checks in LinkFilesToStream ContainerLogFileReader class. 
2. Made the cJSON_GetArrayItem in cJson.c more robust by adding null checks there as well, also imrpoved readability. 
@Microsoft/omsagent-devs @keikhara @jeffaco


source/code/cjson/cJSON.c
```diff
@@ -659,7 +659,18 @@ static char *print_object(cJSON *item,int depth,int fmt,printbuffer *p)
 
 /* Get Array size/item / object item. */
 int    cJSON_GetArraySize(cJSON *array)							{cJSON *c=array->child;int i=0;while(c)i++,c=c->next;return i;}
-cJSON *cJSON_GetArrayItem(cJSON *array,int item)				{cJSON *c=array->child;  while (c && item>0) item--,c=c->next; return c;}
+cJSON *cJSON_GetArrayItem(cJSON *array,int item)				
+{
+    cJSON *c = array ? array->child : NULL;
```
lagalbra: Not too experienced with C any longer, but I would recommend making this check explicit: `array != NULL`

source/code/cjson/cJSON.c
```diff
@@ -659,7 +659,18 @@ static char *print_object(cJSON *item,int depth,int fmt,printbuffer *p)
...
+    while (c && item > 0)
```
lagalbra: Same explicit check suggested here.

PR #28, opened 2017-04-21
Initial commit adding Type=ContainerNodeInventory_CL

samisms:
Added a new provider class Container_HostInventory_Class_Provider . It collects some node inventory data from the rest api version of "docker info". It also parses the node name to determine orchestrator. I have verified the functionality. I will work on adding a unit test for this. Meanwhile I could use some feedback on this. 
image
I have attached a sample of the search query in OMS
I also modified the frequency of data, ContainerNodeInventory_CL, ContainerInventory and ContainerImageInventory are set at a frequency of 1 minute, ContainerLog and Perf are set to a frequency of 15 s since they need to be more real time.
@Microsoft/omsagent-devs @jeffaco @keikhara @amitsara

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -238,7 +238,8 @@ STATIC_PROVIDERLIB_SRCFILES = \
 	$(PROVIDER_DIR)/Container_ContainerStatistics_Class_Provider.cpp \
 	$(PROVIDER_DIR)/Container_DaemonEvent_Class_Provider.cpp \
 	$(PROVIDER_DIR)/Container_ImageInventory_Class_Provider.cpp \
-	$(PROVIDER_DIR)/Container_ContainerLog_Class_Provider.cpp
+	$(PROVIDER_DIR)/Container_ContainerLog_Class_Provider.cpp \
+	$(PROVIDER_DIR)/Container_HostInventory_Class_Provider.cpp
 
 STATIC_PROVIDERLIB_SRCFILES += \
 	$(PROVIDER_DIR)/module.cpp \
@@ -281,6 +282,7 @@ STATIC_PROVIDER_UNITFILES = \
 	$(PROVIDER_TEST_DIR)/Container_ContainerStatistics_Class_Provider_UnitTest.cpp \
 	$(PROVIDER_TEST_DIR)/Container_DaemonEvent_Class_Provider_UnitTest.cpp \
 	$(PROVIDER_TEST_DIR)/Container_ImageInventory_Class_Provider_UnitTest.cpp \
+	$(PROVIDER_TEST_DIR)/Container_HostInventory_Class_Provider_UnitTest.cpp \
 	$(PROVIDER_TEST_DIR)/support/testrunnerlogpolicy.cpp \
 	$(PROVIDER_TEST_DIR)/support/productdependencies.cpp \
 	$(PROVIDER_TEST_DIR)/providertestutils.cpp
--- a/installer/conf/container.conf
+++ b/installer/conf/container.conf
@@ -17,13 +17,13 @@
 <source>
 	type oms_omi
 	object_name "Container"
-	interval 30s
+	interval 15s
 </source>
 
 # Container inventory
 <source>
 	type omi
-	run_interval 30s
+	run_interval 60s
 	tag oms.container.containerinventory
 	items [
 		["root/cimv2","Container_ContainerInventory"]
@@ -33,7 +33,7 @@
 # Image inventory
 <source>
 	type omi
-	run_interval 30s
+	run_interval 60s
 	tag oms.container.imageinventory
 	items [
 		["root/cimv2","Container_ImageInventory"]
@@ -60,10 +60,34 @@
 	]
 </source>
 
+# Container host inventory
+<source>
+	type omi
+	run_interval 60s
+	tag oms.api.ContainerNodeInventory
+	items [
+		["root/cimv2","Container_HostInventory"]
+	]
+</source>
+
 # Filter for correct format to endpoint
 <filter oms.container.**>
 	type filter_container
 </filter>
+
+<match oms.api.ContainerNodeInventory**>
+  type out_oms_api
+  log_level debug
+  buffer_chunk_limit 20m
+  buffer_type file
+  buffer_path %STATE_DIR_WS%/out_oms_containerinventory*.buffer
+  buffer_queue_limit 20
+  flush_interval 20s
+  retry_limit 10
+  retry_wait 15s
+  max_retry_wait 9m
+</match>
+
 
 #separate buffer for containers
 <match oms.container.**>
--- a/source/code/providers/Container_HostInventory_Class_Provider.cpp
+++ b/source/code/providers/Container_HostInventory_Class_Provider.cpp
@@ -1,0 +1,227 @@
+/* @migen@ */
+#include <MI.h>
+#include "Container_HostInventory_Class_Provider.h"
+
+#include <vector>
+#include <string>
+#include <syslog.h>
+#include <stdio.h>
+#include <uuid/uuid.h>
+#include <cstdlib>
+
+#include "../cjson/cJSON.h"
+#include "../dockerapi/DockerRemoteApi.h"
+#include "../dockerapi/DockerRestHelper.h"
+#include "Helper.h"
+
+using namespace std;
+
+MI_BEGIN_NAMESPACE
+
+class ContainerHostInventoryQuery
+{
+public:
+    ///
+    /// \returns Object representing the host
+    ///
+    static Container_HostInventory_Class InspectHost()
+    {
+        Container_HostInventory_Class hostInventoryInstance;
+
+        // Request containers
+        vector<string> request(1, DockerRestHelper::restDockerInfo());
+        vector<cJSON*> response;
+
+        //check to see if its a test environment
+        if(getenv(DOCKER_TESTRUNNER_STRING) != NULL)
+        {
+            //fake response for test
+            char* testResponseStr = getenv(DOCKER_TESTRUNNER_STRING);
+            string tempStr = testResponseStr;
+            cJSON* testJsonObject = parseJson(tempStr);
+            response.push_back(testJsonObject);
+        }
+        else
+        {
+            //regular code path, get response from rest call
+            response = getResponse(request);
+        }
+
+        // See https://docs.docker.com/engine/api/v1.27/#section/Versioning for example output
+        if (!response.empty() && response[0])
+        {   
+            PopulateInstanceFromJson(hostInventoryInstance, response[0]);
+            // Clean up object
+            cJSON_Delete(response[0]);
+        }
+        else
+        {
+            syslog(LOG_WARNING, "API call in Container_HostInventory::InspectHost to get host info returned null");
+        }
+        return hostInventoryInstance;
+    }
+
+    static void PopulateInstanceFromJson(Container_HostInventory_Class &instance, cJSON* responseJson)
+    {
+            instance.InstanceID_value(Guid::NewToString().c_str());
+            string computerName = string((cJSON_GetObjectItem(responseJson, "Name")->valuestring));
+            instance.Computer_value(computerName.c_str());
+            instance.DockerVersion_value((cJSON_GetObjectItem(responseJson, "ServerVersion")->valuestring));
+            instance.OperatingSystem_value((cJSON_GetObjectItem(responseJson, "OperatingSystem")->valuestring));
+            //Get Volume and Network info from "Plugins"
+            cJSON* plugins_entry = cJSON_GetObjectItem(responseJson, "Plugins");
+            if(plugins_entry != NULL)
+            {
+                //Get volume info
+                cJSON* volumeEntry = cJSON_GetObjectItem(plugins_entry, "Volume");
+                string volumeList;
+                if(volumeEntry != NULL)
+                {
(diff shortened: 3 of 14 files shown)


installer/conf/container.conf
```diff
@@ -53,26 +53,50 @@
...
 #separate buffer for containers
 <match oms.container.**>
   type out_oms
   log_level info
-  buffer_chunk_limit 20m
+  buffer_chunk_limit 8m
   buffer_type file
-  buffer_path %STATE_DIR_WS%/out_oms_docker*.buffer
-  buffer_queue_limit 20
+  buffer_path /var/opt/microsoft/omsagent/state/out_oms_docker*.buffer
```
lagalbra: Why are you removing the multi-homing path here?

samisms: Accident. I will revert this

build/Makefile
```diff
@@ -238,7 +238,8 @@ STATIC_PROVIDERLIB_SRCFILES = \
 	$(PROVIDER_DIR)/Container_ContainerStatistics_Class_Provider.cpp \
 	$(PROVIDER_DIR)/Container_DaemonEvent_Class_Provider.cpp \
 	$(PROVIDER_DIR)/Container_ImageInventory_Class_Provider.cpp \
-	$(PROVIDER_DIR)/Container_ContainerLog_Class_Provider.cpp
+	$(PROVIDER_DIR)/Container_ContainerLog_Class_Provider.cpp\
```
jeffaco: Nit: Add space before `\` (at end of line) for consistency.

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+        Container_HostInventory_Class hostInventoryInstance;
+
+        // Request containers
+        vector<string> request(1, DockerRestHelper::restDockerInfo());
+        vector<cJSON*> response = getResponse(request);
+
+        // See https://docs.docker.com/engine/api/v1.27/#section/Versioning for example output
+        if (!response.empty() && response[0])
+        {   
+            printf("inspect host \n");
```
jeffaco: Is this supposed to be a debug message? If so, why not use SCX logging, which I think you pretty much have anyway due to use of the PAL, which I think you'll need for unit test support.

I haven't gotten this far yet, but you do have unit tests for this, right?

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+    ///
+    static Container_HostInventory_Class InspectHost()
+    {
+        Container_HostInventory_Class hostInventoryInstance;
+
+        // Request containers
+        vector<string> request(1, DockerRestHelper::restDockerInfo());
+        vector<cJSON*> response = getResponse(request);
+
+        // See https://docs.docker.com/engine/api/v1.27/#section/Versioning for example output
```
jeffaco: Very nice to include this!

However, are you actually tying to a particular version of the API so you aren't broken down the road by some new version of Docker? I can't quite find the URL here that you're using for the Docker API to answer this myself ...

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+            {
+                hostInventoryInstance.OrchestratorType_value("None");  
+            }   
+                        
+            // Clean up object
+            cJSON_Delete(response[0]);
+        }
+        else
+        {
+            syslog(LOG_WARNING, "API call in Container_HostInventory::InspectHost to get host info returned null");
```
jeffaco: So is syslog your logging mechanism, vs. the PAL logging facilities? If so, you should use that for the `printf` statement above.

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+Container_HostInventory_Class_Provider::~Container_HostInventory_Class_Provider()
+{
+}
+
+void Container_HostInventory_Class_Provider::Load(
+        Context& context)
+{
+    context.Post(MI_RESULT_OK);
```
jeffaco: I presume it's perfectly okay for this provider to be unloaded, and doing so doesn't cause any performance issues?

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+void Container_HostInventory_Class_Provider::EnumerateInstances(
+    Context& context,
+    const String& nameSpace,
+    const PropertySet& propertySet,
+    bool keysOnly,
+    const MI_Filter* filter)
+{
+    printf("Container_HostInventory_Class_Provider ei \n");
```
jeffaco: syslog? PAL logging facilities?

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+    Container_HostInventory_Class queryResult = ContainerHostInventoryQuery::InspectHost();
+    context.Post(queryResult);
+    context.Post(MI_RESULT_OK);
```
jeffaco: Any chance of an exception here? You're not doing anything to capture an exception, which can cause an agent crash ...

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+void Container_HostInventory_Class_Provider::GetInstance(
+    Context& context,
+    const String& nameSpace,
+    const Container_HostInventory_Class& instanceName,
+    const PropertySet& propertySet)
+{
+    context.Post(MI_RESULT_NOT_SUPPORTED);
```
jeffaco: You don't want to support this for efficiency? I did note that you do have a key in the new class, so this can be supported (if it would ever be used). In general, we support these from [SCXcore](https://github.com/Microsoft/SCXcore).

source/code/providers/Container_HostInventory_Class_Provider.cpp
samisms: Not a particular version. GET /info HTTP/1.1\r\nHost: localhost\r\n\r\n .  Since things are still in flux maybe we can lock the version down the line. I would like to be flexible at this point.

installer/conf/container.conf
```diff
@@ -53,26 +53,50 @@
...
 #separate buffer for containers
 <match oms.container.**>
   type out_oms
   log_level info
-  buffer_chunk_limit 20m
+  buffer_chunk_limit 8m
   buffer_type file
   buffer_path %STATE_DIR_WS%/out_oms_docker*.buffer
-  buffer_queue_limit 20
+  buffer_queue_limit 1024
```
jeffaco: Pretty big jump (noticed same buffer_queue_limit up above). That seems pretty big, is that really what you want? How would this impact disk space on customer systems? Is this a situation where the customer system may not ever catch up from a backlog that's that large?

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,218 @@
...
+void Container_HostInventory_Class_Provider::EnumerateInstances(
+    Context& context,
+    const String& nameSpace,
+    const PropertySet& propertySet,
+    bool keysOnly,
+    const MI_Filter* filter)
+{
+    Container_HostInventory_Class queryResult = ContainerHostInventoryQuery::InspectHost();
+    context.Post(queryResult);
+    context.Post(MI_RESULT_OK);
```
jeffaco: Rather than putting the exception handler in InspectHost, the exception handler should go here. In case of an exception, you don't want to post MI_RESULT_OKAY, you want to post MI_RESULT_FAILED.

[SCXCore](https://github.com/Microsoft/SCXcore/blob/master/source/code/providers/SCX_Agent_Class_Provider.cpp#L219) has an example of an exception handler. This is hidden in a macro but, basically, if an exception occurs, then a different result is posted. But if all is good, MI_RESULT_OK is posted.

You do not want to post MI_RESULT_OK if the enumeration was NOT okay ...

See [here](https://github.com/Microsoft/SCXcore/blob/master/source/code/providers/support/scxcimutils.h) for a definition of that macro.

test/code/providers/Container_HostInventory_Class_Provider_UnitTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+class ContainerHostInventoryTest : public CppUnit::TestFixture
+{
+    CPPUNIT_TEST_SUITE(ContainerHostInventoryTest);
+    CPPUNIT_TEST(TestEnumerateInstances);
+    CPPUNIT_TEST_SUITE_END();
+
+public:
+    vector<string> containers;
+
+    static string NewGuid()
```
jeffaco: Who calls this? Is this stale code? Same for RunCommand, who calls that? I don't see any callers for that in TestEnumerateInstances, please cut/paste judiciously!

test/code/providers/Container_HostInventory_Class_Provider_UnitTest.cpp
```diff
@@ -0,0 +1,85 @@
...
+        // Enumerate provider
+        StandardTestEnumerateInstances<mi::Container_HostInventory_Class_Provider>(m_keyNames, context, CALL_LOCATION(errMsg));
+
+        CPPUNIT_ASSERT(context.Size());
+		
+        // Only check that the values are present and within the valid range because it is not possible to create a controlled environment
+        for (unsigned i = 0; i < context.Size(); ++i)
+        {
+            wstring instanceId = context[i].GetProperty(L"InstanceID", CALL_LOCATION(errMsg)).GetValue_MIString(CALL_LOCATION(errMsg));
+            CPPUNIT_ASSERT(instanceId.length());
```
jeffaco: This is a really weak test. You're testing that some non-empty string was returned, but that's it. Can't you do a little more robust checking, like that the values are actually reasonable?

Same issue for everything below, I think you can do a lot more robust checking. A crappy unit test can miss a lot of stuff. A good unit test will catch a lot of problems very early in the DEV cycle (a very good thing).

jeffaco:
By the way, your commit message is really weak. Something better, like perhaps:

Added Container_HostInventory class to enumerate xxxxx (whatever it is you're enumerating)

jeffaco:
Adding @niroyb to discuss queue size

source/code/providers/Container_HostInventory_Class_Provider.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+#include "../cjson/cJSON.h"
+#include "../dockerapi/DockerRemoteApi.h"
+#include "../dockerapi/DockerRestHelper.h"
+
+using namespace std;
+
+MI_BEGIN_NAMESPACE
+
+class Guid
```
jeffaco: This was copied from elsewhere. Please put it in a header file and reference it from there, here in this code and from any other code that needs it.

installer/conf/container.conf
```diff
@@ -60,11 +60,35 @@
...
+  buffer_type file
+  buffer_path %STATE_DIR_WS%/out_oms_containerinventory*.buffer
+  buffer_queue_limit 20
+  flush_interval 20s
+  retry_limit 10
+  retry_wait 15s
+  max_retry_wait 9m
+</match>
+
+
```
jeffaco: Be sure to test these numbers and be sure that the agent has no problem keeping up with a reasonable expected load.

test/code/providers/Container_HostInventory_Class_Provider_UnitTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+            CPPUNIT_ASSERT_EQUAL(std::wstring(L"swarm-agent-71E8D996-0"), context[i].GetProperty(L"Computer", CALL_LOCATION(errMsg)).GetValue_MIString(CALL_LOCATION(errMsg))) ;
+            CPPUNIT_ASSERT_EQUAL(std::wstring(L"Agent"), context[i].GetProperty(L"NodeRole", CALL_LOCATION(errMsg)).GetValue_MIString(CALL_LOCATION(errMsg)));
+            CPPUNIT_ASSERT_EQUAL(std::wstring(L"Swarm"), context[i].GetProperty(L"OrchestratorType", CALL_LOCATION(errMsg)).GetValue_MIString(CALL_LOCATION(errMsg)));
+        }
+    }
+    void TestNonOrchestratedParsing()
```
jeffaco: Insert blank line above this

test/code/providers/Container_HostInventory_Class_Provider_UnitTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+    {
+        wstring errMsg;
+        TestableContext context;
+        vector<wstring> m_keyNames;
+        m_keyNames.push_back(L"InstanceID");
+
+        // Enumerate provider. This just tests if the provider is able to pick up properties from a docker info call on the machine
+        StandardTestEnumerateInstances<mi::Container_HostInventory_Class_Provider>(m_keyNames, context, CALL_LOCATION(errMsg));
+
+        CPPUNIT_ASSERT(context.Size());
```
jeffaco: Please do an ASSERT_EQ that the size is 1, not some non-zero size.

Change all other tests that do this as well.

PR #30, opened 2017-06-09
Adding journald support

samisms:
The docker rest endpoint supports both json-file and journald logging drivers. There is no difference in the rest response between the two and so just allowing journald driver
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/providers/Container_ContainerLog_Class_Provider.cpp
+++ b/source/code/providers/Container_ContainerLog_Class_Provider.cpp
@@ -109,7 +109,8 @@ public:
         deque<Container_ContainerLog_Class> result;
 
         string logDriver = getLogDriverName();
-        if (logDriver.compare("json-file") != 0)
+        //Docker rest api endpoint supports both json-file and journald logging drivers
+        if ((logDriver.compare("json-file") != 0) && (logDriver.compare("journald") != 0))
         {
             return result;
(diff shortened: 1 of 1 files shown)


source/code/providers/Container_ContainerLog_Class_Provider.cpp
```diff
@@ -109,7 +109,8 @@ class ContainerLogQuery
         deque<Container_ContainerLog_Class> result;
 
         string logDriver = getLogDriverName();
-        if (logDriver.compare("json-file") != 0)
+        //Docker rest api endpoint supports both json-file and journald logging drivers
+        if ((logDriver.compare("json-file") != 0) && (logDriver.compare("journald") != 0))
```
lagalbra: Shouldn't this be an OR, not an AND operation? Can the logDriver be identified both as "json-file" and "journald"?

samisms: If its neither then it returns immediately without doing any log collection since we can only support json-file or journald, else it continues with the log parsing

lagalbra: Ok, I see. Looks fine to me.

amitsara:
looks good, sign off

PR #48, opened 2018-01-11
Adding separate buffer for each data type

samisms:
The separate buffer solves the issue where one datatype doesnt block other datatypes from getting sent.


samisms:
Merged

PR #61, opened 2018-01-30
Installer fixes

samisms:
This PR addresses the following things
1. Fixed a copy paste error in in_kube_events.rb . The debug_backtrace was using e instead of errorStr.
2. Modified the kube api client logging file to "/var/opt/microsoft/docker-cimprov/log/kubernetes_client_log.txt". This accounts for a race condition where the omsagent is installed but /var/opt/microsoft/log directory isnt created yet because of which sometimes there are errors while loading KubernetesApiClient.rb . It doesnt have any impact towards functionality , however it shows up as an error in the omsagent log. Moving it docker-cimprov direcroty makes sense since the docker-provider has total control of that directory and the can create the necessary directory and files. 
3. Modified base_container.data to create /var/opt/microsoft/docker-cimprov/log/kubernetes_client_log.txt with appropriate permissions. 
4. Modified base_container.data to setup sudo permission for containerlogtailfilereader. It had been implemetned for a bash shell, however the installer uses a bourne shell instead and so the if condition would throw an error on run and not do the requisite. Changed it to be consistent with bourne shell. 
5. Modified base_container.data so as to clean up properly when run with --remove or --purge 
6. Modified in_kube_logs.rb to use debug log level instead of info. This makes it less verbose.

### DIFF
--- a/installer/datafiles/base_container.data
+++ b/installer/datafiles/base_container.data
@@ -69,6 +69,7 @@ MAINTAINER:              'Microsoft Corporation'
 /var/opt/microsoft/docker-cimprov/state;                755; root; root
 /var/opt/microsoft/docker-cimprov/state/ContainerInventory; 755; root; root
 /var/opt/microsoft/docker-cimprov/state/ImageInventory; 755; root; root
+/var/opt/microsoft/docker-cimprov/log;                  755; root; root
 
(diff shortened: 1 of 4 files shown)


samisms:
Will merge to ci_feature

PR #62, opened 2018-01-31
Merge Vishwa/limits requests to ci_feature

vishiy:
* Enabled collecting requests @ limits for cpu & memory if available at the container level
Metrics (Perf counter name in 'Perf' entity) ---
1)cpuRequestNanoCores
2)cpuLimitNanoCores
3)memoryRequestBytes
4)memoryLimitBytes
with ObjectName = "K8SContainer"

memory units supported - E, P, T, G, M, K, Ei, Pi, Ti, Gi, Mi, Ki
cpu units supported - m

### DIFF
--- a/source/code/plugin/KubernetesApiClient.rb
+++ b/source/code/plugin/KubernetesApiClient.rb
@@ -221,8 +221,8 @@ class KubernetesApiClient
                         if (!pod['spec']['containers'].nil?)
                             pod['spec']['containers'].each do |container|
                                 containerName = container['name']
-                                if (!container['resources'].nil? && !container['resources'][metricCategory] && !container['resources'][metricCategory][metricNameToCollect])
-                                    metricValue = getMetricNumericValue(metricNameToCollect, container['resources'][metricCategory][memoryMetricNameToCollect])
+                                if (!container['resources'].nil? && !container['resources'][metricCategory].nil? && !container['resources'][metricCategory][metricNameToCollect].nil?)
+                                    metricValue = getMetricNumericValue(metricNameToCollect, container['resources'][metricCategory][metricNameToCollect])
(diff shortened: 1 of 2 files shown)


source/code/plugin/KubernetesApiClient.rb
```diff
@@ -210,6 +211,123 @@ def getContainerLogsSinceTime(namespace, pod, container, since, showTimeStamp)
                 end
                 return containerLogs
             end
+
+            def getContainerResourceRequestsAndLimits(metricJSON, metricCategory, hostName, metricNameToCollect, metricNametoReturn)
```
samisms: Minor edit but would you consider renaming getContainerResourceRequestsAndLimits to parse* since the actual get is happening in in_kube_perf.rb

source/code/plugin/KubernetesApiClient.rb
```diff
@@ -210,6 +211,123 @@ def getContainerLogsSinceTime(namespace, pod, container, since, showTimeStamp)
...
+                        end
+                    end
+                    rescue => error
+                    @Log.warn("getcontainerResourceRequestsAndLimits failed: #{error} for metric #{metricCategory} #{metricNameToCollect}")
+                    return metricItems
+                end
+                return metricItems          
+            end #getContainerResourceRequestAndLimits
+
+            def getMetricNumericValue(metricName, metricVal)
```
samisms: Same as above, consider changing get* to parse*

samisms: Also consider moving both getContainerResourceRequestsAndLimits and getMetricNumericValue to in_kube_podinventory  itself. KubernetesApiClient.rb is written as a client to talk to the kube api and deal with properties that are applicable to the cluster as a whole e.g. ClusterName .

PR #84, opened 2018-03-26
* Rotate Log to 10 MB; keep last 2 logs for CAdvisor & Kube API logs

vishiy:
* Rotate Log to 10 MB; keep last 2 logs for CAdvisor & Kube API logs
* Bump up #threads for output plugins to 5 (from 1)
* Increase log level to Trace for all Kube* plugins (to troubleshoot data stoppage issue)
* Remove noisy trace logs

### DIFF
--- a/source/code/plugin/CAdvisorMetricsAPIClient.rb
+++ b/source/code/plugin/CAdvisorMetricsAPIClient.rb
@@ -13,7 +13,7 @@ class CAdvisorMetricsAPIClient
             require_relative 'KubernetesApiClient'
     
             @LogPath = "/var/opt/microsoft/omsagent/log/kubernetes_perf_log.txt"
-            @Log = Logger.new(@LogPath, 'weekly')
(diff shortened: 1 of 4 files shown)


source/code/plugin/CAdvisorMetricsAPIClient.rb
```diff
@@ -13,7 +13,7 @@ class CAdvisorMetricsAPIClient
             require_relative 'KubernetesApiClient'
     
             @LogPath = "/var/opt/microsoft/omsagent/log/kubernetes_perf_log.txt"
```
samisms: Can you update the file path here to var/opt/microsoft/docker-cimprov/ similar to /var/opt/microsoft/docker-cimprov/log/kubernetes_client_log.txt . Trying to remove dependency from omsagent directories as much as we can.

vishiy: will fix in the next round.

PR #109, opened 2018-06-19
plugin changes for test validation suite

rashmichandrashekar:
Adding logs after successful plugin processing.


rashmichandrashekar:
@vishiy : Have updated the plugin code to check the count, pls take a look

PR #112, opened 2018-07-03
Update encoding to UTF-8 for LogEntry

bragi92:


bragi92:
@vishiy Yes, let's wait for confirmation from the customer before we merge this into master. Do you think there is any other form of testing I can do on the host agent for this change?

vishiy:
Thanks @bragi92 . So far from what we saw yesterday, looks we covered all. Since we need to bring in this change to ci_feature as well, it would be worth a quick pass with a private image after the customer confirms that the fix is working?

bragi92:
Agreed.

bragi92:
@vishiy Should we wait to merge it to master when we're done with the tests for the containerized agent?
We will also need to create a release and update the OMSAgent code when their release is done.

PR #115, opened 2018-07-17
force UTF-8 encoding in the containerized OMS Agent

bragi92:


bragi92:
@vishiy Initial testing with containerized agent looks good. Will merge to ci_feature with your approval.

PR #116, opened 2018-07-18
Rashmi/containerinventoryfix

rashmichandrashekar:
This includes truncation of environment variable with size higher than 200kb.
Exception handlers for all providers


rashmichandrashekar:
Adding exception handlers and env variable truncation

PR #119, opened 2018-07-25
Rashmi/segfault fix latest

rashmichandrashekar:
Fix for seg faults in all providers


rashmichandrashekar:
Merging changes for seg fault fixes

PR #123, opened 2018-08-02
fixing the plugin logs for emit stream

rashmichandrashekar:
Fixing log writes to omsagent.log after checking the environment variable for acs-engine


rashmichandrashekar:
Merging changes for acs engine test

PR #129, opened 2018-09-18
Using KubeAPI for getting image,name. Adding more logs

r-dilip:


installer/conf/td-agent-bit.conf
```diff
@@ -2,13 +2,13 @@
     Flush         5
     Log_Level     info
     Parsers_File  /etc/td-agent-bit/parsers.conf
-    Log_File      /var/log/fluent-bit.log
+    Log_File      /etc/td-agent-bit/fluent-bit.log
```
vishiy: How is /var host ?

installer/conf/td-agent-bit.conf
```diff
@@ -2,13 +2,13 @@
...
 
 [INPUT]
     Name tail
     Tag oms.container.log.*
     Path /var/log/containers/*.log
-    DB /var/log/fblogs.db
+    DB /etc/td-agent-bit/fblogs.db
```
vishiy: can u please change this to /var/opt/microsoft/docker-cimprov/state/fblogs.db ?

source/code/go/src/plugins/oms.go
```diff
@@ -108,29 +89,25 @@ func populateMaps() {
...
-		}
-		var containerInventory ContainerInventory
-		unmarshallErr := json.Unmarshal(fileContent, &containerInventory)
-
-		if unmarshallErr != nil {
-			Log("Unmarshall error when reading file %s %s \n", fullPath, unmarshallErr.Error())
+	for _, pod := range pods.Items {
+		for _, status := range pod.Status.ContainerStatuses {
+			lastSlashIndex := strings.LastIndex(status.ContainerID, "/")
+			containerID := status.ContainerID[lastSlashIndex+1 : len(status.ContainerID)]
```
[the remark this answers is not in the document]
r-dilip: that will still be an empty string. So we should be fine here.

PR #130, opened 2018-09-26
Dilipr/mark comments

r-dilip:


source/code/go/src/plugins/oms.go
```diff
@@ -318,39 +288,79 @@ func toString(s interface{}) string {
 func getContainerIDFromFilePath(filepath string) string {
 	start := strings.LastIndex(filepath, "-")
 	end := strings.LastIndex(filepath, ".")
+	if start >= end || start == -1 || end == -1 {
+		// This means the file is not a managed Kubernetes docker log file.
+		// Drop all records from the file
+		return ""
```
vishiy: not logging the file name as part of error ?

source/code/go/src/plugins/out_oms.go
```diff
@@ -50,6 +48,8 @@ func FLBPluginFlush(data unsafe.Pointer, length C.int, tag *C.char) int {
 
 // FLBPluginExit exits the plugin
 func FLBPluginExit() int {
+	KubeSystemContainersRefreshTicker.Stop()
```
vishiy: Is it possible to check if its started before stopping ?

r-dilip: I believe the ticker automatically starts by itself. So the check is not required.

PR #140, opened 2018-10-04
Fixing CPU Utilization and removing Fluent-bit filters

r-dilip:


installer/conf/td-agent-bit.conf
```diff
@@ -12,23 +12,9 @@
...
-
-[FILTER]
-    Name modify
-    Match oms.container.log.*
-    Rename log LogEntry
-    Rename stream LogEntrySource
-    Rename time LogEntryTimeStamp
-    Rename filepath Filepath
-    Add_if_not_present SourceSystem Containers
+    Buffer_Chunk_Size 1m
```
r-dilip: Vishwa, please verify these values

PR #141, opened 2018-10-09
Minor tweaks 1. Remove some logging 2. Added more Error Handling 3. Continue when there is an error with k8s api

r-dilip:


source/code/go/src/plugins/oms.go
```diff
@@ -253,6 +246,10 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 			DataItems: dataItems}
 
 		marshalled, err := json.Marshal(logEntry)
+		if err != nil {
+			Log("Error while Marshalling log Entry: %s", err.Error())
+			return output.FLB_RETRY
```
joby-thomas: Will we end up in a situation where we will keep retrying? Previously we were sending that packet to ODS and now we are retrying from the agent.

vishiy: We should do FLB_OK (and log a metric for jsonmarshaiing error) and continue here as it could be poisonous data and there is no point in asking FB to retry as we will hit the same again..

r-dilip:
Returning an OK instead of RETRY. This Batch will be a lost cause.

PR #149, opened 2018-10-23
Telemetry Updates

r-dilip:


source/code/go/src/plugins/oms.go
```diff
@@ -319,12 +342,14 @@ func InitializePlugin(pluginConfPath string) {
 	pluginConfig, err := ReadConfiguration(pluginConfPath)
 	if err != nil {
 		Log("Error Reading plugin config path : %s \n", err.Error())
+		SendException(err)
```
vishiy: You might want to send the string that you are logging below this line as well. Else you dont know from where all these errors are coming from and hard to match to code.

source/code/go/src/plugins/telemetry.go
```diff
@@ -87,7 +88,7 @@ func initialize(telemetryPushIntervalProperty string, agentVersion string) (int,
 		CommonProperties["ACSResourceName"] = ""
 		CommonProperties["AKS_RESOURCE_ID"] = aksResourceID
 		splitStrings := strings.Split(aksResourceID, "/")
-		if len(aksResourceID) > 0 && len(aksResourceID) < 10 {
+		if len(splitStrings) > 0 && len(splitStrings) < 10 {
 			CommonProperties["SubscriptionID"] = splitStrings[2]
 			CommonProperties["ResourceGroupName"] = splitStrings[4]
 			CommonProperties["ClusterName"] = splitStrings[8]
```
vishiy: Check for bounds rather than assuming [8] exists

source/code/go/src/plugins/telemetry.go
```diff
@@ -138,3 +144,10 @@ func SendEvent(eventName string, dimensions map[string]string) {
 
 	TelemetryClient.Track(event)
 }
+
+// SendException  send an event to the configured app insights instance
+func SendException(err interface{}) {
+	if TelemetryClient != nil {
+		TelemetryClient.TrackException(err)
```
vishiy: We need all dimensions like we have for events & metrics.

source/code/go/src/plugins/telemetry.go
```diff
@@ -110,19 +111,24 @@ func SendContainerLogFlushRateMetric(telemetryPushIntervalProperty string, agent
...
 	SendEvent(eventNameContainerLogInit, make(map[string]string))
 
 	for ; true; <-ContainerLogTelemetryTicker.C {
 		SendEvent(eventNameDaemonSetHeartbeat, make(map[string]string))
-		DataUpdateMutex.Lock()
+		elapsed := time.Since(start)
+		ContainerLogTelemetryMutex.Lock()
 		flushRate := FlushedRecordsCount / FlushedRecordsTimeTaken * 1000
-		Log("Flushed Records : %f Time Taken : %f flush Rate : %f", FlushedRecordsCount, FlushedRecordsTimeTaken, flushRate)
+		logRate := FlushedRecordsCount / float64(elapsed/time.Second)
```
vishiy: Can u pls explain what you are trying to do here ? I see start and then for everytick you calculate elapsed from start ? Dont you have to reset start to elapsed ?

source/code/go/src/plugins/oms.go
r-dilip: Sounds good

source/code/go/src/plugins/telemetry.go
r-dilip: will get it automatically, since the common dimensions are on the telemetry client object.

PR #159, opened 2018-10-25
Changes to send omsagent/omsagent-rs kubectl logs to App Insights

r-dilip:


installer/conf/td-agent-bit.conf
```diff
@@ -14,6 +14,15 @@
     Path_Key filepath
     Skip_Long_Lines On
 
+[INPUT]
+    Name tail
+    Tag oms.container.log.flbplugin.*
+    Path /var/log/containers/omsagent*.log
+    DB /var/opt/microsoft/docker-cimprov/state/fblogs.db
```
vishiy: do you know if this is best practice (specifying & using and same state db  for 2 input tails?). we need to make sure this telemetry enablement does not impact log collection. i would rather use a different db .

installer/conf/td-agent-bit.conf
```diff
@@ -14,6 +14,15 @@
     Path_Key filepath
     Skip_Long_Lines On
 
+[INPUT]
+    Name tail
+    Tag oms.container.log.flbplugin.*
```
vishiy: the oms.container.log.* tails omsagent*.log as well. its that we need to send this to our telemetry AI (and also to customers workspace if kube-sys log collection is enabled). should we need anither tailer which is doing subset of the already exsting one? In this case, we can simply check the file name and route to AI (and also to workspace if kube-sys collection is enabled).

source/code/go/src/plugins/out_oms.go
```diff
@@ -50,6 +50,12 @@ func FLBPluginFlush(data unsafe.Pointer, length C.int, tag *C.char) int {
 		}
 		records = append(records, record)
 	}
+
+	incomingTag := C.GoString(tag)
+	if strings.Contains(strings.ToLower(incomingTag), "flbplugin") {
```
vishiy: pls check for full tag

source/code/go/src/plugins/telemetry.go
```diff
@@ -103,51 +146,14 @@ func initialize(telemetryPushIntervalProperty string, agentVersion string) (int,
...
-	TelemetryClient.Track(event)
-}
-
-// SendException  send an event to the configured app insights instance
-func SendException(err interface{}) {
-	if TelemetryClient != nil {
-		TelemetryClient.TrackException(err)
-	}
+	traceEntry := strings.Join(logLines, "\n")
+	TelemetryClient.TrackTrace(traceEntry, 1)
```
vishiy: do u know if there is a size limit imposed by AI for traceEntry in TrackTrace?

installer/conf/td-agent-bit.conf
r-dilip: if you look at the db, it is basically a set of files, along with the last position that was read. So, in my opinion, should be ok. that being said, will do more testing to see if there is any impact to log collection

installer/conf/td-agent-bit.conf
r-dilip: As discussed offline, this way, we can keep the logs and telemetry separate by having two separate tailers

source/code/go/src/plugins/out_oms.go
r-dilip: Will do.

PR #162, opened 2018-11-07
Fix Telemetry Bug -- Computer is blank

r-dilip:


source/code/go/src/plugins/oms.go
```diff
@@ -398,6 +391,13 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 	Computer = strings.TrimSuffix(ToString(containerHostName), "\n")
 	Log("Computer == %s \n", Computer)
 
+	ret, err := InitializeTelemetryClient(agentVersion)
```
vishiy: Can u pls explain how moving this down here fixes this issue ?

vishiy: I see it now. WksID & computer were being set after initialize was called.

PR #174, opened 2018-12-16
Rashmi/kubenodeinventory

rashmichandrashekar:
Moving docker version and OS to KubeNodeInventory
Adding pod and controller count to telemetry
Adding Node core capacity and memory to telemetry

### DIFF
--- a/installer/conf/kube.conf
+++ b/installer/conf/kube.conf
@@ -118,6 +118,19 @@
   max_retry_wait 9m
 </match>
 
+<match oms.api.ContainerNodeInventory**>
+  type out_oms_api
+  log_level debug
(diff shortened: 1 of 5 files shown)


installer/conf/kube.conf
```diff
@@ -118,6 +118,19 @@
   max_retry_wait 9m
```
vishiy: this change will have to be made in the aks yaml as well (with the 2 commands i gave you). Please track it as part of next deployment. Also, omsdocker repo needs to be updated with yaml changes + helm chart needs to be updated with yaml changes. Please track all these workitems,

rashmichandrashekar: Just created work items for these.

PR #181, opened 2019-01-04
Exclude docker containers from container inventory

rashmichandrashekar:
Exclude docker containers from container inventory

### DIFF
--- a/source/code/plugin/DockerApiClient.rb
+++ b/source/code/plugin/DockerApiClient.rb
@@ -86,7 +86,10 @@ class DockerApiClient
             containers = getResponse(request, true, false)
             if !containers.nil? && !containers.empty?
                 containers.each do |container|
-                    ids.push(container['Id'])
+                    labels = container['Labels']
+                    if !labels.nil? && !labels['io.kubernetes.pod.uid'].nil?
+                        ids.push(container['Id'])
+                    end
                 end
             end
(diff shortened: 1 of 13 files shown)


source/code/plugin/DockerApiClient.rb
```diff
@@ -86,7 +86,10 @@ def listContainers()
             containers = getResponse(request, true, false)
             if !containers.nil? && !containers.empty?
                 containers.each do |container|
-                    ids.push(container['Id'])
+                    labels = container['Labels']
+                    if !labels.nil? && !labels['io.kubernetes.pod.uid'].nil?
```
vishiy: Can you please make this a case-insensitive comparison, just in case... Thanks.

rashmichandrashekar: Sure. Fixed

vishiy: Hi Rashmi - I was intending to say the label value (kubernetes.pod.uid) being case insensitive (not the property name)

rashmichandrashekar: Sure.

PashaPash:
Is that a fix for "pause-amd64" container inventory?

rashmichandrashekar:
@Pashapash: yes it is

PR #187, opened 2019-01-24
add ContainerNotRunning column to KubePodInventory

bragi92:


source/code/plugin/in_kube_podinventory.rb
```diff
@@ -188,6 +189,10 @@ def parse_and_emit_records(podInventory, serviceList)
               #Picking up both container and node start time from cAdvisor to be consistent
               if containerStatus.keys[0] == "running"
```
ganga1980: :)

bragi92: :)

source/code/plugin/in_kube_podinventory.rb
```diff
@@ -171,6 +171,7 @@ def parse_and_emit_records(podInventory, serviceList)
               containerRestartCount = container['restartCount']		
               record['ContainerRestartCount'] = containerRestartCount
               containerStatus = container['state']
+              record['ContainerNotRunningReason'] = ''
```
vishiy: Rename something like 'ContainerStatusReason' so that in the future if we want to do reason for non 'not-running' status, we can use the same field?

bragi92: updated to 'ContainerStatusReason'

bragi92:
UX and workflow changes deployed in all regions - good to merge now.

PR #188, opened 2019-01-30
PR for pushing to topic branch

r-dilip:


source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+
+        def start
+            super
+            @@custom_metrics_template = '
+            { 
+                "time": "%{timestamp}", 
+                "data": { 
+                    "baseData": { 
+                        "metric": "%{metricName}", 
+                        "namespace": "Insights.Containers/node",
```
vishiy: These might change again. For now Keiko agreed to 'Insights.Container/nodes' (singular 'Container' and plural 'nodes'). Please update.

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+module Fluent
+    require 'logger'
+    require 'json'
+
+	class CAdvisor2MdmFilter < Filter
+		Fluent::Plugin.register_filter('filter_cadvisor2mdm', self)
+		
+		config_param :enable_log, :integer, :default => 0
+        config_param :log_path, :string, :default => '/var/opt/microsoft/omsagent/log/filter_cadvisor2mdm.log'
+        config_param :custom_metrics_azure_regions, :string
```
vishiy: Please out all our logs in one place. /var/opt/microsoft/dockercimprov/logs/

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+                begin 
+                    nodeInventory = JSON.parse(KubernetesApiClient.getKubeResourceInfo('nodes').body)
+                rescue Exception => e
+                    @log.info "Error when getting nodeInventory from kube API. Exception: #{e.class} Message: #{e.message} "
+                    ApplicationInsightsUtility.sendExceptionTelemetry(e.backtrace)
+                end
+                if !nodeInventory.nil? 
+                    cpu_limit_json = KubernetesApiClient.parseNodeLimits(nodeInventory, "capacity", "cpu", "cpuCapacityNanoCores")
+                    if !cpu_limit_json.nil? 
+                        @cpu_limit = cpu_limit_json[0]['DataItems'][0]['Collections'][0]['Value']
```
vishiy: knit - please rename to capacity (both cpu & memory) for more clarity..

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+                @memory_limit = 0.0 
+                
+                begin 
+                    nodeInventory = JSON.parse(KubernetesApiClient.getKubeResourceInfo('nodes').body)
+                rescue Exception => e
+                    @log.info "Error when getting nodeInventory from kube API. Exception: #{e.class} Message: #{e.message} "
+                    ApplicationInsightsUtility.sendExceptionTelemetry(e.backtrace)
+                end
+                if !nodeInventory.nil? 
+                    cpu_limit_json = KubernetesApiClient.parseNodeLimits(nodeInventory, "capacity", "cpu", "cpuCapacityNanoCores")
```
vishiy: Can you also do a quick check on the Ux repo, that the Ux is using the 'cpuCapacityNanoCores' and not 'cpuAllocatableNanoCores' to calculate % ?

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+        def build_metrics_hash
+            @log.debug "Building Hash of Metrics to Collect"
+            metrics_to_collect_arr = @metrics_to_collect.split(',').map(&:strip)
+            metrics_hash = metrics_to_collect_arr.map {|x| [x.downcase,true]}.to_h
+            @log.info "Metrics Collected : #{metrics_hash}"
+            return metrics_hash
+        end
+
+        def check_custom_metrics_availability
+            aks_region = ENV['AKS_REGION']
```
vishiy: You should check explicitly for AKS_REGION & AKS_RESOURCE_ID as you depend on both.

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+                @log.debug "Returning false for check_custom_metrics_availability"
+                false
+            end
+        end
+
+		def shutdown
+			super
+		end
+
+        def filter(tag, time, record)
```
vishiy: Please add global rescue for all 'implicitly' called methods , else if you crash once, fluentD will unload your filter plugin and will not be running until the agent (and hence fluentD) is restarted.

source/code/plugin/out_mdm.rb
```diff
@@ -0,0 +1,219 @@
...
+      else
+        @log.info "Last Failed POST attempt to MDM was made #{((Time.now - @last_post_attempt_time)/60).round(1)} min ago. This is less than the current retry threshold of #{@retry_mdm_post_wait_minutes} min. NO-OP"
+      end
+    end
+
+    def send_to_mdm(post_body) 
+      begin
+        access_token = get_access_token
+        uri = URI.parse(@@post_request_url)
+        http = Net::HTTP.new(uri.host, uri.port)
```
vishiy: whats the default time out ?

source/code/plugin/out_mdm.rb
```diff
@@ -0,0 +1,219 @@
...
+        http.use_ssl = true
+        request = Net::HTTP::Post.new(uri.request_uri)
+        request['Content-Type'] = "application/x-ndjson"
+        request['Authorization'] = "Bearer #{access_token}"
+        request.body = post_body.join("\n")
+        response = http.request(request)
+        response.value # this throws for non 200 HTTP response code
```
vishiy: there will not be any response.code for timeout. Please check for response.null & response.code.

source/code/plugin/out_mdm.rb
```diff
@@ -0,0 +1,219 @@
...
+        @log.info "HTTP Post Response Code : #{response.code}"
+        ApplicationInsightsUtility.sendCustomEvent("AKSCustomMetricsMDMSendSuccessful", {})
+      rescue Net::HTTPServerException => e
+        @log.info "Failed in Post: #{e} Code: #{response.code}"
+        if response.code == 403.to_s
+          @log.info "Response Code #{response.code} Updating @last_post_attempt_time"
```
vishiy: this can throw if there is no response.code. Again, you need global handler for all implictly called methods (which is the entry point for your custom methods)

source/code/plugin/out_mdm.rb
```diff
@@ -0,0 +1,219 @@
...
+      begin
+        access_token = get_access_token
+        uri = URI.parse(@@post_request_url)
+        http = Net::HTTP.new(uri.host, uri.port)
+        http.use_ssl = true
+        request = Net::HTTP::Post.new(uri.request_uri)
+        request['Content-Type'] = "application/x-ndjson"
+        request['Authorization'] = "Bearer #{access_token}"
+        request.body = post_body.join("\n")
+        response = http.request(request)
```
vishiy: so the magic here is value() raising if non-2xx ? Please ensure that the data indeed go to buffer and get resent next time.

source/code/plugin/out_mdm.rb
vishiy: Make httpclient static and reuse so you dont have to new up every time.

source/code/plugin/out_mdm.rb
```diff
@@ -0,0 +1,219 @@
...
+        return [tag, record].to_msgpack
+      else
+        return ""
+      end
+    end
+
+    # This method is called every flush interval. Send the buffer chunk to MDM. 
+    # 'chunk' is a buffer chunk that includes multiple formatted records
+    def write(chunk)
+      if !@first_post_attempt_made || (Time.now > @last_post_attempt_time + retry_mdm_post_wait_minutes*60)
```
vishiy: Can you please explain what you are trying to do here ? I see you set both first_ & last_ only when you see 403. So initially you see 200 and then if you start to see 403, whats the behavior ?

source/code/plugin/filter_cadvisor2mdm.rb
r-dilip: Done

source/code/plugin/filter_cadvisor2mdm.rb
r-dilip: Done

source/code/plugin/filter_cadvisor2mdm.rb
r-dilip: UX used cpuCapacityNanoCores

source/code/plugin/out_mdm.rb
r-dilip: Correct. So I simply need to re-raise the exception

source/code/plugin/out_mdm.rb
r-dilip: Both open and read timeouts are 60s, which I think is a good default value

r-dilip: Made it an instance variable.

source/code/plugin/filter_cadvisor2mdm.rb
r-dilip: I will handle all HTTP Server Exceptions inside the filter, everything else will be handled by the global handler, else we can't do retry

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -0,0 +1,214 @@
...
+            @process_incoming_stream = check_custom_metrics_availability
+            @metrics_to_collect_hash = build_metrics_hash
+            @log.debug "After check_custom_metrics_availability process_incoming_stream #{@process_incoming_stream}"
+
+            # initialize cpu and memory limit 
+            if @process_incoming_stream
+                @cpu_limit = 0.0
+                @memory_limit = 0.0 
+                
+                begin
```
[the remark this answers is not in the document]
r-dilip: Scoped to the current node

source/code/plugin/out_mdm.rb
r-dilip: this will be handled by a different exception.

source/code/plugin/out_mdm.rb
r-dilip: null check added

PR #194, opened 2019-02-16
Zero Fill for Missing Pod Phases, Change Namespace Dimension to Kubernetes namespace, as it might be confused with metrics namespace in Metrics Explorer

r-dilip:


source/code/plugin/filter_inventory2mdm.rb
```diff
@@ -56,7 +56,7 @@ class Inventory2MdmFilter < Filter
                         "namespace": "insights.container/pods", 
                         "dimNames": [ 
                         "phase", 
-                        "namespace", 
+                        "Kubernetes namespace",
```
vishiy: why do you want to add 'Kubernetes' ?

vishiy: due to already existing "namespace"?

PR #211, opened 2019-04-04
Fix omsagent crash Error when kube-api returns non-200, send events for HTTP Errors

r-dilip:


source/code/plugin/out_mdm.rb
```diff
@@ -167,10 +168,12 @@ def send_to_mdm(post_body)
           # Not raising exception, as that will cause retries to happen
         elsif !response.code.empty? && response.code.start_with?("4")
           # Log 400 errors and continue
+          ApplicationInsightsUtility.sendCustomEvent("AKSCustomMetricsMDMPostError-#{response.code} ", {})
```
vishiy: Can you remove this ? We don't want to log 403's as they will be too many (even one from every node every 60 mins)

source/code/plugin/out_mdm.rb
```diff
@@ -167,10 +168,12 @@ def send_to_mdm(post_body)
...
           @log.info "Non-retryable HTTPServerException when POSTing Metrics to MDM #{e} Response: #{response}"
         else
           # raise if the response code is non-400
           @log.info "HTTPServerException when POSTing Metrics to MDM #{e} Response: #{response}"
+          ApplicationInsightsUtility.sendCustomEvent("AKSCustomMetricsMDMPostError-#{response.code} ", {})
```
vishiy: the 'else' part can have response.code as empty.

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -194,20 +199,21 @@ def get_metric_records(record, metric_name, metric_value, percentage_metric_valu
...
                     new_es.add(time, filtered_record) if filtered_record
-                } if filtered_records
-              rescue => e
-                router.emit_error_event(tag, time, record, e)
-              end
-            }
+                    } if filtered_records
+                }
+            rescue => e
+                ApplicationInsightsUtility.sendExceptionTelemetry(e.backtrace)
```
vishiy: Not sure this will be useful? How could we query for exceptions from this filter plugin with this approach ? We are entirely relying on the stack to even identify these exceptions (and they could be different/empty every time). What we would need is a custom dimension identifying which plugin/place is the exception about or else append the exception message with plug-in name that would be more useful..

source/code/plugin/out_mdm.rb
r-dilip: we specifically look for 403 in the block above this. So these wont log 403's

source/code/plugin/out_mdm.rb
r-dilip: the idea here is this:
we will only get a finite set of response codes. 400s, 500s, or empty. So we should still be able to filter them out. 
The set of event names again is finite, based on the response code.

source/code/plugin/filter_cadvisor2mdm.rb
r-dilip: I will add a task for this, and address that.

source/code/plugin/out_mdm.rb
vishiy: what is the exception you log in line #167 ? It is inside the 403 check block.

source/code/plugin/out_mdm.rb
vishiy: You should use 'dimension' for any metric rather than using metric name for this.  It is easier to do this way but pretty ugly. This will cause metric creep. Please create a back log. We should review all telemetry. Imagine every api call logging metric like this. Its not good & hard to chart & report.

source/code/plugin/out_mdm.rb
r-dilip: LIne 167 is an exception, which was there before. I removed it now. And Added the response code and error as dimensions with a single fixed custom Event Name "AKSCustomMetricsMDMPostError"

PR #214, opened 2019-04-15
Merge from vishwa/telegraftcp to ci_feature for telegraf changes

vishiy:


installer/conf/td-agent-bit.conf
```diff
@@ -23,10 +23,37 @@
...
 [OUTPUT]
     Name                            oms
     EnableTelemetry                 true
     TelemetryPushIntervalSeconds    300
-    Match                           oms.container.log.*
-    AgentVersion                    ciprod03122019
\ No newline at end of file
+    Match                           oms.container.*
+
+#[OUTPUT]
```
rashmichandrashekar: Should these be removed?

installer/scripts/TelegrafTCPErrorTelemetry.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+countErr=$(grep -iF  "socket_writer" /var/opt/microsoft/docker-cimprov/log/telegraf.log | wc -l | tr -d '\n')
+echo "telegraf,AKS_RESOURCE_ID=${AKS_RESOURCE_ID} telegrafTCPWriteErrorCountTotal=${countErr}i"
```
rashmichandrashekar: Don't we need AcsResourceName for aks-engine clusters?

source/code/go/src/plugins/oms.go
```diff
@@ -207,6 +250,174 @@ func updateKubeSystemContainerIDs() {
...
+			Computer: 	Computer, //this is the collection agent's computer name, not necessarily to which computer the metric applies to
+		}
+
+		//Log ("la metric:%v", laMetric)
+		laMetrics = append(laMetrics, &laMetric)
+	}
+	return laMetrics, nil
+}
+
+//send metrics from Telegraf to LA. 1) Translate telegraf timeseries to LA metric(s) 2) Send it to LA as 'ContainerMetrics' fixed type
```
rashmichandrashekar: Should this be 'InsightsMetrics' type?

installer/conf/td-agent-bit.conf
vishiy: This is the one that gets oms agent container logs for telemetry. So we need it.

installer/scripts/TelegrafTCPErrorTelemetry.sh
vishiy: I am just adding this as a tag as i cannot emit a metric without a tag. this will be overridden.

source/code/go/src/plugins/oms.go
vishiy: yup. will change the comment.

PR #218, opened 2019-05-28
Fix Unscheduled Pod bug, remove excess telemetry

r-dilip:


source/code/plugin/out_mdm.rb
```diff
@@ -156,7 +156,7 @@ def send_to_mdm(post_body)
         response = @http_client.request(request)
         response.value # this throws for non 200 HTTP response code
         @log.info "HTTP Post Response Code : #{response.code}"
-        ApplicationInsightsUtility.sendCustomEvent("AKSCustomMetricsMDMSendSuccessful", {})
+        #ApplicationInsightsUtility.sendCustomEvent("AKSCustomMetricsMDMSendSuccessful", {})
```
r-dilip: TODO: Check that there are no dashboard charts depending on this event.

vishiy: Right now, this metric is the only way to figure how many clusters & nodes are sending metrics to MDM. If you disable this, we will lose that.

source/code/plugin/out_mdm.rb
```diff
@@ -156,7 +157,11 @@ def send_to_mdm(post_body)
         response = @http_client.request(request)
         response.value # this throws for non 200 HTTP response code
         @log.info "HTTP Post Response Code : #{response.code}"
-        ApplicationInsightsUtility.sendCustomEvent("AKSCustomMetricsMDMSendSuccessful", {})
+        if @last_telemetry_sent_time.nil? || @last_telemetry_sent_time + 60 * 60 < Time.now
```
vishiy: should this be '>' ?

PR #222, opened 2019-06-06
Fix scenario where controller name is empty

r-dilip:


source/code/plugin/filter_inventory2mdm.rb
```diff
@@ -176,6 +176,10 @@ def process_pod_inventory_records(es)
                     podControllerNameDimValue = record['DataItems'][0]['ControllerName']
                     podNodeDimValue = record['DataItems'][0]['Computer']
 
+                    if podControllerNameDimValue.nil? || podControllerNameDimValue.empty?
```
vishiy: please add a backlog bug for 400 telemetry.

PR #226, opened 2019-06-13
fix fluent bit tuning for perf run

vishiy:


installer/conf/td-agent-bit.conf
```diff
@@ -1,5 +1,5 @@
 [SERVICE]
-    Flush         10
+    Flush         15
```
r-dilip: was 10 still resulting in high CPU?

vishiy: yes. 10 is a little high. we were 30. lets start with 15 and we can further reduce in future.

PR #233, opened 2019-07-03
Fix Region Spacing bug for backdoor agent onboarding scenarios

r-dilip:


source/code/plugin/out_mdm.rb
```diff
@@ -67,7 +67,9 @@ def start
         return
       end
 
-      @@post_request_url = @@post_request_url_template % {aks_region: aks_region, aks_resource_id: aks_resource_id}
+      aks_region = aks_region.gsub(" ","")
+
+      @@post_request_url = @@post_request_url_template % {aks_region: aks_region), aks_resource_id: aks_resource_id}
```
rashmichandrashekar: Is this url case insensitive?

PR #234, opened 2019-07-08
Add Logs Size To Telemetry

r-dilip:


source/code/go/src/plugins/oms.go
```diff
@@ -476,6 +477,8 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 			Name:                  stringMap["Name"],
 		}
 
+		FlushedRecordsSize += float64(unsafe.Sizeof(stringMap["LogEntry"]))
```
vishiy: Can u please check if this works ? Because sizeof doc says the below -
"The size does not include any memory possibly referenced by x" . Not sure how stringMap is implemented. Can you check with logs of known sizes if it is correct?

PR #235, opened 2019-07-08
Collect container last state

daweim0:
This PR adds the ContainerLastStatus field to KubePodInventory. It writes the fields lastState, reason, startedAt, and finishedAt from the `kubectl describe pod` command into ContainerLastStatus (as a Json object). If a pod has no last state then `{}` is written into the ContainerLastStatus field.

(tagging @ganga1980)

### DIFF
--- a/source/code/plugin/in_kube_podinventory.rb
+++ b/source/code/plugin/in_kube_podinventory.rb
@@ -317,6 +317,19 @@ module Fluent
                   record["ContainerStatusReason"] = containerStatus[containerStatus.keys[0]]["reason"]
                 end
               end
+
+              # Record the last state of the container. This may have information on why a container was killed.
+              if container["lastState"].keys.length > 0
+                lastStateName = container["lastState"].keys[0]
+                record["ContainerLastStatus"] = Hash.new
+                record["ContainerLastStatus"]["lastState"] = lastStateName  # get the name of the last state (ex: terminated)
+                record["ContainerLastStatus"]["reason"] = container["lastState"][lastStateName]["reason"]  # (ex: OOMKilled)
+                record["ContainerLastStatus"]["startedAt"] = container["lastState"][lastStateName]["startedAt"]  # (ex: 2019-07-02T14:58:51Z)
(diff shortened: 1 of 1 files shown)


source/code/plugin/in_kube_podinventory.rb
```diff
@@ -317,6 +317,19 @@ def parse_and_emit_records(podInventory, serviceList)
                   record["ContainerStatusReason"] = containerStatus[containerStatus.keys[0]]["reason"]
                 end
               end
+
+              # Record the last state of the container. This may have information on why a container was killed.
+              if container["lastState"].keys.length > 0
```
vishiy: can u check if "lastState" always exist ? What do you see for a newly created deployment for example where there is no 'last' state?

source/code/plugin/in_kube_podinventory.rb
```diff
@@ -317,6 +317,19 @@ def parse_and_emit_records(podInventory, serviceList)
...
+                lastStateName = container["lastState"].keys[0]
+                record["ContainerLastStatus"] = Hash.new
```
vishiy: Again, you are assuming all these properties exist all the time. Is that true ? Can you please add check for existence ?

source/code/plugin/in_kube_podinventory.rb
```diff
@@ -317,6 +317,19 @@ def parse_and_emit_records(podInventory, serviceList)
                   record["ContainerStatusReason"] = containerStatus[containerStatus.keys[0]]["reason"]
                 end
               end
+
+              # Record the last state of the container. This may have information on why a container was killed.
+              if container["lastState"].keys.length > 0
+                lastStateName = container["lastState"].keys[0]
```
vishiy: keys[0] exists all the time ?

source/code/plugin/in_kube_podinventory.rb
vishiy: For safer side, can you put this entire block under rescue so if this fails it causes no impact ?

daweim0: Yes.

PR #247, opened 2019-08-06
Rashmi/fix prom telemetry

rashmichandrashekar:
Changes to fix prometheus telemetry
Changes to replace the placeholders in td-agent-bit conf file

### DIFF
--- a/source/code/plugin/in_kube_podinventory.rb
+++ b/source/code/plugin/in_kube_podinventory.rb
@@ -285,7 +285,6 @@ module Fluent
                 controllerValue = controllerData[record["ControllerKind"]]
                 controllerData[record["ControllerKind"]] += 1
               end
-              $log.warn "controller data: #{controllerData}"
             end
           end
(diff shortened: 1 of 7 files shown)


source/code/plugin/in_kube_podinventory.rb
```diff
@@ -277,6 +278,13 @@ def parse_and_emit_records(podInventory, serviceList)
             record["ControllerName"] = items["metadata"]["ownerReferences"][0]["name"]
             if telemetryFlush == true
               controllerSet.add(record["ControllerKind"] + record["ControllerName"])
+              #Adding controller kind to telemetry ro information about customer workload
+              if (controllerData[record["ControllerKind"]].nil?)
```
vishiy: check for not nil

rashmichandrashekar: There's a check for ownerreferences that i forgot to mention. Its hidden. So resolving this.

PR #252, opened 2019-08-16
Gangams/enable ai telemetry in mc

ganga1980:
updated instance of telemetry client in go and ruby plugins to consume ai telemetry endpoint


source/code/plugin/ApplicationInsightsUtility.rb
```diff
@@ -70,7 +72,10 @@ def initializeUtility()
           @@Tc = ApplicationInsights::TelemetryClient.new
         elsif !encodedAppInsightsKey.nil?
           decodedAppInsightsKey = Base64.decode64(encodedAppInsightsKey)
-          @@Tc = ApplicationInsights::TelemetryClient.new decodedAppInsightsKey
+          telemetrySynchronousSender = ApplicationInsights::Channel::SynchronousSender.new appInsightsEndpoint
```
vishiy: Looks good. (default behavior is sync). 
Based on the OMS-Docker feedbacj, this PR will have to change (to set the endpoint address only if its overridden)

PR #261, opened 2019-09-04
Fixing null check out_mdm bug, tomlparser bug, exposing Replica Set service name as an ENV variable

r-dilip:


installer/conf/container.conf
```diff
@@ -76,7 +76,7 @@
   heartbeat_type tcp
 
   <server>
-    host healthmodel-replicaset-service.kube-system
+    host "#{ENV['REPLICASET_SERVICE_ENDPOINT']}"
```
rashmichandrashekar: Is this expected to be passed in the yaml?

r-dilip: https://github.com/microsoft/OMS-docker/pull/303 is the PR for the yaml change(s) along with the clusterrole name change revert

installer/scripts/tomlparser.rb
```diff
@@ -121,10 +122,12 @@ def populateSettingValuesFromConfigMap(parsedConfig)
   end
 
   begin
-    if !parsedConfig.nil?  && !parsedConfig[:agent_settings][:health_model].nil? && !parsedConfig[:agent_settings][:health_model][:enabled].nil?
+    if !parsedConfig.nil? && !parsedConfig[:agent_settings].nil? && !parsedConfig[:agent_settings][:health_model].nil? && !parsedConfig[:agent_settings][:health_model][:enabled].nil?
         @enable_health_model = parsedConfig[:agent_settings][:health_model][:enabled]
-        puts "enable_health_model = #{@enable_health_model}"
+    else
+        @enable_health_model = false
```
vishiy: Given the fact you will be working on the health config for this sprint, it would make sense to move all health model to config parsing to its own code, so we can nicely keep the sub-sections (logcollection, prometheusmetriccollection, healththresholds) seperate and each idempotent. Its okay for duplicate a few lines o code as it greatly improves testability and also created less impact for customers when they err on a sub-section. Please move this to its own as part of upcoming health config work.

r-dilip: I will address this as part of config PR

PR #264, opened 2019-09-12
Gangams/add telemetry hybrid

ganga1980:
This PR has the following changes 
 
   -  Added the ProviderID, KernelVersion and osImage as dimensions to node telemetry
   -  ProviderID field to Kube Node Inventory Table which will be used to show the cloud type in Ux

Note: 
1. ID of the node assigned by the cloud provider in the format: <ProviderName>://<ProviderSpecificNodeID>
2. still I am discussing with Azure stack team about detecting Azure stack since both Azure and Azurestack same cloud provider name (i.e.azure).
3. OnPrem K8s doesnt have the ProviderID so in the code ProviderID is hardcoded as "onprem".

### DIFF
--- a/source/code/plugin/in_kube_nodes.rb
+++ b/source/code/plugin/in_kube_nodes.rb
@@ -87,12 +87,12 @@ module Fluent
 
               if !items["spec"]["providerID"].nil? && !items["spec"]["providerID"].empty?
                 if File.file?(@@AzStackCloudFileName) # existence of this file indicates agent running on azstack
-                  record["ProviderID"] = "azurestack"
+                  record["KubernetesProviderID"] = "azurestack"
                 else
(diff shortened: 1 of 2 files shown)


source/code/plugin/in_kube_nodes.rb
```diff
@@ -84,6 +84,13 @@ def enumerate
               record["Labels"] = [items["metadata"]["labels"]]
               record["Status"] = ""
 
+              if !items["spec"]["providerID"].nil? && !items["spec"]["providerID"].empty?
+                record["ProviderID"] = items["spec"]["providerID"]
+              else
+                record["ProviderID"] = "onprem"
```
vishiy: anything in 'record' goes to customer workspace (not telemetry). Is that the intention? If so, did you also add the 'field' 'ProviderID' to KubeNode type ?

PR #266, opened 2019-09-26
Fix the Dupe Perf Data Issue from the DaemonSet

r-dilip:


r-dilip:
this change will leave a log line in the omsagent.log like:
**[warn]: no match for kubehealth.DaemonSet.Node** 
every so often (I observed it exponentially backing off, i.e every 1min, then every 2min, then every 4min, ....)

installer/conf/container-health.conf
```diff
@@ -68,7 +68,7 @@
 </match>
 
 
-<match oms.api.KubeHealth.DaemonSet**>
+<match kubehealth.DaemonSet**>
```
vishiy: when you go to IP discovery route (ignite changes i guess) change the below host in out_forward to IP as well)

installer/conf/kube.conf
```diff
@@ -55,7 +55,7 @@
 #Kubernetes health
 <source>
   type kubehealth
- tag oms.api.KubeHealth.ReplicaSet
+ tag kubehealth.ReplicaSet
```
vishiy: kube.conf changes requires change in omsdocker/.yaml (for aks-engine) and the chart (for AKS). Please ensure you change both as part of roll out.

r-dilip: Yes. that is the plan

PR #268, opened 2019-09-30
PR for 1. Container Memory CPU monitor 2. Configuration for Node Conditions 3. Fixed Type Changes 4. Use Env variable, and health_forward (that handles network errors at init) 5. Unit Tests

r-dilip:


r-dilip:
A couple of items not addressed here
1. Split out the config parsing for health. (I will remove it altogether, since the flag will be ignored) 
2. A minor change to not set state as warning if limits are not set for 'kube-system' containers

source/code/plugin/filter_cadvisor_health_node.rb
```diff
@@ -231,7 +142,7 @@ def process_node_cpu_record(record, metric_value)
                 health_record[HealthMonitorRecordFields::MONITOR_ID] = monitor_id
                 health_record[HealthMonitorRecordFields::MONITOR_INSTANCE_ID] = monitor_instance_id
                 health_record[HealthMonitorRecordFields::DETAILS] = health_monitor_record
-                health_record[HealthMonitorRecordFields::AGENT_COLLECTION_TIME] =  time_now
+                health_record[HealthMonitorRecordFields::TIME_GENERATED] =  time_now
```
vishiy: you saying this is not used. please remove.

installer/conf/kube.conf
```diff
@@ -55,7 +55,7 @@
 #Kubernetes health
 <source>
   type kubehealth
- tag oms.api.KubeHealth.ReplicaSet
+ tag kubehealth.ReplicaSet
```
vishiy: please get port # from env for listener

source/code/plugin/filter_cadvisor_health_node.rb
```diff
@@ -263,7 +178,7 @@ def process_node_memory_record(record, metric_value)
...
```
vishiy: +1

source/code/plugin/filter_cadvisor_health_container.rb
```diff
@@ -5,66 +5,57 @@ module Fluent
     require 'logger'
     require 'json'
     require_relative 'oms_common'
-    require_relative 'HealthMonitorUtils'
-    require_relative 'HealthMonitorState'
     require_relative "ApplicationInsightsUtility"
+    Dir[File.join(__dir__, './health', '*.rb')].each { |file| require file }
```
rashmichandrashekar: I remember we had problems with this earlier. Are we still planning to keep the same bevaiour?

PR #270, opened 2019-10-07
Send agg monitor signal on details change

r-dilip:


source/code/plugin/health/health_monitor_state.rb
```diff
@@ -142,6 +143,10 @@ def update_state(monitor, #UnitMonitor/AggregateMonitor
                     @@first_record_sent[monitor_instance_id] = true
                     health_monitor_instance_state.should_send = true
                     set_state(monitor_instance_id, health_monitor_instance_state)
+                elsif agg_monitor_details_changed?(is_aggregate_monitor, deleted_record, health_monitor_instance_state.prev_records[0])
```
vishiy: Not sure what will happen if there are no prev_records ? [0] will always exist ?

r-dilip: Will never hit a scenario where the records will be empty. The else section of this code ensures that

PR #273, opened 2019-10-09
Bug fixes for unscheduled pods/agg_monitor_details_change deep comparison

r-dilip:


rashmichandrashekar:
Could you pls make these changes in ci_feature so that we dont have to cherry pick multiple fixes?

PR #275, opened 2019-10-10
Fix to use declaration and assignment instead of assignment

rashmichandrashekar:


rashmichandrashekar:
Merging without approval since it is a variable declaration  change

PR #286, opened 2019-10-14
Remove limit set check for customer workloads

r-dilip:


source/code/plugin/health/health_container_cpu_memory_aggregator.rb
```diff
@@ -219,29 +219,25 @@ def get_records
...
 
-                if size == 1
-                    state_index = 0
-                else
-                    state_threshold = config['StateThresholdPercentage'].to_f
-                    count = ((state_threshold*size)/100).ceil
-                    state_index = size - count
-                end
-                v["state"] = v["records"][state_index]["state"]
+            if size == 1
```
vishiy: so this looks like ptile calculation using threshold. Isn't this common for many monitors (especially workloads & controllers). If so why is this not a utility ?

PR #287, opened 2019-10-18
Fix the revert that removed explicit require

r-dilip:


source/code/plugin/health/agg_monitor_id_labels.rb
```diff
@@ -1,3 +1,5 @@
+require_relative 'health_model_constants'
```
vishiy: Did u check in other places/files for this issue?

vishiy: Also will cannot be rescuing the 1011 tag again. Guess deployment has gone past canary...

r-dilip: Yes. This was the only file. 
I am planning to use a different tag and planning to hot fix. We have not seen this issue in AKS clusters.

PR #289, opened 2019-10-24
Bug Fixes for exceptions in telemetry, remove limit set check

r-dilip:


source/code/plugin/health/health_container_cpu_memory_aggregator.rb
```diff
@@ -163,11 +175,24 @@ def get_records
             container_cpu_memory_records = []
 
             @cpu_records.each{|resource_key, record|
+
+                cpu_limit_mc = 1.0
+                if record["limit"].is_a?(Numeric)
```
rashmichandrashekar: what happens if record["limit"] is nil?

source/code/plugin/health/health_container_cpu_memory_aggregator.rb
```diff
@@ -163,11 +175,24 @@ def get_records
...
+                    cpu_limit_mc = record["limit"]/1000000.to_f
+                else
+                    @log.info "CPU Limit is not a number #{record['limit']}"
+                    if !@@limit_is_array_event_sent
```
rashmichandrashekar: Is this an aggregation to send this event only once?
If so, we will send this event either for container/node for cpu/memory.. is that what is intended?

source/code/plugin/health/health_container_cpu_memory_aggregator.rb
r-dilip: it will still use the value as 1.0

source/code/plugin/health/health_container_cpu_memory_aggregator.rb
r-dilip: I slightly changed the logic to send it once per workload per cluster.

source/code/plugin/health/health_kubernetes_resources.rb
```diff
@@ -200,6 +167,25 @@ def get_workload_kind(pod)
             end
         end
 
+        private
+        def get_replica_set_owner_ref(controller_name)
+            if @workload_name_cache.key?(controller_name)
```
rashmichandrashekar: we should add begin and rescue for this code block

source/code/plugin/health/health_kubernetes_resources.rb
```diff
@@ -200,6 +167,25 @@ def get_workload_kind(pod)
             end
         end
 
+        private
+        def get_replica_set_owner_ref(controller_name)
```
rashmichandrashekar: We should add exception handling for this block

source/code/plugin/health/health_container_cpu_memory_aggregator.rb
rashmichandrashekar: wont this throw an exception in that case?

PR #294, opened 2019-11-05
Fix for Nodes Aspect not showing up in draft cluster

r-dilip:


source/code/plugin/health/health_model_definition_parser.rb
```diff
@@ -29,6 +29,7 @@ def parse_file
                 labels = entry['labels']  if entry['labels']
                 aggregation_algorithm = entry['aggregation_algorithm'] if entry['aggregation_algorithm']
                 aggregation_algorithm_params = entry['aggregation_algorithm_params'] if entry['aggregation_algorithm_params']
+                default_parent_monitor_id = entry['default_parent_monitor_id'] if entry['default_parent_monitor_id']
```
r-dilip: This is defined in health_model_definition.json.
When I build the hierarchy, i stop building when a parent id for a monitor is null. (Usually happens only for the top level cluster monitor). In this case, since this was not initialized, and it is a fall through case, (doesnt happen in AKS, since one of the conditions is met) , the parent monitor id was being returned as nil, and the hierarchy building is stopped. So there is no (node pool --> all_nodes ) monitors.

PR #301, opened 2019-11-14
Rashmi/http leak fixes

rashmichandrashekar:
PR to merge http client connection close

### DIFF
--- a/source/code/plugin/DockerApiClient.rb
+++ b/source/code/plugin/DockerApiClient.rb
@@ -3,7 +3,7 @@
 
 class DockerApiClient
   require "socket"
-  require 'yajl/json_gem'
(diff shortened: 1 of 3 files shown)


source/code/plugin/DockerApiClient.rb
```diff
@@ -49,6 +48,10 @@ def getResponse(request, isMultiJson, isVersion)
           ApplicationInsightsUtility.sendExceptionTelemetry(errorStr)
         end
         return nil
+      ensure
+        ## REMOVE LOG BEFORE MERGE
+        $log.warn "Closing docker socket connection"
+        socket.close
```
vishiy: nil check before close ?

PR #323, opened 2020-01-16
For ARO, stop collecting inventory of master and infra

ganga1980:
In ARO, in v3, monitoring agent (both replicaset and deamonset) deployed only to compute nodes. Since the replicaset gets the inventory such as nodes, pods and events etc .. from Kube API. The monitoring data collected by the replicaset agent has the inventory of all the nodes (master, infra and compute), but where as the monitoring data (such as perf and logs ) collected by the daemonset only for the compute nodes. So, this causing incomplete data in Azure Monitoring Ux, since Ux integrates the data from both daemonset and replicaset for the visualization.  And also this partial data related to master and infra nodes will cost the customer. This PR addresses this issue by stop collecting inventory of master and infra nodes.

### DIFF
--- a/source/code/plugin/KubernetesApiClient.rb
+++ b/source/code/plugin/KubernetesApiClient.rb
@@ -255,7 +255,7 @@ class KubernetesApiClient
     def getWindowsNodes
       winNodes = []
       begin
-        resourceUri = isAROCluster ? "nodes?labelSelector=node-role.kubernetes.io/compute%3Dtrue": "nodes"
+        resourceUri = isAROCluster ? "nodes?labelSelector=node-role.kubernetes.io%2Fcompute%3Dtrue": "nodes"
         nodeInventory = JSON.parse(getKubeResourceInfo(resourceUri).body)
         @Log.info "KubernetesAPIClient::getWindowsNodes : Got nodes from kube api"
         # Resetting the windows node cache
--- a/source/code/plugin/KubernetesApiClient.rb
+++ b/source/code/plugin/KubernetesApiClient.rb
@@ -18,7 +18,7 @@ class KubernetesApiClient
   @@ClusterName = nil
   @@ClusterId = nil
   @@IsNodeMaster = nil
-  @@IsAROCluster = nil
+  @@IsAROV3Cluster = nil
   #@@IsValidRunningNode = nil
   #@@IsLinuxCluster = nil
   @@KubeSystemNamespace = "kube-system"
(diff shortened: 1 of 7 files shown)


source/code/plugin/KubernetesApiClient.rb
```diff
@@ -152,6 +153,21 @@ def getClusterId
       return @@ClusterId
     end
 
+    def isAROCluster
+      return @@IsAROCluster if !@@IsAROCluster.nil?
+      @@IsAROCluster = false
+      begin
+        cluster = getClusterId
+        if cluster && !cluster.nil? && !cluster.empty? &&
```
r-dilip: Nit: just use !cluster.nil? && !cluster.empty?

source/code/plugin/KubernetesApiClient.rb
```diff
@@ -357,6 +374,14 @@ def getContainerResourceRequestsAndLimits(metricJSON, metricCategory, metricName
             podUid = pod["metadata"]["uid"]
           end
 
+          # For ARO, skip the pods scheduled on to master or infra nodes to ingest
+          if isAROCluster && !pod["spec"]["nodeName"].nil? &&
```
r-dilip: wrap this is a begin exception end block, or check for pod["spec"].nil? as well.

source/code/plugin/KubernetesApiClient.rb
```diff
@@ -357,6 +373,14 @@ def getContainerResourceRequestsAndLimits(metricJSON, metricCategory, metricName
             podUid = pod["metadata"]["uid"]
           end
 
+          # For ARO, skip the pods scheduled on to master or infra nodes to ingest
+          if isAROCluster && !pod["spec"].nil? && !pod["spec"]["nodeName"].nil? &&
+            ( pod["spec"]["nodeName"].downcase.include?("infra-") ||
```
vishiy: please use starts_with instead if include. Include checks for substrings. and if a node has 'infra-' or 'master-' anywhere in its name it will be excluded when u use include

source/code/plugin/filter_cadvisor2mdm.rb
```diff
@@ -145,7 +145,12 @@ def ensure_cpu_memory_capacity_set
             end
 
             begin
-                nodeInventory = JSON.parse(KubernetesApiClient.getKubeResourceInfo("nodes?fieldSelector=metadata.name%3D#{@@hostName}").body)
+                resourceUri = "nodes?fieldSelector=metadata.name%3D#{@@hostName}"
+                # For ARO, filter out all other node roles other than compute
+                if KubernetesApiClient.isAROCluster
+                    resourceUri = resourceUri + "&labelSelector=node-role.kubernetes.io/compute%3Dtrue"
```
vishiy: so when u pass in fieldselector & label selector, is it an AND between those two?

source/code/plugin/in_kube_events.rb
```diff
@@ -101,6 +101,14 @@ def parse_and_emit_records(events, eventQueryState, newEventQueryState, batchTim
           if !eventQueryState.empty? && eventQueryState.include?(eventId)
             next
           end
+
+          nodeName = items["source"].key?("host") ? items["source"]["host"] : (OMS::Common.get_hostname)
+          # For ARO, drop the master and infra node sourced events to ingest
+          if KubernetesApiClient.isAROCluster && !nodeName.nil? && !nodeName.empty? &&
+             ( nodeName.downcase.include?("infra-") || nodeName.downcase.include?("master-") )
```
vishiy: starts_with instead of include

source/code/plugin/in_kube_podinventory.rb
```diff
@@ -265,6 +265,13 @@ def parse_and_emit_records(podInventory, serviceList, batchTime = Time.utc.iso86
           record["Name"] = items["metadata"]["name"]
           podNameSpace = items["metadata"]["namespace"]
 
+          # For ARO, skip the pods scheduled on to master or infra nodes
+          if KubernetesApiClient.isAROCluster && !items["spec"].nil? && !items["spec"]["nodeName"].nil? &&
+             ( items["spec"]["nodeName"].downcase.include?("infra-") ||
```
vishiy: starts_with

source/code/plugin/filter_cadvisor2mdm.rb
ganga1980: > so when u pass in fieldselector & label selector, is it an AND between those two?

yes, that's right.

PR #328, opened 2020-02-12
changes to use msi if service principal does not exist

rashmichandrashekar:
changes to use msi if service principal does not exist

### DIFF
--- a/source/code/plugin/in_kube_events.rb
+++ b/source/code/plugin/in_kube_events.rb
@@ -4,7 +4,6 @@
 module Fluent
   class Kube_Event_Input < Input
     Plugin.register_input("kubeevents", self)
-
     @@KubeEventsStateFile = "/var/opt/microsoft/docker-cimprov/state/KubeEventQueryState.yaml"
 
     def initialize
@@ -20,6 +19,9 @@ module Fluent
 
       # 30000 events account to approximately 5MB
       @EVENTS_CHUNK_SIZE = 30000
+
+      # Initializing events count for telemetry
+      @eventsCount = 0
     end
 
     config_param :run_interval, :time, :default => 60
@@ -55,6 +57,7 @@ module Fluent
         batchTime = currentTime.utc.iso8601
         eventQueryState = getEventQueryState
         newEventQueryState = []
+        @eventsCount = 0
 
         # Initializing continuation token to nil
         continuationToken = nil
@@ -80,6 +83,13 @@ module Fluent
         # Setting this to nil so that we dont hold memory until GC kicks in
         eventList = nil
         writeEventQueryState(newEventQueryState)
+
+        # Flush AppInsights telemetry once all the processing is done, only if the number of events flushed is greater than 0
+        if (@eventsCount > 0)
+          telemetryProperties = {}
+          ApplicationInsightsUtility.sendMetricTelemetry("EventCount", @eventsCount, {})
+        end
+
       rescue => errorStr
         $log.warn "in_kube_events::enumerate:Failed in enumerate: #{errorStr}"
         $log.debug_backtrace(errorStr.backtrace)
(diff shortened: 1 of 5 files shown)


source/code/plugin/out_mdm.rb
```diff
@@ -12,27 +12,36 @@ def initialize
...
 
       @@token_resource_url = "https://monitoring.azure.com/"
       @@grant_type = "client_credentials"
       @@azure_json_path = "/etc/kubernetes/host/azure.json"
       @@post_request_url_template = "https://%{aks_region}.monitoring.azure.com%{aks_resource_id}/metrics"
-      @@token_url_template = "https://login.microsoftonline.com/%{tenant_id}/oauth2/token"
+      @@aad_token_url_template = "https://login.microsoftonline.com/%{tenant_id}/oauth2/token"
+
+      # msiEndpoint is the well known endpoint for getting MSI authentications tokens
+      @@msi_endpoint_template = "http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&client_id=%{user_assigned_client_id}&resource=%{resource}"
```
vishiy: egress needs to be opened up ?

r-dilip: Use a constant and use it here

source/code/plugin/in_kube_events.rb
```diff
@@ -20,6 +19,12 @@ def initialize
 
       # 30000 events account to approximately 5MB
       @EVENTS_CHUNK_SIZE = 30000
+
+      # Initializing events count for telemetry
+      @eventsCount = 0
+
+      # Initilize enable/disable normal event collection
+      @collectAllKubeEvents = false
```
joby-thomas: By default we are collecting all events today right? Doesn't it make sense to initialize it as true and mark false when settings is enabled?

source/code/plugin/out_mdm.rb
```diff
@@ -57,51 +66,102 @@ def start
...
+          end
+        rescue => err
+          @log.info "Exception in get_access_token: #{err}"
+          if (retries < 2)
+            retries += 1
+            @log.info "Retrying request to get token - retry number: #{retries}"
+            sleep(retries)
+            retry
+          else
+          @get_access_token_backoff_expiry = Time.now + 29 * 60
```
r-dilip: Like we discussed: Since the token is valid for an hour, we should try and refresh it every hour. we are basically doubling the number of token refreshes across entire AKS, and it all adds up

source/code/plugin/out_mdm.rb
r-dilip: when I did the spike, this was accessible out of the box with standard ports.

source/code/plugin/out_mdm.rb
```diff
@@ -57,51 +66,102 @@ def start
...
+            @token_expiry_time = Time.now + 29 * 60 # set the expiry time to be ~thirty minutes from current time
+            @cached_access_token = parsed_json["access_token"]
+          @log.info "Successfully got access token"
+          end
+        rescue => err
+          @log.info "Exception in get_access_token: #{err}"
+          if (retries < 2)
+            retries += 1
+            @log.info "Retrying request to get token - retry number: #{retries}"
+            sleep(retries)
```
joby-thomas: intention here to sleep 1 second, 2 seconds and 3 second before retries?

source/code/plugin/out_mdm.rb
rashmichandrashekar: checking with Saurya regarding this

source/code/plugin/in_kube_events.rb
rashmichandrashekar: By default we collect only kube events with level warning. So the default value in this case would be false.

source/code/plugin/out_mdm.rb
rashmichandrashekar: As discussed offline the url is defined as a constant already. So leaving this as is.

source/code/plugin/out_mdm.rb
rashmichandrashekar: Yes. it is exponential backoff.

source/code/plugin/out_mdm.rb
rashmichandrashekar: I have confirmed with the imds team that this wouldnt be a problem. so leaving it to 30 minutes as this would reduce the loss of data in case there are intermittent issues.

PR #334, opened 2020-02-27
Gangams/add support cri runtime

ganga1980:
updates related to supporting CRI compatible run-times. For k8s clusters, which have docker, old code will be used else kubelet APIs will be used.


ganga1980:
since made another pr with container specific envvars hence closing this.

PR #337, opened 2020-03-08
Gangams/add support cri runtime docker env

ganga1980:
updates related to supporting CRI compatible run-times. For k8s clusters, which have docker, old code will be used else kubelet APIs will be used. For the container environment vars,  read from  /proc/#{cGroupPid}/environ to keep the existing behavior of collecting kubernetes and docker envvars.


installer/conf/telegraf.conf
```diff
@@ -532,7 +532,8 @@
   name_prefix="container.azm.ms/"
   ## An array of urls to scrape metrics from.
   urls = ["$CADVISOR_METRICS_URL"]
-  fieldpass = ["kubelet_docker_operations", "kubelet_docker_operations_errors"]
+  ## Include "$KUBELET_RUNTIME_OPERATIONS_TOTAL_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_TOTAL_METRIC" when we add for support for 1.18
+  fieldpass = ["$KUBELET_RUNTIME_OPERATIONS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_METRIC"]
```
vishiy: I see that you don't set these env for non docker clusters in oms-docker PR. Which means we will be passing everything (as pass list is empty) - meaning we will collect all cadvisor metrics exposed @ /metrics.

source/code/go/src/plugins/oms.go
```diff
@@ -734,9 +741,22 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 
 		stringMap := make(map[string]string)
 
-		stringMap["LogEntry"] = ToString(record["log"])
+		logEntry := ToString(record["log"])
+		logEntryTimeStamp := ToString(record["time"])
+		if isDockerContainerRuntimeEngine == false {
```
vishiy: Please check for 'non-docker' and switch/if as appropriate to minimize impact. Leave/assume that default runtime is docker/moby. So this way, if your container runtime detection logic fails, only containerd clusters (which are very low count as of now) are impacted and not all of the clusters we monitor.

source/code/go/src/plugins/oms.go
```diff
@@ -734,9 +741,22 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+			//CRI compatiable runtimes uses the CRIO parser which has logtag followed space and then log line
+			// so trimming the log tag (P or F) 
+			logEntry = strings.TrimSpace(logEntry)
+			if strings.HasPrefix(logEntry, "P ") {
+			   logEntry = strings.TrimPrefix(logEntry, "P ")
```
vishiy: This will be an expensive operation to trim on every single logline. So basically we were blindly flushing the logline before and now we are processing & trimming every single log line. Are you sure that this ("P ") is passed-in to us in the same logEntry field ?

source/code/go/src/plugins/oms.go
```diff
@@ -958,6 +978,13 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 		Log("ResourceID=%s", ResourceID)
 		Log("ResourceName=%s", ResourceName)
 	}
+	
+	containerRuntime = os.Getenv(ContainerRuntimeEnv)
+	Log("Container Runtime engine %s", containerRuntime)
+	isDockerContainerRuntimeEngine = false
+	if strings.EqualFold(containerRuntime, "docker") {
```
vishiy: make the comparison opposite (meaning check for containerd, if not default to docker), everywhere.

source/code/go/src/plugins/oms.go
```diff
@@ -958,6 +978,13 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 		Log("ResourceID=%s", ResourceID)
 		Log("ResourceName=%s", ResourceName)
 	}
+	
+	containerRuntime = os.Getenv(ContainerRuntimeEnv)
```
vishiy: init containerRuntime to docker (default). Then get from os.Getenv, and assign it to containerRuntime only if its not empty.

source/code/plugin/ApplicationInsightsUtility.rb
```diff
@@ -108,15 +109,21 @@ def initializeUtility()
...
 
-    def getDockerInfo()
-      dockerInfo = DockerApiClient.dockerInfo
-      if (!dockerInfo.nil? && !dockerInfo.empty?)
-        @@CustomProperties["DockerVersion"] = dockerInfo["Version"]
-        #@@CustomProperties["DockerApiVersion"] = dockerInfo["ApiVersion"]
+    def getContainerRuntimeInfo()
+      containerRuntime = ENV[@@EnvContainerRuntime]
+      if !containerRuntime.nil? && !containerRuntime.empty?
+        @@CustomProperties["ContainerRunTime"] = containerRuntime
```
vishiy: How about containerRunTimeVersion for non docker/moby runtimes ?

source/code/plugin/in_kube_nodes.rb
```diff
@@ -188,7 +192,12 @@ def parse_and_emit_records(nodeInventory, batchTime = Time.utc.iso8601)
             properties["Computer"] = record["Computer"]
             properties["KubeletVersion"] = record["KubeletVersion"]
             properties["OperatingSystem"] = nodeInfo["operatingSystem"]
-            properties["DockerVersion"] = dockerVersion
+            properties["ContainerRuntimeVersion"] = containerRuntimeVersion
```
vishiy: just one thing is enough for telemetry. We will pay more by duplicating the versions in both 'dockerversion' and 'containerruntimeversion' dimensions for docker based clusters (which is our majority as of now). To not break our telemery queries, just put the runtime version in dockerVersion dimension.

source/code/plugin/in_containerinventory.rb
```diff
@@ -190,29 +193,263 @@ def inspectContainer(id, nameMap, clusterCollectEnvironmentVar)
       return containerInstance
     end
 
+    def getContainerInventoryRecords(batchTime, clusterCollectEnvironmentVar)
```
vishiy: This is the same as we do for windows nodes (derive containerinventory from /pods), except that the /pods call is to the kubelet (rather than API server). I assume everything should be the same (except env var part) ??

@rashmichandrashekar - Can you please take a look at this method to check what it is different and if we need to duplicate the parsing code ?

rashmichandrashekar:
I see an empty file with no changes.. Is that needed?

source/code/go/src/plugins/oms.go
```diff
@@ -37,6 +37,9 @@ const ResourceIdEnv = "AKS_RESOURCE_ID"
 //env variable which has ResourceName for NON-AKS
 const ResourceNameEnv = "ACS_RESOURCE_NAME"
 
+//env variable which has container run time name
+const ContainerRuntimeEnv = "CONTAINER_RUN_TIME"
```
r-dilip: Nit: Runtime is usually a single word

source/code/plugin/CAdvisorMetricsAPIClient.rb
```diff
@@ -73,6 +73,11 @@ def getNodeCapacityFromCAdvisor(winNode: nil)
       return getResponse(winNode, relativeUri)
     end
 
+    def getPodsFromCAdvisor(winNode: nil)
```
rashmichandrashekar: Do we need this method? All this method is doing is passing a relative uri.. Why not just call the get response method like the rest of the calls?

source/code/plugin/in_containerinventory.rb
rashmichandrashekar: @ganga1980 - As we discussed earlier i thought the plan was to refactor the code so that we used kube api parsing to be consistent across win container inventory and these changes?

installer/conf/telegraf.conf
ganga1980: as discussed, default will be set these.
export KUBELET_RUNTIME_OPERATIONS_METRIC="kubelet_docker_operations"
export KUBELET_RUNTIME_OPERATIONS_ERRORS_METRIC="kubelet_docker_operations_errors"

source/code/go/src/plugins/oms.go
ganga1980: since the default set as "docker" in main.sh so we are good on this.

source/code/plugin/CAdvisorMetricsAPIClient.rb
ganga1980: as discussed, the current implementation cleaner interms readability so followed the same pattern.

source/code/plugin/in_containerinventory.rb
ganga1980: yes, I have a task to refactor this code on the board. let me see if i can get it tomorrow else handle this separate task.

source/code/plugin/in_kube_nodes.rb
ganga1980: fixed. followed the same pattern as containernodeinventory i.e. DockerVersion fields just version if it is docker else <runtime>://<version>

source/code/plugin/ApplicationInsightsUtility.rb
ganga1980: there is no containerRunTimeVersion  in /pods api call so this has to be surfaced from /nodes KubeAPI call. let me see if there is any way to do this.

source/code/go/src/plugins/oms.go
ganga1980: done.

rashmichandrashekar:
@ganga1980 - Please add corresponding tests in aks-engine as that might help catching issues in newer versions.

source/code/go/src/plugins/oms.go
ganga1980: removed this piece of code since not required any more.

source/code/go/src/plugins/oms.go
ganga1980: since the regex doing this stuff so removed this logic.

source/code/go/src/plugins/oms.go
ganga1980: fixed.

source/code/plugin/in_containerinventory.rb
ganga1980: done refactoring.

installer/conf/azm-containers-parser.conf
```diff
@@ -0,0 +1,18 @@
+[PARSER]
```
rashmichandrashekar: Could you remove the empty file, adding comment here since it doesnt let me comment on that.

source/code/plugin/ApplicationInsightsUtility.rb
```diff
@@ -110,11 +111,17 @@ def initializeUtility()
...
+        if containerRuntime.casecmp("docker") == 0
+          dockerInfo = DockerApiClient.dockerInfo
+          if (!dockerInfo.nil? && !dockerInfo.empty?)
+            @@CustomProperties["DockerVersion"] = dockerInfo["Version"]
+            #@@CustomProperties["DockerApiVersion"] = dockerInfo["ApiVersion"]
```
rashmichandrashekar: Do we need the commented line?

ganga1980: > Do we need the commented line?
This line was commented before. not sure why.

installer/conf/azm-containers-parser.conf
ganga1980: I assume, you are referring azm-containers-parser.conf fie.  If Yes, this is not empty file.

rashmichandrashekar: No, i am referring to the file above this in this PR. The name of the file is ,

rashmichandrashekar: https://github.com/microsoft/Docker-Provider/blob/545a6ff72605be48e42535126c94df1f527b98f4/%2C

ganga1980: > https://github.com/microsoft/Docker-Provider/blob/545a6ff72605be48e42535126c94df1f527b98f4/%2C

> thanks for catching empty file. just deleted it. :)

installer/conf/azm-containers-parser.conf
```diff
@@ -0,0 +1,18 @@
...
+    Time_Key     time
+    Time_Format  %Y-%m-%dT%H:%M:%S.%L
+    Time_Keep    On
+    # Command      |  Decoder | Field | Optional Action
+    # =============|==================|=================
+    Decode_Field_As   escaped    log
+
+[PARSER]
+    # http://rubular.com/r/tjUt3Awgg4
+    Name cri
```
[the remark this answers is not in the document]
ganga1980: > should this be renamed to crio?
> since same logging format for all CRI compatiable runtimes so fluentbit guys also renamed cri from cri-o so cri is the right naming.

source/code/plugin/ApplicationInsightsUtility.rb
ganga1980: We have telemetry captured in NodeCoreCapacity metric via DockerVersion which has runtime and version so we are good with that.  This one only used for the debugging purpose and here we can only either runtime name for non-docker environments or docker version for docker envs.

PR #341, opened 2020-04-13
Enable telemetry for cpu and memory usage in windows for omsagent pods

bragi92:


source/code/plugin/CAdvisorMetricsAPIClient.rb
```diff
@@ -513,7 +548,7 @@ def getContainerMemoryMetricItems(metricJSON, hostName, memoryMetricNameToCollec
               begin
                 # we can only do this much now. Ideally would like to use the docker image repository to find our pods/containers
                 # cadvisor does not have pod/container metadata. so would need more work to cache as pv & use
-                if (podName.downcase.start_with?("omsagent-") && podNamespace.eql?("kube-system") && containerName.downcase.start_with?("omsagent") && metricNametoReturn.eql?("memoryRssBytes"))
+                if (podName.downcase.start_with?("omsagent-") && podNamespace.eql?("kube-system") && containerName.downcase.start_with?("omsagent"))
```
bragi92, line 551: Question: Do we want both workingset and rss values to be collected for the linux agent or just one of them?

bragi92: for now, making it such that we collect memoryrss for linux and memoryworkingset for windows

PR #349, opened 2020-05-05
Change environment variable parsing to to_json instead of to_s

bragi92:
to_s latter adds extra escape characters

### DIFF
--- a/source/code/plugin/kubernetes_container_inventory.rb
+++ b/source/code/plugin/kubernetes_container_inventory.rb
@@ -4,6 +4,7 @@
 class KubernetesContainerInventory
   require "yajl/json_gem"
   require "time"
+  require "json"
   require_relative "omslog"
   require_relative "ApplicationInsightsUtility"
   
@@ -203,7 +204,7 @@ class KubernetesContainerInventory
               envVars = File.read(environFilePath, 200000)             
               if !envVars.nil? && !envVars.empty?
                 envVars = envVars.split("\0")
-                envValueString = envVars.join('')
+                envValueString = envVars.to_json
                 envValueStringLength = envValueString.length
                 $log.info("KubernetesContainerInventory::environment vars filename @ #{environFilePath} envVars size @ #{envValueStringLength}")
                 if envValueStringLength >= 200000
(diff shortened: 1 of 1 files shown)


source/code/plugin/kubernetes_container_inventory.rb
```diff
@@ -203,7 +203,7 @@ def obtainContainerEnvironmentVars(containerId)
               envVars = File.read(environFilePath, 200000)             
               if !envVars.nil? && !envVars.empty?
                 envVars = envVars.split("\0")
-                envValueString = envVars.to_s
+                envValueString = envVars.join('')
```
ganga1980, line 206: Not sure  whats the problem with to_s, in this case, envars will be string array ["a1=b1","a2=b2"] and to_s will be converting to string i.e. '["a1=b1","a2=b2"]' where as join will result "a1=b1" "a2=b2" which is not JSON string at all. This is going to break the JSON parsing functionality completely.   Please help me to understand what is the issue with to_s at all. https://blog.bigbinary.com/2012/06/26/to_str-in-ruby.html

source/code/plugin/kubernetes_container_inventory.rb
```diff
@@ -273,12 +273,12 @@ def obtainWindowsContainerEnvironmentVars(pod, container)
                   value = "secretKeyRef_#{secretName}_#{secretKey}"
                 end
               else
-                value = envVar["valueFrom"].to_s
+                value = envVar["valueFrom"].join('')
```
ganga1980, line 276: similar all other places .. we shouldn't use join('') and keep the JSON representation to break JSON based tools.  It might work in the UX since removed the usage of JSON parser with the custom parser.

source/code/plugin/kubernetes_container_inventory.rb
bragi92: @ganga1980  This is a draft PR - its not ready for review yet. I know the current commited code isn't a JSON string. It was something I was playing around with.

When you try using to_s on an array in ruby you get extra escaped characters.
PS C:\Users\kadubey\Documents\temp> cat .\hello.rb                                                                                                                                                                                                     temp = ["test=#@$#24123das"]

puts temp.to_s;
PS C:\Users\kadubey\Documents\temp> ruby .\hello.rb                                                                                                                                                                                                    ["test=\#@$#24123das"]

The extra \ being added in front of the # is wrong.

bragi92: Updated to use to_json to do the conversion. This does not add extra escape characters and as the function name suggets its also a valid json string which matches the behavior that the current and older agents exhibit.

bragi92: This PR is ready for review now @ganga1980

PR #351, opened 2020-05-18
Gangams/enable proxy support (for non-AKS) remote endpoints

ganga1980:
This PR is enables proxy support for the agent  so that  Linux and Windows agent can be used behind proxy environment.

For external outbound traffic of Non-AKS, OMS, ODS, AI, and onboarding check communications will be routed via Proxy if Proxy configured.

Refer latest design doc https://microsoft.sharepoint.com/:w:/t/Operations_Management_Suite/EWITL1uDq6BPpgv8taRppd4BxKE5H9st5AgRT8nLIAiZsA?e=fIwOVV

### DIFF
--- a/source/code/go/src/plugins/oms.go
+++ b/source/code/go/src/plugins/oms.go
@@ -1001,11 +1001,11 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 			if err != nil {
 				message := fmt.Sprintf("Error Reading omsproxy configuration %s\n", err.Error())
 				Log(message)
+				// if we fail to read proxy configuration, then everything will be fall apart including sending  exceptions to AI
 				SendException(message)
-				time.Sleep(30 * time.Second)
 				log.Fatalln(message)
 			} else {
-				ProxyEndpoint = strings.TrimSpace(string(proxyConfig))		
+				ProxyEndpoint = strings.TrimSpace(string(proxyConfig))
 			}
 		}
 	} else {
--- a/source/code/go/src/plugins/oms.go
+++ b/source/code/go/src/plugins/oms.go
@@ -994,16 +994,15 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 		}
 		// read proxyendpoint if proxy configured
 		ProxyEndpoint = ""
-		proxyConfigPath := pluginConfig["omsproxy_conf_path"]
-		if _, err := os.Stat(proxyConfigPath); err == nil {
-			Log("Reading proxy configuration for Linux from %s", proxyConfigPath)
-			proxyConfig, err := ioutil.ReadFile(proxyConfigPath)
+		proxySecretPath := pluginConfig["omsproxy_secret_path"]
+		if _, err := os.Stat(proxySecretPath); err == nil {
+			Log("Reading proxy configuration for Linux from %s", proxySecretPath)
+			proxyConfig, err := ioutil.ReadFile(proxySecretPath)
 			if err != nil {
 				message := fmt.Sprintf("Error Reading omsproxy configuration %s\n", err.Error())
 				Log(message)
-				// if we fail to read proxy configuration, then everything will be fall apart including sending  exceptions to AI
+				// if we fail to read proxy secret, AI telemetry might not be working as well
 				SendException(message)
-				log.Fatalln(message)
 			} else {
 				ProxyEndpoint = strings.TrimSpace(string(proxyConfig))
 			}
--- a/source/code/go/src/plugins/oms.go
+++ b/source/code/go/src/plugins/oms.go
@@ -1005,7 +1005,7 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 				time.Sleep(30 * time.Second)
 				log.Fatalln(message)
 			} else {
-				ProxyEndpoint = strings.TrimSpace(proxyConfig)		
+				ProxyEndpoint = strings.TrimSpace(string(proxyConfig))		
 			}
 		}
 	} else {
--- a/source/code/go/src/plugins/oms.go
+++ b/source/code/go/src/plugins/oms.go
@@ -101,6 +101,8 @@ var (
 	enrichContainerLogs bool
 	// container runtime engine configured on the kubelet
 	containerRuntime string
+	// Proxy endpoint in format http(s)://<user>:<pwd>@<proxyserver>:<port>
+	ProxyEndpoint string
 )
 
 var (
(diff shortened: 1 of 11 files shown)


source/code/plugin/out_mdm.rb
```diff
@@ -154,7 +144,7 @@ def get_access_token
             parsed_json = JSON.parse(token_response.body)
             @token_expiry_time = Time.now + @@tokenRefreshBackoffInterval * 60 # set the expiry time to be ~thirty minutes from current time
             @cached_access_token = parsed_json["access_token"]
-            @log.info "Successfully got access token"
```
rashmichandrashekar: nit: formatting seems to be off.. could you pls fix it if possible?

source/code/go/src/plugins/oms.go
```diff
@@ -990,11 +992,28 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
...
+		ProxyEndpoint = ""
+		proxyConfigPath := pluginConfig["omsproxy_conf_path"]
+		if _, err := os.Stat(proxyConfigPath); err == nil {
+			Log("Reading proxy configuration for Linux from %s", proxyConfigPath)
+			proxyConfig, err := ioutil.ReadFile(proxyConfigPath)
+			if err != nil {
+				message := fmt.Sprintf("Error Reading omsproxy configuration %s\n", err.Error())
+				Log(message)
+				SendException(message)
+				time.Sleep(30 * time.Second)
```
vishiy, line 1005: Remove sleep. You are just extending the crashlopp back offs by 2 mins for every backoff. Why fatal? If its fatal (meaning nothing works without proxy), how do you expect appinsights telemetry to work without proper proxy being configured ?

source/code/go/src/plugins/utils.go
```diff
@@ -68,7 +69,20 @@ func CreateHTTPClient() {
...
 	tlsConfig.BuildNameToCertificate()
-	transport := &http.Transport{TLSClientConfig: tlsConfig}
+	transport := &http.Transport{TLSClientConfig: tlsConfig}	
+	// set the proxy if the proxy configured
+	if ProxyEndpoint != "" {
+		proxyEndpointUrl, err := url.Parse(ProxyEndpoint)		
+		if err != nil {
+			message := fmt.Sprintf("Error parsing Proxy endpoint %s", err.Error())
+			SendException(message)
+			time.Sleep(30 * time.Second)
```
vishiy, line 79: No sleep. Proxy uri is a free form input coming from the customer. If it is incorrect (or in a format that we dont understand), how will the customer know that the issue is with proxy due to which the container keeps restarting and they need to fix the proxy uri ?

source/code/go/src/plugins/oms.go
ganga1980: Good catch. **AI** may not work if we fail to read the proxy config at all.  If ended up failing to read proxy or bad proxy configuration nothing going to work so sleep is not matter at all. let me remove the seelp.

source/code/go/src/plugins/utils.go
ganga1980: > No sleep. Proxy uri is a free form input coming from the customer. If it is incorrect (or in a format that we dont understand), how will the customer know that the issue is with proxy due to which the container keeps restarting and they need to fix the proxy uri ?
> proxy will be validated in the base agent and also will add the helm client to do the validation. customer should be seeing onboarding failures (helm chart installation and logs of agent pods ) if the proxy endpoint is invalid similar how they see if the workspace is invalid.

source/code/plugin/out_mdm.rb
ganga1980: @rashmichandrashekar , this file reverted. looks like you reviewed before revert?, btw, i didnt see any issue with formatting.

source/code/plugin/ApplicationInsightsUtility.rb
```diff
@@ -21,10 +22,12 @@ class ApplicationInsightsUtility
   @@EnvApplicationInsightsEndpoint = "APPLICATIONINSIGHTS_ENDPOINT"
   @@EnvControllerType = "CONTROLLER_TYPE"
   @@EnvContainerRuntime = "CONTAINER_RUNTIME"
+  @DefaultAppInsightsEndpoint = "https://dc.services.visualstudio.com/v2/track"
```
vishiy, line 25: do we need to hard-code the AI ingestion endpoint here ? I see that the API that requires proxy to be passed needs it, but this endpoint is already in the appinsights sdk that can be reused, rather then being defined yet again

source/code/plugin/lib/application_insights/channel/asynchronous_sender.rb
```diff
@@ -20,15 +20,16 @@ class AsynchronousSender < SenderBase
       SERVICE_ENDPOINT_URI = 'https://dc.services.visualstudio.com/v2/track'
       # Initializes a new instance of the class.
       # @param [String] service_endpoint_uri the address of the service to send
+      # @param [Hash] proxy server configuration to send (optional)
       #   telemetry data to.
-      def initialize(service_endpoint_uri = SERVICE_ENDPOINT_URI)
+      def initialize(service_endpoint_uri = SERVICE_ENDPOINT_URI, proxy = {})
```
vishiy, line 25: You can set the endpoint here from where its defined in AI SDK.

source/code/plugin/lib/application_insights/channel/asynchronous_sender.rb
```diff
@@ -20,15 +20,16 @@ class AsynchronousSender < SenderBase
       SERVICE_ENDPOINT_URI = 'https://dc.services.visualstudio.com/v2/track'
       # Initializes a new instance of the class.
       # @param [String] service_endpoint_uri the address of the service to send
+      # @param [Hash] proxy server configuration to send (optional)
```
vishiy, line 23: This should go to AI Ruby SDK and not checked in as part of our code. We just pull the gem and use it.

source/code/go/src/plugins/oms.go
```diff
@@ -990,11 +992,28 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 		} else {
 			Computer = strings.TrimSuffix(ToString(containerHostName), "\n")
 		}
+		// read proxyendpoint if proxy configured
+		ProxyEndpoint = ""
+		proxyConfigPath := pluginConfig["omsproxy_conf_path"]
+		if _, err := os.Stat(proxyConfigPath); err == nil {
+			Log("Reading proxy configuration for Linux from %s", proxyConfigPath)
+			proxyConfig, err := ioutil.ReadFile(proxyConfigPath)
```
vishiy, line 1000: I see that in your oms docker PR, there is an env being set (PROXY_ENDPOINT). If so why can't you consume that rather then read from config file ?

source/code/go/src/plugins/telemetry.go
```diff
@@ -159,6 +161,25 @@ func InitializeTelemetryClient(agentVersion string) (int, error) {
 		Log("Overriding the default AppInsights EndpointUrl with %s", appInsightsEndpoint)
 		telemetryClientConfig.EndpointUrl = envAppInsightsEndpoint
 	}
+	// if the proxy configured set the customized httpclient with proxy
+	isProxyConfigured := false
+	if ProxyEndpoint != "" {
```
vishiy, line 166: So you are checking if proxy is a valid uri here ? This is too late. Ideally you should validate proxy uri in main.sh (which is the entry point), and set the env variable if its valid and everybody should just read from env var & use it. If its invalid, just dont set the env var, and things will fail as it is before.

source/code/go/src/plugins/oms.go
```diff
@@ -990,11 +992,28 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
...
+			if err != nil {
+				message := fmt.Sprintf("Error Reading omsproxy configuration %s\n", err.Error())
+				Log(message)
+				// if we fail to read proxy configuration, then everything will be fall apart including sending  exceptions to AI
+				SendException(message)
+				log.Fatalln(message)
```
vishiy, line 1006: So if i do a kubectl logs on the pod which is set with invalid proxy, how do i know its crashlooping due to invalid proxy ? Wht would i see in the log that will tell me that proxy is invalid ?

source/code/plugin/lib/application_insights/channel/asynchronous_sender.rb
ganga1980: Done.

source/code/go/src/plugins/telemetry.go
ganga1980: Done.

source/code/plugin/lib/application_insights/channel/asynchronous_sender.rb
ganga1980: Done. Same change made to AI SDK repo as well, but yet to merged that PR.

PR #360, opened 2020-05-26
turnoff per container inventory telemetry 

ganga1980:
removed the putting per container telemetry to ai and also handle the exceptions reading the file which is possible if any of the container gone meaning (cproc file gone) which is expected.

### DIFF
--- a/source/code/plugin/kubernetes_container_inventory.rb
+++ b/source/code/plugin/kubernetes_container_inventory.rb
@@ -7,7 +7,7 @@ class KubernetesContainerInventory
   require "json"
   require_relative "omslog"
   require_relative "ApplicationInsightsUtility"
-  
+
   # cache the container and cgroup parent process
   @@containerCGroupCache = Hash.new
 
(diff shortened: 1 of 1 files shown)


source/code/plugin/kubernetes_container_inventory.rb
```diff
@@ -175,17 +175,20 @@ def obtainContainerEnvironmentVars(containerId)
...
+                    @@containerCGroupCache[containerId] = cGroupPid
+                  end
+                else
                   @@containerCGroupCache[containerId] = cGroupPid
                 end
-              else
-                @@containerCGroupCache[containerId] = cGroupPid
               end
+            rescue SystemCallError # ignore Error::ENOENT,Errno::ESRCH which is expected if any of the container gone while we read
             end
```
rashmichandrashekar, line 192: Can you add a log with a message or exception details rather than noop?

ganga1980: > Can you add a log with a message or exception details rather than noop?

Intentionally didnt log the errors related to file errors since these will be huge if the container getting deleted.

PR #368, opened 2020-06-10
Disable Health Plugin, introduce config map

r-dilip:


source/plugins/ruby/filter_cadvisor_health_container.rb
```diff
@@ -20,6 +20,7 @@ class CAdvisor2ContainerHealthFilter < Filter
         @@object_name_k8s_container = 'K8SContainer'
         @@counter_name_cpu = 'cpuusagenanocores'
         @@counter_name_memory_rss = 'memoryrssbytes'
+        @@cluster_health_model_enabled = HealthMonitorUtils.is_cluster_health_model_enabled
```
vishiy, line 23: I am assuming this function already exists in HM Utils from private preview as i dont see it in this PR ?

build/linux/installer/scripts/tomlparser-health-config.rb
```diff
@@ -0,0 +1,73 @@
...
+  rescue => errorStr
+    ConfigParseErrorLogger.logError("Exception while parsing config map for enabling health: #{errorStr}, using defaults, please check config map for errors")
+    return nil
+  end
+end
+
+# Use the ruby structure created after config parsing to set the right values to be used as environment variables
+def populateSettingValuesFromConfigMap(parsedConfig)
+  begin
+    if !parsedConfig.nil?  && !parsedConfig[:agent_settings][:health_model].nil? && !parsedConfig[:agent_settings][:health_model][:enabled].nil?
```
vishiy, line 39: By default configmap will not have [:agent_settings], as we will not put this setting there for this, so you should check for [:agent_settings].nil as well. Else it will always err and goto rescue block in all clusters.

PR #375, opened 2020-06-25
fix: Add tolerations for windows daemonset + add proper labels that match aks-rp

bragi92:


kubernetes/omsagent.yaml
```diff
@@ -582,20 +582,20 @@ spec:
...
-      component: oms-agent
-      tier: node
+      component: oms-agent-win
+      tier: node-win
    template:
     metadata:
       labels:
-        component: oms-agent
-        tier: node
+        component: oms-agent-win
```
vishiy, line 597: do you know if '-win' was missing before, or is it a merge issue ? Is it '-win' in RP ?

bragi92: yeah, -win is there in the RP. I think it got missed when it was added to Docker-Provider so I'm adding it back here.

The current test I've done is with the -win label being present.

vishiy: ok thanks. As long as we keep the labels the same between linux & windows in each yaml (RP,AKS-E,HELM), and not copy between them, we are good.

PR #378, opened 2020-06-26
Node Capacity from Metrics

r-dilip:


source/plugins/ruby/kubelet_utils.rb
```diff
@@ -17,11 +18,14 @@ def get_node_capacity
...
         memory_capacity = 1.0
 
-        response = CAdvisorMetricsAPIClient.getNodeCapacityFromCAdvisor(winNode: nil)
+        response = CAdvisorMetricsAPIClient.getAllMetricsCAdvisor(winNode: nil)
         if !response.nil? && !response.body.nil?
-          cpu_capacity = JSON.parse(response.body)["num_cores"].nil? ? 1.0 : (JSON.parse(response.body)["num_cores"] * 1000.0)
-          memory_capacity = JSON.parse(response.body)["memory_capacity"].nil? ? 1.0 : JSON.parse(response.body)["memory_capacity"].to_f
-          @log.info "CPU = #{cpu_capacity}mc Memory = #{memory_capacity / 1024 / 1024}MB"
+          all_metrics = response.body.split("\n")
+          cpu_capacity = all_metrics.select{|m| m.start_with?('machine_cpu_cores') }.first.split.last.to_f * 1000
```
vishiy, line 24: did you check the payload that no other metric in it has these 2 metric names as prefix ?

vishiy, line 24: better to check the first element after the spilt is indeed 'machine_cpu_cores' for future proofing, so newer metrics with this prefix doesn't break us.

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -17,11 +18,14 @@ def get_node_capacity
         cpu_capacity = 1.0
         memory_capacity = 1.0
 
-        response = CAdvisorMetricsAPIClient.getNodeCapacityFromCAdvisor(winNode: nil)
+        response = CAdvisorMetricsAPIClient.getAllMetricsCAdvisor(winNode: nil)
```
ganga1980, line 21: these calls can also fail, probably checking for status code and also emitting telemetry is good idea.

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -17,11 +18,14 @@ def get_node_capacity
...
         if !response.nil? && !response.body.nil?
-          cpu_capacity = JSON.parse(response.body)["num_cores"].nil? ? 1.0 : (JSON.parse(response.body)["num_cores"] * 1000.0)
-          memory_capacity = JSON.parse(response.body)["memory_capacity"].nil? ? 1.0 : JSON.parse(response.body)["memory_capacity"].to_f
-          @log.info "CPU = #{cpu_capacity}mc Memory = #{memory_capacity / 1024 / 1024}MB"
+          all_metrics = response.body.split("\n")
+          cpu_capacity = all_metrics.select{|m| m.start_with?('machine_cpu_cores') }.first.split.last.to_f * 1000
+          @log.info "CPU Capacity #{cpu_capacity}"
+          memory_capacity_e = all_metrics.select{|m| m.start_with?('machine_memory_bytes') }.first.split.last
```
ganga1980, line 26: as vishwa commented in other comment, having strict check on metric name will ensure we are picking the right metric.

source/plugins/ruby/kubelet_utils.rb
r-dilip: doing an exact comparison

r-dilip: Done

source/plugins/ruby/kubelet_utils.rb
r-dilip: we are already emitting telemetry on failure here. (L33)

ganga1980: > we are already emitting telemetry on failure here. (L33)

> @r-dilip, what I am saying,  extract metrics only when response.code == 200 and if its not 200, just emit the telemetry.  this we can avoid sending un-necessarily full call stack for all non-200 responses.

PR #393, opened 2020-07-07
Gangams/enable release pipe line

ganga1980:
This PR has the following changes 

1.  Enable release pipeline to automatically deploy bits on to CIDEV AKS cluster in Build subscription
2.  Instructions for the release pipeline 
3. Readme updates
4. updates to gitignore and gitattributes files
5. Instructions for production release
6. Fix line endings to LF for yaml so that sed command can be used.


kubernetes/omsagent.yaml
```diff
@@ -1,728 +1,731 @@
...
+          imagePullPolicy: IfNotPresent
+          resources:
+            limits:
+              cpu: 1
+              memory: 750Mi
+            requests:
+              cpu: 150m
+              memory: 250Mi
+          env:
+            # azure devops pipeline uses AKS_RESOURCE_ID and AKS_REGION hence ensure to uncomment these
```
vishiy, line 483: why change default from non-aks to AKS ?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -1,105 +1,105 @@
-kind: ConfigMap
```
rashmichandrashekar, line 1: if its whitespace only, can you undo changes to this file?

kubernetes/omsagent.yaml
ganga1980: > why change default from non-aks to AKS ?

> pipelines are replacing these place-holders to deploy onto ci_dev and ci_prod clusters. do you see any issue  changing this?

kubernetes/container-azm-ms-agentconfig.yaml
ganga1980: > if its whitespace only, can you undo changes to this file?

> sure, let me do it.

kubernetes/omsagent.yaml
ganga1980: > > why change default from non-aks to AKS ?
> 
> > pipelines are replacing these place-holders to deploy onto ci_dev and ci_prod clusters. do you see any issue  changing this? for non-AKS, clusters, release pipelines will uses the HELM chart which I am planning to do next PR.

kubernetes/container-azm-ms-agentconfig.yaml
ganga1980: > > if its whitespace only, can you undo changes to this file?
> 
> > sure, let me do it.

Done.

kubernetes/omsagent.yaml
vishiy: no, not an issue. should be ok.

PR #396, opened 2020-07-10
fix jq oomkilled issue

ganga1980:
selecting only agent pods to determine the container runtime. validate on my local cluster  and yet to validate on the clusters which has the repro.  I will validate Sarah and also we can validate with customer during customer call.

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -81,13 +81,13 @@ if [ -e "/etc/omsagent-secret/WSID" ]; then
 
       if [ -e "/etc/omsagent-secret/PROXY" ]; then
             export PROXY_ENDPOINT=$(cat /etc/omsagent-secret/PROXY)
-            # Validate Proxy Endpoint URL 
+            # Validate Proxy Endpoint URL
             # extract the protocol://
             proto="$(echo $PROXY_ENDPOINT | grep :// | sed -e's,^\(.*://\).*,\1,g')"
             # convert the protocol prefix in lowercase for validation
             proxyprotocol=$(echo $proto | tr "[:upper:]" "[:lower:]")
             if [ "$proxyprotocol" != "http://" -a "$proxyprotocol" != "https://" ]; then
-               echo "-e error proxy endpoint should be in this format http(s)://<user>:<pwd>@<hostOrIP>:<port>"           
+               echo "-e error proxy endpoint should be in this format http(s)://<user>:<pwd>@<hostOrIP>:<port>"
             fi
             # remove the protocol
             url="$(echo ${PROXY_ENDPOINT/$proto/})"
@@ -97,7 +97,7 @@ if [ -e "/etc/omsagent-secret/WSID" ]; then
             pwd="$(echo $creds | cut -d':' -f2)"
             # extract the host and port
             hostport="$(echo ${url/$creds@/} | cut -d/ -f1)"
-            # extract host without port    
+            # extract host without port
             host="$(echo $hostport | sed -e 's,:.*,,g')"
             # extract the port
             port="$(echo $hostport | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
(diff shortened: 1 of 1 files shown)


kubernetes/linux/main.sh
```diff
@@ -259,16 +259,15 @@ else
...
+      ITEMS_COUNT=$(cat podsResponseFile | jq '.items | length')
       echo "found items count: $ITEMS_COUNT"
       if [ $ITEMS_COUNT -gt 0 ]; then
-            # exclude the pods which doesnt have containerId. could happen if the container fails to start because of bad image and tag etc..
-            podsWithValidContainerId=$(echo $podsResponse | jq -r '[.items[] | select( .status.containerStatuses != null and .status.containerStatuses[].containerID != null and  .status.containerStatuses[].containerID != "")]')
-            ITEMS_COUNT_WITH_CONTAINER_ID=$(echo $podsWithValidContainerId | jq '. | length')
-            if [ $ITEMS_COUNT_WITH_CONTAINER_ID -gt 0 ]; then
-                  containerRuntime=$(echo $podsWithValidContainerId | jq -r '.[0].status.containerStatuses[0].containerID' | cut -d ':' -f 1)
-                  nodeName=$(echo $podsWithValidContainerId | jq -r '.[0].spec.nodeName')
+            # select omsagent pods
```
vishiy, line 265: Do you the root cause and how will this fix the actual issue ?

kubernetes/linux/main.sh
```diff
@@ -259,16 +259,15 @@ else
...
+            omsagentPods=$(cat podsResponseFile | jq -r '[.items[] | select( any(.metadata.name; startswith("omsagent-")))]')
+            ITEMS_COUNT_WITH_OMS_AGENT=$(echo $omsagentPods | jq '. | length')
+            if [ $ITEMS_COUNT_WITH_OMS_AGENT -gt 0 ]; then
+                  containerRuntime=$(echo $omsagentPods | jq -r '.[0].status.containerStatuses[0].containerID' | cut -d ':' -f 1)
+                  nodeName=$(echo $omsagentPods | jq -r '.[0].spec.nodeName')
```
vishiy, line 270: Why do we need this tool, rather use python for json parsing , which is reliable & tested ?

kubernetes/linux/main.sh
```diff
@@ -259,16 +259,15 @@ else
...
       echo "found items count: $ITEMS_COUNT"
       if [ $ITEMS_COUNT -gt 0 ]; then
-            # exclude the pods which doesnt have containerId. could happen if the container fails to start because of bad image and tag etc..
-            podsWithValidContainerId=$(echo $podsResponse | jq -r '[.items[] | select( .status.containerStatuses != null and .status.containerStatuses[].containerID != null and  .status.containerStatuses[].containerID != "")]')
-            ITEMS_COUNT_WITH_CONTAINER_ID=$(echo $podsWithValidContainerId | jq '. | length')
-            if [ $ITEMS_COUNT_WITH_CONTAINER_ID -gt 0 ]; then
-                  containerRuntime=$(echo $podsWithValidContainerId | jq -r '.[0].status.containerStatuses[0].containerID' | cut -d ':' -f 1)
-                  nodeName=$(echo $podsWithValidContainerId | jq -r '.[0].spec.nodeName')
+            # select omsagent pods
+            omsagentPods=$(cat podsResponseFile | jq -r '[.items[] | select( any(.metadata.name; startswith("omsagent-")))]')
```
vishiy, line 266: please dont cat json in to a bash variable. You can simply pipe in bash

kubernetes/linux/main.sh
ganga1980: > Do you know the root cause and how will this fix the actual issue ?

> issue seems enumerating and validating large containerStatuses array. I have created issue on jq to confirm this issue. The fix is just checks for the our agent pods based on the metadata fields on any condition. I have validated on large JSON and empty JSON etc.. seems working fine. Planning to do more testing on this with larger size JSONs with many containers/pods etc..  I dont know customer has the same issue or not but we will have to confirm.  Seems having >100 containers per pod is not common.

vishiy: i am not sure about that, because our telemetry says there are few more clusters in this state. I suspect its because of something in the payload. So if you copy omsagent container 100 times in the payload does it repro ?

kubernetes/linux/main.sh
ganga1980: > Why do we need this tool, rather use python for json parsing , which is reliable & tested ?
> Let me explore Python JSON parser and see if that is better and works for all the scenarios particularly with larger payloads.  Jq is popular tool and well tested as well.

kubernetes/linux/main.sh
ganga1980: > i am not sure about that, because our telemetry says there are few more clusters in this state. I suspect its because of something in the payload. So if you copy omsagent container 100 times (with different names) in the payload does it repro ?

> this is not issue with payload. on same payload fix in this pr or without this condition .status.containerStatuses[].containerID != null and  .status.containerStatuses[].containerID != "" works fine. So,  issue specific checking for containerID null or empty in ContainerStatuses array as the number of containers growing
.status.containerStatuses[].containerID != null and  .status.containerStatuses[].containerID != ""

since the expression is to enumerate containerStatuses array and check for containerID is not null and empty.
The result can be large if the containers and pods are too many.

updated fix is just to check our agent pods based on metadata and at max we will have two pods on node (rs and ds) scheduled.

ganga1980:
I will be abandon this PR and will create PR first against ci_feature_prod with all hot fixes and  make

PR #410, opened 2020-07-30
Saaror patch 2

saaror:
Added Health Private preview instructions.


Health/onboarding_instructions.md
```diff
@@ -0,0 +1,27 @@
...
+```cmd:agent-settings: |-
+    [agent_settings.health_model]
+      enabled = true
+```
+2. Run the following kubectl command:
+   `kubectl apply -f <configmap_yaml_file.yaml>`
+   
+Example: `kubectl apply -f container-azm-ms-agentconfig.yaml`.
+
+The configuration change can take a few minutes to finish before taking effect, and all omsagent pods in the cluster will restart. The restart is a rolling restart for all omsagent pods, not all restart at the same time. When the restarts are finished, a message is displayed that's similar to the following and includes the result: `configmap "container-azm-ms-agentconfig"` created.
```
vishiy, line 19: This message will be shown right after configmap is created (first time configmap) or when config map is updated (if it already had existed). Restarts will happen ~4mins from this message. So just remove this ?

saaror: updated.

PR #418, opened 2020-08-05
Final collector changes

ayusheesingh:
Minor changes to example and refactoring of code to account for service discovery and capturing telemetry on collector usage


kubernetes/omsagent-otel.yaml
```diff
@@ -327,7 +327,7 @@ spec:
       serviceAccountName: omsagent
       containers:
         - name: omsagent
-          image: "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:ciprod05262020"
+          image: ayushisingh/omsagent-ci:08032020
```
ganga1980, line 330: we should push the image to acr which will replicate to mcr and use that image.

kubernetes/omsagent-otel.yaml
```diff
@@ -691,25 +694,28 @@ spec:
...
-            memory: 2Gi
+            cpu: 150m
+            memory: 600Mi
           requests:
-            cpu: 200m
-            memory: 400Mi
+            cpu: 75m
+            memory: 225Mi
+        env: 
+          - name: APPLICATIONINSIGHTS_AUTH
```
ganga1980, line 713: we should put this in dockerfile similar to what we do in container insighsts agent.

kubernetes/omsagent-otel.yaml
```diff
@@ -336,7 +336,7 @@ spec:
             requests:
               cpu: 75m
               memory: 225Mi
-          env:
+          env: 
             - name: AKS_RESOURCE_ID
               value: "/subscriptions/72c8e8ca-dc16-47dc-b65c-6b5875eb600a/resourcegroups/ayushi-test/providers/Microsoft.ContainerService/managedClusters/t-aysi_cluster"
```
ganga1980, line 341: we should revert this to place holder value.

PR #423, opened 2020-08-06
Fix for zero fill metrics leaking in unsupported regions

rashmichandrashekar:
Adding check for zero filled metrics since these were leaking for unsupported regions

### DIFF
--- a/source/plugins/ruby/podinventory_to_mdm.rb
+++ b/source/plugins/ruby/podinventory_to_mdm.rb
@@ -93,67 +93,71 @@ class Inventory2MdmConvertor
   end
 
   def get_pod_inventory_mdm_records(batch_time)
+    records = []
     begin
-      # generate all possible values of non_phase_dim_values X pod Phases and zero-fill the ones that are not already present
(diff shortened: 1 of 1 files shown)


source/plugins/ruby/podinventory_to_mdm.rb
```diff
@@ -93,67 +93,71 @@ def initialize(custom_metrics_azure_regions)
...
-        if key_elements.length != 4
-          next
-        end
 
-        # get dimension values by key
-        podNodeDimValue = key_elements[0]
-        podNamespaceDimValue = key_elements[1]
-        podControllerNameDimValue = key_elements[2]
-        podPhaseDimValue = key_elements[3]
+          # get dimension values by key
```
ganga1980, line 116: bunch whitespaces and i think, we need to get rid off these to have cleaner code. I use this setting in vscode "files.trimTrailingWhitespace": true to avoid getting these whitespaces

rashmichandrashekar: this is coming from formatter. I have ruby formatter that is formatting the code

PR #428, opened 2020-08-18
Gangams/add containerd support to windows agent

ganga1980:
Add containerd support for windows agent. tested on aks-engine cluster with 1.18 with containerd and docker on both aks and aks-engine.


kubernetes/omsagent.yaml
```diff
@@ -660,6 +660,10 @@ spec:
             valueFrom:
                fieldRef:
                  fieldPath: spec.nodeName
+          - name: NODE_IP
```
bragi92, line 663: Are you also making a PR in the HELM repo and the AKS-RP repo's?

build/windows/installer/conf/fluent-cri-parser.conf
```diff
@@ -0,0 +1,6 @@
+<parse>
+  @type  regexp
+  expression ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<log>.*)$
```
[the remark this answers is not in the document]
ganga1980: > Is the parser for linux different?

Parser logic is same for Linux and windows but for linux for we use Fluentbit and for windows we use Fluentd. hence we need to maintain separate parsers since Fluentd and Fluentbit has separate syntax for parsers.

kubernetes/omsagent.yaml
ganga1980: > Are you also making a PR in the HELM repo and the AKS-RP repo's?
> These changes goes as part of release agent.

PR #440, opened 2020-09-16
[Merge] dev to prod merge for ciprod09162020 non-aks release

ganga1980:
dev to prod merge for ciprod09162020 non-aks release


ganga1980:
since going with creating prod image from ci_dev so cancelling this merge for now. we can switch to single branch approach which would avoid spending time on resolving weird conflicts.

PR #445, opened 2020-09-25
Gangams/sept agent release tasks

ganga1980:
This PR has following changes

1. Turn off MDM Ingestion for non-supported clusters and only allow AKS and Arc K8s
2. Enable kubelet_runtime_operations_total and kubelet_runtime_operations_errors_total metrics
3. Enable OpenSSL::SSL::VERIFY_PEER for AI ruby client to prevent MTM attacks
4. handle both beta.kubernetes.io/os  and kubernetes.io/os label since beta one deprecated
5. Move reading of azure.json file for aks code path since this doesnt exist on-prem clusters of arc k8s

### DIFF
--- a/build/linux/installer/conf/telegraf.conf
+++ b/build/linux/installer/conf/telegraf.conf
@@ -632,8 +632,7 @@
   name_prefix="container.azm.ms/"
   ## An array of urls to scrape metrics from.
   urls = ["$CADVISOR_METRICS_URL"]
-  ## Include "$KUBELET_RUNTIME_OPERATIONS_TOTAL_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_TOTAL_METRIC" when we add for support for 1.18
-  fieldpass = ["$KUBELET_RUNTIME_OPERATIONS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_METRIC"]
+  fieldpass = ["$KUBELET_RUNTIME_OPERATIONS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_TOTAL_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_TOTAL_METRIC"]
 
   metric_version = 2
   url_tag = "scrapeUrl"
@@ -675,7 +674,7 @@
   name_prefix="container.azm.ms/"
   ## An array of urls to scrape metrics from.
   urls = ["$CADVISOR_METRICS_URL"]
-  
+
   fieldpass = ["kubelet_running_pod_count","volume_manager_total_volumes", "kubelet_node_config_error", "process_resident_memory_bytes", "process_cpu_seconds_total"]
 
   metric_version = 2
@@ -690,7 +689,7 @@
   ## Optional TLS Config
   tls_ca = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
   insecure_skip_verify = true
-  
+
 
 ## prometheus custom metrics
 [[inputs.prometheus]]
@@ -731,7 +730,7 @@
   #name_prefix="container.azm.ms/"
   ## An array of urls to scrape metrics from.
   urls = $AZMON_INTEGRATION_NPM_METRICS_URL_LIST_NODE
-  
+
   metric_version = 2
   url_tag = "scrapeUrl"
 
--- a/charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
+++ b/charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
@@ -25,12 +25,13 @@ spec:
     dockerProviderVersion: {{ .Values.omsagent.image.dockerProviderVersion }}
     schema-versions: "v1"
   spec:
-    nodeSelector:
-    {{- if semverCompare "^1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion }}
+   nodeSelector:
       kubernetes.io/os: windows
-    {{- else -}}
+{{- else }}
+   nodeSelector:
       beta.kubernetes.io/os: windows
-    {{- end }}
+{{- end }}
    {{- if .Values.omsagent.rbac }}
    serviceAccountName: omsagent
    {{- end }}
--- a/charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
+++ b/charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
@@ -25,8 +25,12 @@ spec:
     dockerProviderVersion: {{ .Values.omsagent.image.dockerProviderVersion }}
     schema-versions: "v1"
   spec:
-   nodeSelector:
+    nodeSelector:
+    {{- if semverCompare "^1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+      kubernetes.io/os: windows
+    {{- else -}}
       beta.kubernetes.io/os: windows
+    {{- end }}
    {{- if .Values.omsagent.rbac }}
    serviceAccountName: omsagent
    {{- end }}
--- a/charts/azuremonitor-containers/values.yaml
+++ b/charts/azuremonitor-containers/values.yaml
@@ -59,11 +59,18 @@ omsagent:
           nodeSelectorTerms:
             - labelSelector:
               matchExpressions:
-              {{- if semverCompare "^1.14-0" .Capabilities.KubeVersion.GitVersion -}}
                 - key: kubernetes.io/os
-              {{- else -}}
+                  operator: In
+                  values:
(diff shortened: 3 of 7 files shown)


charts/azuremonitor-containers/values.yaml
```diff
@@ -71,6 +82,21 @@ omsagent:
     affinity:
       nodeAffinity:
         requiredDuringSchedulingIgnoredDuringExecution:
```
rashmichandrashekar, line 84: Are these changes from this release?

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
```diff
@@ -66,19 +66,19 @@ def send(data_to_send)
...
           http = Net::HTTP.new uri.hostname, uri.port
-        else 
+        else
           http = Net::HTTP.new(uri.hostname, uri.port, @proxy[:addr], @proxy[:port], @proxy[:user], @proxy[:pass])
         end
-        if uri.scheme.downcase == 'https'
+        if uri.scheme.downcase == "https"
           http.use_ssl = true
-          http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+          http.verify_mode = OpenSSL::SSL::VERIFY_PEER
```
rashmichandrashekar, line 74: This needs to validated in airgapped clouds

build/linux/installer/conf/telegraf.conf
```diff
@@ -632,8 +632,7 @@
   name_prefix="container.azm.ms/"
   ## An array of urls to scrape metrics from.
   urls = ["$CADVISOR_METRICS_URL"]
-  ## Include "$KUBELET_RUNTIME_OPERATIONS_TOTAL_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_TOTAL_METRIC" when we add for support for 1.18
-  fieldpass = ["$KUBELET_RUNTIME_OPERATIONS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_METRIC"]
+  fieldpass = ["$KUBELET_RUNTIME_OPERATIONS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_METRIC", "$KUBELET_RUNTIME_OPERATIONS_TOTAL_METRIC", "$KUBELET_RUNTIME_OPERATIONS_ERRORS_TOTAL_METRIC"]
```
rashmichandrashekar, line 635: Did we decide to go with the duplicate metrics approach?

source/plugins/ruby/out_mdm.rb
```diff
@@ -106,6 +108,9 @@ def start
             @cluster_identity = ArcK8sClusterIdentity.new
             @cached_access_token = @cluster_identity.get_cluster_identity_token
           else
+            # azure json file only used for aks and doesnt exist in non-azure envs
```
vishiy, line 111: what is the actual issue due to which this fix is needed ? When i check the cidev cluster i see custom metrics . Can u explain the problem for this fix ?

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
```diff
@@ -53,9 +53,9 @@ def initialize(service_endpoint_uri, proxy = {})
       def send(data_to_send)
         uri = URI(@service_endpoint_uri)
         headers = {
-          'Accept' => 'application/json',
-          'Content-Type' => 'application/json; charset=utf-8',
-          'Content-Encoding' => 'gzip'
+          "Accept" => "application/json",
```
vishiy, line 56: Are the quote changes intentional ?

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
ganga1980: > This needs to validated in airgapped clouds

>  this shouldnt impact on airgapped clouds since we do all other places. This is just server cert validation  to prevent MITM attacks which was not done AI ruby SDK and all other places we do it for example AI go sdk and 

https://github.com/microsoft/Docker-Provider/blob/ci_dev/source/plugins/ruby/KubernetesApiClient.rb#L50
https://github.com/microsoft/Docker-Provider/blob/ci_dev/build/windows/installer/scripts/rubyKeepCertificateAlive/omsagenthelper.rb#L94

source/plugins/ruby/out_mdm.rb
ganga1980: > what is the actual issue due to which this fix is needed ? When i check the cidev cluster i see custom metrics . Can u explain the problem for this fix ?

>  as you can see, code comment has the explanation  " # azure json file only used for aks and doesnt exist in non-azure envs". since azure.json file doesnt exist on-prem k8s cluster so we were crashing.  we only use azure json file for aks so moved under code path.

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
ganga1980: > Are the quote changes intentional ?
> vs code automatically adding that. I can revert it but doesnt do any harm.

build/linux/installer/conf/telegraf.conf
ganga1980: > Did we decide to go with the duplicate metrics approach?

> discussed many times.  :)

charts/azuremonitor-containers/values.yaml
ganga1980: > Are these changes from this release?

> yes.  these are to handle both deprecated beta.kubernetes.os and kubernetes.os labels.

source/plugins/ruby/out_mdm.rb
vishiy: so we added metrics for non-aks and tested in aks ?

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
ganga1980: > > Are the quote changes intentional ?
> > vs code automatically adding that. I can revert it but doesnt do any harm.

>> reverted the quotes though this shouldnt impact anything for easy code reviews.

charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
```diff
@@ -25,8 +25,13 @@ spec:
     dockerProviderVersion: {{ .Values.omsagent.image.dockerProviderVersion }}
     schema-versions: "v1"
   spec:
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion }}
```
vishiy, line 28: i am not sure about this change. we should ideally do a 'or' so we work on both, as  this change requires customer workloads to be updated and most people dont. and also in on prem envs, this is highly customizable.

vishiy (approved):

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
rashmichandrashekar: @ganga1980 - This is the trusted CA validation that happens when you change this to verify peer.. This has caused problems for ods and oms endpoints. Do you know which CA is used in AIrgapped clouds? This is generally different in these clouds and may not be in the trusted store and connections will fail.

build/linux/installer/conf/telegraf.conf
rashmichandrashekar: The last conversation i remember was about handling versions. May be this changed.As long as this is discussed its fine.

charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
ganga1980: > i am not sure about this change. we should ideally do a 'or' so we work on both, as this change requires customer workloads to be updated and most people dont. and also in on prem envs, this is highly customizable.

> No, this doesnt require customer to upgrade their k8s cluster since we handle pre 1.14 and higher.

vishiy: i am not saying this agent change requires customers to upgrade k8s. This label change requires customers workload to be updated (just like our agent) ,and most customers, just add this label back instead of updating their side, as node labels are templated in k8s installation.

source/plugins/ruby/lib/application_insights/channel/sender_base.rb
ganga1980: created a task to handle this. let me reach out Joby on the instructions to get access.

charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
ganga1980: > i am not saying this agent change requires customers to upgrade k8s. This label change requires customers workload to be updated (just like our agent) ,and most customers, just add this label back instead of updating their side, as node labels are templated in k8s installation.

> Since we are using nodeselector, we cant express  `or` condition and nodeSelector only supports `and`.  If the customer doesnt upgraded k8s version but just adds the  `kubernetes.io/os` then this should work as long as they dont remove beta.kubernetes.io label.

PR #451, opened 2020-09-30
Install CA certs from wireserver

rashmichandrashekar:
Install CA certs from wireserver so that communication to OMS and ODS endpoints succeed


kubernetes/windows/main.ps1
```diff
@@ -263,6 +263,27 @@ function Generate-Certificates {
     C:\\opt\\omsagentwindows\\certgenerator\\certificategenerator.exe
 }
 
+function Bootstrap-CACertificates {
+    try {
+        # This is required when the root CA certs are different for some clouds.
+        $caCerts=Invoke-WebRequest 'http://168.63.129.16/machine?comp=acmspackage&type=cacertificates&ext=json' -UseBasicParsing | ConvertFrom-Json
```
ganga1980, line 269: since this outbound network call, we need to honor the proxy configuration, if its specified similar to other network call we do in this file.

ganga1980, line 269: since us{nat;sec} clouds are stringent on security, wondering why http instead of https here?

ganga1980 (changes requested):

kubernetes/windows/main.ps1
rashmichandrashekar: @ganga1980 - could you please remove the cloud names from this comment since it is open source?

kubernetes/windows/main.ps1
```diff
@@ -263,6 +263,27 @@ function Generate-Certificates {
     C:\\opt\\omsagentwindows\\certgenerator\\certificategenerator.exe
 }
 
+function Bootstrap-CACertificates {
```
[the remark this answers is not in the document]
rashmichandrashekar: @ganga1980 - could you please remove the cloud names from this comment since it is open source?

kubernetes/windows/main.ps1
rashmichandrashekar: here too..

rashmichandrashekar: To answer you qs - this was the url provided by the compute team. So i didnt want to make any changes to that.

ganga1980: > since new clouds are stringent on security, wondering why http instead of https here?



> us{nat;sec}

ganga1980: I think, we need to understand why they are not supporting https. I am pretty sure they might have support https or plan to enable this when we raise this question.

rashmichandrashekar: @ganga1980 - You can find more information here - https://docs.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16#scope-of-ip-address-1686312916

This is an IP address owned by Microsoft and this is an excerpt from the doc - " The communication between this special IP address and the resources is safe because only the internal Azure platform can source a message from this IP address"

rashmichandrashekar: This is the IP address over which the VM talks to the physical server on which the vm is hosted. I dont think this requires proxy settings.

kubernetes/windows/main.ps1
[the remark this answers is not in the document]
rashmichandrashekar: In other clouds this endpoint doesnt return any certificates. So this is a no-op. I dont think we will need any additional check here.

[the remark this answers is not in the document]
rashmichandrashekar: I have added the check for AKS clusters only and in public this doesnt return any results.. so this will be a no-op

vishiy (approved):

kubernetes/windows/main.ps1
```diff
@@ -288,6 +309,12 @@ Start-Transcript -Path main.txt
 Remove-WindowsServiceIfItExists "fluentdwinaks"
 Set-EnvironmentVariables
 Start-FileSystemWatcher
+
+#Bootstrapping CA certs for non public clouds and AKS clusters
+if (![string]::IsNullOrEmpty([System.Environment]::GetEnvironmentVariable("AKS_RESOURCE_ID"))) {
```
ganga1980, line 314: Rashmi, Since AKS_RESOURCE_ID env used for both AKS, Azure Arc K8s and ARO cluster. We need to add check for AKS cluster type by checking the cluster resource id has "Microsoft.ContainerService/managedClusters" in it. Please do case insensitive check since the cluster resource id can come different casing.

vishiy: We discussed this in the afternoon and decided to go this way because we want all managed clusters to be supported in these clouds. That said, you need to provide  better tooling/API support for different cluster flavors (ARO,ARC,AKS-E etc..). Just doing this env var check in many many places is not the correct & efficient thing to do. What we need is a fully typed cluster metadata class with properties that identifies cluster attributes that we are interested in.

ganga1980: Create task in the oct agent release https://msazure.visualstudio.com/InfrastructureInsights/_sprints/taskboard/Containers/InfrastructureInsights/Iron/20Oct02?workitem=8463970 to handle cluster type detection in unified fashion across all the places. We can discuss about what would be best and consistent, and efficient way to do this.

ganga1980 (approved):

PR #452, opened 2020-09-30
grwehner/pv-volume-name-in-mdm

gracewehner:
Add the volume name to mdm dimensions and zero fill it


source/plugins/ruby/constants.rb
```diff
@@ -76,6 +76,7 @@ class Constants
   TELEGRAF_DISK_METRICS = "container.azm.ms/disk"
   OMSAGENT_ZERO_FILL = "omsagent"
   KUBESYSTEM_NAMESPACE_ZERO_FILL = "kube-system"
+  VOLUME_NAME_ZERO_FILL = "volume"
```
vishiy, line 79: for zero filled metrics, the volume name could be '-' ? That will make it more clear that this is indeed a zero filled one ?

rashmichandrashekar (dismissed):

vishiy (approved):

PR #455, opened 2020-10-07
grwehner/pv inventory

gracewehner:
Add fluent plugin to request persistent volume info from the kubernetes api and send to LA for new PV inventory table


source/plugins/ruby/constants.rb
```diff
@@ -77,13 +77,17 @@ class Constants
...
+    "flocker", "gcePersistentDisk", "glusterfs", "hostPath", "iscsi", "local", "nfs",
+    "photonPersistentDisk", "portworxVolume", "quobyte", "rbd", "scaleIO", "storageos", "vsphereVolume"]
 
   #Telemetry constants
   CONTAINER_METRICS_HEART_BEAT_EVENT = "ContainerMetricsMdmHeartBeatEvent"
   POD_READY_PERCENTAGE_HEART_BEAT_EVENT = "PodReadyPercentageMdmHeartBeatEvent"
   CONTAINER_RESOURCE_UTIL_HEART_BEAT_EVENT = "ContainerResourceUtilMdmHeartBeatEvent"
   PV_USAGE_HEART_BEAT_EVENT = "PVUsageMdmHeartBeatEvent"
   PV_KUBE_SYSTEM_METRICS_ENABLED_EVENT = "CollectPVKubeSystemMetricsEnabled"
+  PV_INVENTORY_HEART_BEAT_EVENT = "KubePVInventoryHeartBeatEvent"
```
vishiy, line 90: Do you think we need hearbeat telemetry for this plugin ? We have for a few, but i dont think we use them much.

source/plugins/ruby/in_kube_pvinventory.rb
```diff
@@ -0,0 +1,253 @@
...
+      require "yajl"
+      require "time"
+      require_relative "KubernetesApiClient"
+      require_relative "ApplicationInsightsUtility"
+      require_relative "oms_common"
+      require_relative "omslog"
+      require_relative "constants"
+
+      # Response size is around 1500 bytes per PV
+      @PV_CHUNK_SIZE = "5000"
```
vishiy, line 20: How much (in terms of bytes) does 5k PVs attribute to in the payload ?

source/plugins/ruby/in_kube_pvinventory.rb
```diff
@@ -0,0 +1,253 @@
...
+        $log.debug_backtrace(errorStr.backtrace)
+        ApplicationInsightsUtility.sendExceptionTelemetry(errorStr)
+      end
+    end
+
+    def getPVCInfo(item)
+      begin
+        if !item["spec"].nil? && !item["spec"]["claimRef"].nil?
+          claimRef = item["spec"]["claimRef"]
+          pvcNamespace = claimRef["namespace"]
```
vishiy, line 171: claimRef always has name & namespace, so nil/empty check not needed here ?

source/plugins/ruby/in_kube_pvinventory.rb
```diff
@@ -0,0 +1,253 @@
...
+      begin
+        if !item["spec"].nil?
+          (Constants::PV_TYPES).each do |pvType|
+      
+            # PV is this type
+            if !item["spec"][pvType].nil?
+
+              # Get additional info if azure disk/file
+              typeInfo = {}
+              if pvType == "azureDisk"
```
vishiy, line 195: may be tolower(pvType) == "azuredisk" /"azurefile" -- just in case

source/plugins/ruby/in_kube_pvinventory.rb
gracewehner: claimRef will always have a name & namespace, but if a PV doesn't have a claim, I think there is no claimRef field at all for spec and item["spec"]["claimRef"] would be nil. I can double check.

source/plugins/ruby/in_kube_pvinventory.rb
gracewehner: From testing, there's an average response size of about 1500 bytes per PV in the payload. So 7.5 MB for 5k PVs. I kept the chunk size very conservative since I wasn't sure if customers might have longer names than I did for the PV, PVC, disk, or share file. We can increase it though.

source/plugins/ruby/constants.rb
gracewehner: The heartbeat itself is probably not needed. I am using it [here](https://github.com/microsoft/Docker-Provider/pull/455/files#diff-24c9dec7f951dc00ce99813aad1129dcR136) to collect telemetry on the PV types and the count of each type. This info might be nice to know for supporting more detailed type info for other types, even if it's sent with a very large interval. Do we want to collect this?

source/plugins/ruby/in_kube_pvinventory.rb
vishiy: look fine then. thanks.

source/plugins/ruby/constants.rb
vishiy: ok. so you are using this event to get telemetry about pv types we see in the cluster as well. If so, please leave it as is. Yes, its very useful to have this telemetry.

source/plugins/ruby/in_kube_pvinventory.rb
gracewehner: PV spec with no claim is below...so we have to check if claimRef exists to know if there is a PVC. Or am I misunderstanding what you meant?
image

vishiy (approved):
please dont merge until we have the table created.

PR #456, opened 2020-10-08
Update onboarding_instructions.md

saaror:
Updated the documentation to reflect where to update the config map.

### DIFF
--- a/Health/onboarding_instructions.md
+++ b/Health/onboarding_instructions.md
@@ -6,15 +6,28 @@ For on-boarding to Health(Tab), you would need to complete two steps
 
 
 ## Configure agent through ConfigMap
-1. Include the following section in ConfigMap yaml file
+1. If you are configuring your existing ConfigMap, append the following section in your existing ConfigMap yaml file
 ```
+#Append this section in your existing configmap
 agent-settings: |-
        # agent health model feature settings   
     [agent_settings.health_model]   
       # In the absence of this configmap, default value for enabled is false   
       enabled = true
 ```
-2. Run the following kubectl command:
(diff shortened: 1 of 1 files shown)


Health/onboarding_instructions.md
```diff
@@ -7,8 +7,11 @@ For on-boarding to Health(Tab), you would need to complete two steps
 
 ## Configure agent through ConfigMap
 1. Include the following section in ConfigMap yaml file
-```cmd:agent-settings: |-
-    [agent_settings.health_model]
+```
```
rashmichandrashekar, line 10: I think we need to be more explicit here.
We need to mention, if they already have the cofigmap applied, they can append this to the file.(So that they dont override the existing configmap with default values)
Else, they can get the latest configmap file with default values and add this to the end.

rashmichandrashekar (dismissed):

Health/onboarding_instructions.md
```diff
@@ -15,7 +15,7 @@ agent-settings: |-
       # In the absence of this configmap, default value for enabled is false   
       enabled = true
 ```
-2. Else if you don't have ConfigMap, download the new ConfigMap from [here.](https://github.com/microsoft/OMS-docker/blob/ci_feature_prod/Kubernetes/container-azm-ms-agentconfig.yaml) & then set `enabled =true`
+2. Else if you don't have ConfigMap, download the new ConfigMap from [here.](https://github.com/microsoft/Docker-Provider/blob/ci_dev/kubernetes/container-azm-ms-agentconfig.yaml) & then set `enabled =true`
```
rashmichandrashekar, line 18: this should be pointing to ci_prod.

rashmichandrashekar (approved):

PR #466, opened 2020-10-29
fixing the windows-perf-dups

rashmichandrashekar:


source/plugins/ruby/in_win_cadvisor_perf.rb
```diff
@@ -70,6 +69,7 @@ def enumerate()
           @@winNodeQueryTimeTracker = DateTime.now.to_time.to_i
         end
         @@winNodes.each do |winNode|
+          eventStream = MultiEventStream.new
```
vishiy, line 72: the insightsmetrics one is also adding (for GPU). Currently gpu sku is not supported on windows, but good to new that on every iteration as well :)

rashmichandrashekar: oh yeah. i missed that. i will fix it.

gracewehner: Yes I think the PV metrics duplicate too. I had tested the PV changes with only one windows node...so didn't notice that this would happen. Good to know I should always test with multiple 👍

rashmichandrashekar: @gracewehner - i think this should fix the PV metrics too right?? Did i miss something else?

gracewehner: No you're good, moving the insightsmetricseventstream = MultiEventStream.new will fix it, just wanted to clarify :)

rashmichandrashekar: okay.. no worries 👍

vishiy (approved):

gracewehner (approved):

PR #471, opened 2020-11-11
grwehner/mdm custom metric regions

gracewehner:
- Remove custom metrics region check for public cloud by removing the custom_metrics_azure_regions variable for plugins and changing to only check the cloud environment. This is set as an environment variable using the workspace domain in the main.sh file so that it'll work with AKS and ARC.

My questions:
Should the check be more explicit by specifically checking if the workspace domain is indeed for a national cloud? Should custom metrics be available for any other cluster with a workspace domain that isn't opinsights.azure.com? What should the strategy be if the domain is not set?

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -150,6 +150,17 @@ else
       echo "LA Onboarding:Workspace Id not mounted, skipping the telemetry check"
 fi
 
+# Set environment variable for if public cloud by checking the workspace domain.
+if [-z $DOMAIN]; then
+  $ClOUD_ENVIRONMENT = "unknown"
+elif [$DOMAIN == "opinsights.azure.com"]; then
+  $CLOUD_ENVIRONMENT = "public"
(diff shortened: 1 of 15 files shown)


vishiy (dismissed):

kubernetes/linux/main.sh
```diff
@@ -150,6 +150,17 @@ else
       echo "LA Onboarding:Workspace Id not mounted, skipping the telemetry check"
 fi
 
+# Set environment variable for if public cloud by checking the workspace domain.
+if [ -z $domain ]; then
+  ClOUD_ENVIRONMENT="unknown"
+elif [ $domain == "opinsights.azure.com" ]; then
```
rashmichandrashekar, line 156: i think as we are starting to add more constants here, we can move the string constants to a file. Not in the scope of this PR. But we can discuss and make the changes later.

rashmichandrashekar (approved):

PR #473, opened 2020-12-02
Gangams/fix rs ooming

ganga1980:
Changes related to OOM fix for the replicaset 

-  prefer affinity  to schedule rs pod on to ephemeral node if available
-  use RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR to 0.9 to make ruby trigger full GC 
-  adjust reader threads to 2 for Kube pod inventory, Kube Node inventory, Perf and Events
-  lower chunk size to get lower payload from api-server for resources in rs pod
-  lower write size to fluentd to avoid bottle neck with large buffer file writes 
- optimize the service list records and also fixed two issues related to kubeservices 1) multiple emits 2) putting wrong service name stamp if the service has 0 selectors.
-  refactored the code to avoid loading of PodInventory and NodeInventory through multiple enumerations
- Make chunk and emit stream sizes are configurable via configmap
- Use server side filtering to get the windows nodes


build/linux/installer/scripts/tomlparser-health-config.rb
```diff
@@ -13,32 +13,83 @@
 @configMapMountPath = "/etc/config/settings/agent-settings"
 @configSchemaVersion = ""
 @enable_health_model = false
+@nodesChunkSize = 0
+@podsChunkSize = 0
+@eventsChunkSize = 0
+@deploymentsChunkSize = 0
+@hpaChunkSize = 0
+@podsEmitStreamBatchSize = 0
```
vishiy, line 21: can chunk & emit size be same ? Whats the difference in perf we see for default pods (500 chunk, 200 emit, v 500 chunk. 500 emit) ?

build/linux/installer/scripts/tomlparser-health-config.rb
```diff
@@ -13,32 +13,83 @@
 @configMapMountPath = "/etc/config/settings/agent-settings"
```
vishiy, line 13: please rename this file to agent-settings

kubernetes/linux/Dockerfile
```diff
@@ -15,6 +15,7 @@ ENV HOST_VAR /hostfs/var
 ENV AZMON_COLLECT_ENV False
 ENV KUBE_CLIENT_BACKOFF_BASE 1
 ENV KUBE_CLIENT_BACKOFF_DURATION 0
+ENV RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR 0.9
```
vishiy, line 18: how much mem improvements you see here due to this, and also whats the cost of cpu due to frequent gc (especially on smaller clusters) ?

build/linux/installer/scripts/tomlparser-health-config.rb
```diff
@@ -13,32 +13,83 @@
...
 @configSchemaVersion = ""
 @enable_health_model = false
+@nodesChunkSize = 0
```
vishiy, line 16: Cleaner to define the default here ( instead of zero ) , and always set it (to default if not config map), and remove the defaults in each plugin, so this is all in one place.

build/linux/installer/scripts/tomlparser-health-config.rb
```diff
@@ -13,32 +13,83 @@
...
 # Use the ruby structure created after config parsing to set the right values to be used as environment variables
 def populateSettingValuesFromConfigMap(parsedConfig)
   begin
-    if !parsedConfig.nil? && !parsedConfig[:agent_settings].nil? && !parsedConfig[:agent_settings][:health_model].nil? && !parsedConfig[:agent_settings][:health_model][:enabled].nil?
+    if !parsedConfig.nil? && !parsedConfig[:agent_settings].nil?
+      if !parsedConfig[:agent_settings][:health_model].nil? && !parsedConfig[:agent_settings][:health_model][:enabled].nil?
         @enable_health_model = parsedConfig[:agent_settings][:health_model][:enabled]
         puts "enable_health_model = #{@enable_health_model}"
+      end
+      chunk_config = parsedConfig[:agent_settings][:chunk_config]
```
vishiy, line 55: i dont see a nil check. Have tested by setting the last setting in this code sequence and see if it gets set as expected ?

build/linux/installer/scripts/tomlparser-health-config.rb
```diff
@@ -13,32 +13,83 @@
...
+      if !chunk_config.nil?
+        nodesChunkSize = chunk_config[:NODES_CHUNK_SIZE]
+        if !nodesChunkSize.nil? && is_number?(nodesChunkSize)
```
vishiy, line 58: check between a min & max, so one cannot ddos the api server by say setting this setting to 1.

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -276,7 +280,8 @@ def getPods(namespace)
     def getWindowsNodes
       winNodes = []
       begin
-        resourceUri = getNodesResourceUri("nodes")
+        # get only windows nodes
+        resourceUri = getNodesResourceUri("nodes?labelSelector=kubernetes.io%2Fos%3Dwindows")
```
vishiy, line 284: what happens for older nodes with beta label? Its not good to hardcode label selectors in code as labels are highly volatile.

source/plugins/ruby/in_kubestate_deployments.rb
```diff
@@ -2,230 +2,237 @@
...
+        @deploymentsRunningTotal = 0
+
+        # Initializing continuation token to nil
+        continuationToken = nil
+        $log.info("in_kubestate_deployments::enumerate : Getting deployments from Kube API @ #{Time.now.utc.iso8601}")
+        continuationToken, deploymentList = KubernetesApiClient.getResourcesAndContinuationToken("deployments?limit=#{@DEPLOYMENTS_CHUNK_SIZE}", api_group: @DEPLOYMENTS_API_GROUP)
+        $log.info("in_kubestate_deployments::enumerate : Done getting deployments from Kube API @ #{Time.now.utc.iso8601}")
+        if (!deploymentList.nil? && !deploymentList.empty? && deploymentList.key?("items") && !deploymentList["items"].nil? && !deploymentList["items"].empty?)
+          # debug logs to track the payload size
+          deploymentsSizeInKB = (deploymentList.to_s.length) / 1024
```
vishiy, line 84: you are calculating string length, not storage size with sizeof. Is this how you estimated payload size ?

source/plugins/ruby/in_kube_podinventory.rb
```diff
@@ -535,25 +394,238 @@ def run_periodic
...
+                selectorLabels = kubeServiceRecord["SelectorLabels"][0]
+              end
+              if !selectorLabels.nil? && !selectorLabels.empty?
+                selectorLabels.each do |key, value|
+                  if !(labels.select { |k, v| k == key && v == value }.length > 0)
+                    break
                   end
+                  found = found + 1
                 end
+                # service can have no selectors
```
vishiy, line 626: can you explain this change ? We discover only services based on label selectors.

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -396,42 +401,67 @@ def getPodUid(podNameSpace, podMetadata)
...
-          if !pod["spec"]["initContainers"].nil? && !pod["spec"]["initContainers"].empty?
-            podContainers = podContainers + pod["spec"]["initContainers"]
-          end
+        nodeName = ""
+        #for unscheduled (non-started) pods nodeName does NOT exist
+        if !pod["spec"]["nodeName"].nil?
+          nodeName = pod["spec"]["nodeName"]
+        end
+        # For ARO, skip the pods scheduled on to master or infra nodes to ingest
+        if isAROv3MasterOrInfraPod(nodeName)
```
vishiy, line 420: these are all un-necessary changes, that just triggers testing ARO. I see the same for features like GPU. Its ok to touch, but please test & validate them

source/plugins/ruby/in_kube_podinventory.rb
```diff
@@ -122,275 +163,158 @@ def enumerate(podList = nil)
...
           ApplicationInsightsUtility.sendCustomEvent("KubePodInventoryHeartBeatEvent", telemetryProperties)
           ApplicationInsightsUtility.sendMetricTelemetry("PodCount", @podCount, {})
+          ApplicationInsightsUtility.sendMetricTelemetry("ServiceCount", @serviceCount, {})
           telemetryProperties["ControllerData"] = @controllerData.to_json
           ApplicationInsightsUtility.sendMetricTelemetry("ControllerCount", @controllerSet.length, telemetryProperties)
           if @winContainerCount > 0
             telemetryProperties["ClusterWideWindowsContainersCount"] = @winContainerCount
             ApplicationInsightsUtility.sendCustomEvent("WindowsContainerInventoryEvent", telemetryProperties)
           end
+          ApplicationInsightsUtility.sendMetricTelemetry("PodInventoryE2EProcessingLatencyMs", @podInventoryE2EProcessingLatencyMs, telemetryProperties)
```
vishiy, line 177: what is the use of these 2 metrics as you are just getting the last one for the last chunk ? What we need is keep track of an agg (like max,p95) during the telemetryFlush interval, and send send that as agg for that interval.

build/linux/installer/scripts/tomlparser-health-config.rb
ganga1980: Done.

kubernetes/linux/Dockerfile
ganga1980: Around ~100MB in large clusters and verified that no significant impact  on the CPU on small cluster (26 nodes and 890 pods).  Here is the cpu usage of the  cluster with RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR = 0.9 and without this.
image

build/linux/installer/scripts/tomlparser-health-config.rb
ganga1980: are you referring "nil" check for chunk_config? if yes, its there already and all nil checks are in place. lmk, if i am missing anything. yes, verified code sequence and no bail out etc..

build/linux/installer/scripts/tomlparser-health-config.rb
ganga1980: Sure, make sense. thanks for the feedback. Added min and max checks for both chunk and emit sizes.

source/plugins/ruby/KubernetesApiClient.rb
ganga1980: This standard node label and which is available starting from 1.14 https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.14.md#deprecations and verified the same.  From our telemetry, we have 2 clusters under < 1.14, out of this one already failed provisioning state and other one doesnt have any windows nodes from AKS telemetry. From k8s doc, windows container supported added in 1.14 https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/ so I think, we are good on this.

source/plugins/ruby/in_kubestate_deployments.rb
ganga1980: AFIK, there is no sizeof operator in ruby. are you referring https://apidock.com/ruby/String/bytesize as sizeof ?  btw, this is just estimate for the debug and will remove these as not needed for prod.

source/plugins/ruby/in_kube_podinventory.rb
ganga1980: as discussed in the email, this is just not  tagging service name if the service has zero selectors and service and pod are in the same namespace.  saw one of Prometheus service doesnt have selectors. since initial value of found =0, found == selectorLabels.length. This will be true if the length of service labal selector is 0.

source/plugins/ruby/KubernetesApiClient.rb
ganga1980: As we discussed in the email, these are changes as part of emitting smaller streams and perf optimizations. Sure, will ensure no regressions.

source/plugins/ruby/in_kube_podinventory.rb
ganga1980: > what is the use of these 2 metrics as you are just getting the last one for the last chunk ? What we need is keep track of an agg (like max,p95) during the telemetryFlush interval, and send send that as agg for that interval.

> similar to other metrics, just emitting at the flush intervals and please note, these metrics are emitted not just last chunk rather accumulated values of the all the chunks. This will help us to understand our ingestion latency where we spending.

build/linux/installer/scripts/tomlparser-health-config.rb
ganga1980: Yes, chunk and emit can be same size. Emit size controls the size of the file write and where as chunk size helps on the memory usage.  Based on the testing lower chunk size gives lower memory usage but that costs additional round trip latency.   To keep the ingestion latencies under 1 min, default value of pod chunk is 1K and pod emit size 200 items. Added the logic to validate the lower and upper bounds of these values and also added comments around this.

build/linux/installer/scripts/tomlparser-health-config.rb
ganga1980: Done.

charts/azuremonitor-containers/templates/omsagent-rs-configmap.yaml
```diff
@@ -108,24 +108,24 @@ data:
...
-     buffer_path %STATE_DIR_WS%/state/out_oms_kubepv*.buffer
-     buffer_queue_limit 20
-     buffer_queue_full_action drop_oldest_chunk
-     flush_interval 20s
-     retry_limit 10
-     retry_wait 5s
-     max_retry_wait 5m
+      type out_oms
+      log_level debug
+      num_threads 5
```
gracewehner, line 113: Does this also need to be changed to 'num_threads 2' like KubePodInventory, KubeEvents etc. ? I also have it as 5 in omsagent.yaml

ganga1980: > PV is not high volume data so I think we are fine with num_threads as 5 for PV.  If the data type is high volume then higher number of threads will make size of the file reads are large which will cause OOM kill if there is not enough memory. The optimizations which I have done are adjusting number of threads to avoid large file reads which in turn handle the OOM killer situtation.

daweim0 (approved):

gracewehner (approved):

PR #481, opened 2021-01-01
grwehner/jan agent tasks

gracewehner:
Tasks:
- Windows agent fix to use log filtering settings in config map.
- Error handling for kubelet_utils get_node_capacity in case /metrics/cadvsior endpoint fails.
- Look into not using environment var for workspace key: Looks like it's only used as an env var for windows agent and only by the certificate generator. Made a change to remove the env var and for the certificate generator to read directly from the file that the secret is mounted to instead. This change would affect customers if they are using this environment variable to get the workspace key.

### DIFF
--- a/build/windows/installer/conf/fluent.conf
+++ b/build/windows/installer/conf/fluent.conf
@@ -29,15 +29,6 @@
   @include fluent-docker-parser.conf
 </source>
 
-<filter  oms.container.**>
-  @type record_transformer
-  # fluent-plugin-record-modifier more light-weight but needs to be installed (dependency worth it?)
-  remove_keys tailed_path
-  <record>
-    filepath ${record["tailed_path"]}
-  </record>
-</filter>
-
 <filter oms.container.log.la>
   @type grep
   <exclude>
@@ -46,6 +37,14 @@
   </exclude>
 </filter>
 
+<filter  oms.container.**>
+  @type record_transformer
+  # fluent-plugin-record-modifier more light-weight but needs to be installed (dependency worth it?)
+  remove_keys tailed_path
+  <record>
+    filepath ${record["tailed_path"]}
+  </record>
+</filter>
 
 <match oms.container.**>
   @type forward
--- a/build/windows/installer/conf/fluent.conf
+++ b/build/windows/installer/conf/fluent.conf
@@ -7,6 +7,7 @@
 <source>
   @type tail
   path "#{ENV['AZMON_LOG_TAIL_PATH']}"
+  exclude_path "#{ENV['AZMON_CLUSTER_LOG_TAIL_EXCLUDE_PATH']}"
   pos_file /var/opt/microsoft/fluent/fluentd-containers.log.pos
   tag oms.container.log.la
   @log_level trace
@@ -35,6 +36,14 @@
   <record>
     filepath ${record["tailed_path"]}
   </record>
+</filter>
+
+<filter oms.container.log.la>
+  @type grep
+  <exclude>
+    key stream
+    pattern "#{ENV['AZMON_LOG_EXCLUSION_REGEX_PATTERN']}"
+  </exclude>
 </filter>
 
 
--- a/build/windows/installer/conf/fluent.conf
+++ b/build/windows/installer/conf/fluent.conf
@@ -6,7 +6,7 @@
 
 <source>
   @type tail
-  path /var/log/containers/*.log
+  path "#{ENV['AZMON_LOG_TAIL_PATH']}"
   pos_file /var/opt/microsoft/fluent/fluentd-containers.log.pos
   tag oms.container.log.la
   @log_level trace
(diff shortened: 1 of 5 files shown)


source/plugins/ruby/filter_cadvisor2mdm.rb
```diff
@@ -308,8 +308,16 @@ def ensure_cpu_memory_capacity_set
...
-        @cpu_capacity = capacity_from_kubelet[0]
-        @memory_capacity = capacity_from_kubelet[1]
+
+        # Error handling in case /metrics/cadvsior endpoint fails
+        if !capacity_from_kubelet.nil? && capacity_from_kubelet.length > 1
+          @cpu_capacity = capacity_from_kubelet[0]
+          @memory_capacity = capacity_from_kubelet[1]
+        else
+          # cpu_capacity and memory_capacity keep initialized value of 0.0
+          @log.info "Error getting capacity_from_kubelet: cpu_capacity and memory_capacity"
```
ganga1980, line 318: should be the error rather than info and we should also emit to AI telemetry so that it will be helpful to handle the customer incidents.

build/windows/installer/conf/fluent.conf
```diff
@@ -37,6 +38,14 @@
   </record>
 </filter>
 
+<filter oms.container.log.la>
```
ganga1980, line 41: Since we have multiple filter plugins. is this filter invoked before record_transformer filter plugin?

ganga1980 (changes requested):

build/windows/installer/conf/fluent.conf
gracewehner: Looks like from the docs the filters are evaluated top-down. I'll move this filter before the record_transformer one.

source/plugins/ruby/filter_cadvisor2mdm.rb
gracewehner: I will change it to @log.error. I believe AI telemetry will be sent here when KubeletUtils.get_node_capacity fails: https://github.com/microsoft/Docker-Provider/blob/ef9d726c7053fba0254fc897aff124e5a5a2be34/source/plugins/ruby/kubelet_utils.rb#L33

build/windows/installer/conf/fluent.conf
ganga1980: > Looks like from the docs the filters are evaluated top-down. I'll move this filter before the record_transformer one.

> I see.  can you please point me the doc about "top-down" evaluation for the filter plugins?

source/plugins/ruby/filter_cadvisor2mdm.rb
ganga1980: > I will change it to @log.error. I believe AI telemetry will be sent here when KubeletUtils.get_node_capacity fails:
> 
> https://github.com/microsoft/Docker-Provider/blob/ef9d726c7053fba0254fc897aff124e5a5a2be34/source/plugins/ruby/kubelet_utils.rb#L33

> Ok, then pls change either error or warn instead of info here. I see many places in existing code, we use info instead of error/warn for the error logs.

build/windows/installer/conf/fluent.conf
gracewehner: Yes I read this [here](https://docs.fluentd.org/filter#how-to-use) starting at:
> Like the <match> directive for output plugins, <filter> matches against a tag. Once the event is processed by the filter, the event proceeds through the configuration top-down. Hence, if there are multiple filters for the same tag, they are applied in descending order. Hence, in the following example:

ganga1980 (approved):

kubernetes/windows/main.ps1
```diff
@@ -64,19 +64,11 @@ function Set-EnvironmentVariables {
         $wsID = Get-Content /etc/omsagent-secret/WSID
     }
 
-    # Set DOMAIN
+    # Set WSID
     [System.Environment]::SetEnvironmentVariable("WSID", $wsID, "Process")
     [System.Environment]::SetEnvironmentVariable("WSID", $wsID, "Machine")
 
-    $wsKey = ""
```
vishiy, line 71: Grace, There is another place this is going thru ENV (one agent). Please keep this work item active (or close & recreate a new one), just so that we can track it.

gracewehner: Thanks Vishwa, will do

build/windows/installer/conf/fluent.conf
vishiy: order should not matter here, as long as the field 'stream' which this filter looks for is not renamed/removed by other filters.

vishiy (approved):

build/windows/installer/conf/fluent.conf
ganga1980: i think, order helps on the perf side for example, in this case, if the logs filtering happens first and then subsequent filters will have less number of records to process, right? This will be helpful in high volume log scenario.

vishiy: Can you point to where you see this ? All of them are chained to same stream of data (data not passed/duped between filters). So as long as data reaches the stream you have it there. Thats why we don't tail excluded namespace files (rather than getting them here & dropping those).

PR #483, opened 2021-01-06
fix ci envvar collection in large pods

ganga1980:
This fix addresses inconsistency in collection of the environment variables if the pod has more than > 40 containers.  Because of this inconsistency ci_dev cluster has huge drop in the container inventory if ended up getting no envvars for the large pods where pod has high number of containers. For example, on dev cluster, "many-containers-in-pod" has more than 100 containers.

### DIFF
--- a/source/plugins/ruby/kubernetes_container_inventory.rb
+++ b/source/plugins/ruby/kubernetes_container_inventory.rb
@@ -193,25 +193,41 @@ class KubernetesContainerInventory
       $log.info("KubernetesContainerInventory::obtainContainerEnvironmentVars @ #{Time.now.utc.iso8601}")
       envValueString = ""
       begin
-        unless @@containerCGroupCache.has_key?(containerId)
+        isCGroupPidFetchRequired = false 
+        if !@@containerCGroupCache.has_key?(containerId)
+          isCGroupPidFetchRequired = true 
+        else
+          cGroupPid = @@containerCGroupCache[containerId]
+          if cGroupPid.nil? || cGroupPid.empty?            
+            isCGroupPidFetchRequired = true
+            @@containerCGroupCache.delete(containerId)
+          elsif !File.exist?("/hostfs/proc/#{cGroupPid}/environ")              
+            isCGroupPidFetchRequired = true
+            @@containerCGroupCache.delete(containerId)                       
+          end        
+        end
+
+        if isCGroupPidFetchRequired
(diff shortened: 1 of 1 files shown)


source/plugins/ruby/kubernetes_container_inventory.rb
```diff
@@ -193,25 +193,41 @@ def obtainContainerEnvironmentVars(containerId)
...
+            @@containerCGroupCache.delete(containerId)                       
+          end        
+        end
+
+        if isCGroupPidFetchRequired
           $log.info("KubernetesContainerInventory::obtainContainerEnvironmentVars fetching cGroup parent pid @ #{Time.now.utc.iso8601} for containerId: #{containerId}")
           Dir["/hostfs/proc/*/cgroup"].each do |filename|
             begin
-              if File.file?(filename) && File.foreach(filename).grep(/#{containerId}/).any?
+              if File.file?(filename) && File.exist?(filename) && File.foreach(filename).grep(/#{containerId}/).any?
```
vishiy, line 214: What has 'more 40 containers in a pod' do with this ? How does this fix take care of that ?

ganga1980: > What has 'more than 40 containers in a pod' do with this ? How does this fix take care of that ?

> if the pod has more than ~40 containers, then cproc process was deleted and created new process and this happening for the pods which are high number of containers. Very first time we are able to get envvar from the cproc pid and there after that process gone so we are getting empty env vars . This was reproducible with pod ~40 containers. Fix to refetch cproc pid if the file cproc doesnt exist any more and use parent process and also added logic to ignore the cproc which are not numeric pods .

vishiy: are you saying pids have changed for the same container between last time & current ? are the containers restarting ?

ganga1980: > are you saying pids have changed for the same container between last time & current ? are the containers restarting ?

> correct. This reproducible with large pods (>40 containers also, pod has same container image across all the containers) and no container/pod restart. This was the yaml which deployed to dev cluster and used the same https://raw.githubusercontent.com/microsoft/Docker-Provider/ci_dev/test/scenario/yamls/many-containers-in-pod.yaml for repro and validating the fix with different containers. Fix to refetch cgroup id and also use parent id.

vishiy: I dont know what this 40 is about. But based on what you are saying, for PIds that are reused, you will be picking up wrong envs due to wrong mappin in the cache. I will approve this PR for now, but i am not sure whats the issue here and we should understand the root cause of it.

vishiy (approved):

PR #486, opened 2021-01-08
Gangams/add fbsettings configurable via configmap

ganga1980:
adding fbit settings FBIT_SERVICE_FLUSH_INTERVAL, FBIT_TAIL_BUFFER_CHUNK_SIZE and FBIT_TAIL_BUFFER_MAX_SIZE configurable via configmap.  These config settings will be under 
Here is the final naming 
 agent-settings: |- 
    [agent_settings.fbit_config]   
      log_flush_interval_secs = "1"	
      tail_buf_chunksize_megabytes= "1"  
      tail_buf_maxsize_megabytes = "1"

### DIFF
--- a/build/linux/installer/scripts/tomlparser-agent-config.rb
+++ b/build/linux/installer/scripts/tomlparser-agent-config.rb
@@ -153,20 +153,18 @@ def populateSettingValuesFromConfigMap(parsedConfig)
         end
 
         fbitTailBufferMaxSizeMBs = fbit_config[:tail_buf_maxsize_megabytes]
-        if !fbitTailBufferMaxSizeMBs.nil? && is_number?(fbitTailBufferMaxSizeMBs) && fbitTailBufferMaxSizeMBs.to_i > 0 
+        if !fbitTailBufferMaxSizeMBs.nil? && is_number?(fbitTailBufferMaxSizeMBs) && fbitTailBufferMaxSizeMBs.to_i > 0           
           if fbitTailBufferMaxSizeMBs.to_i >= @fbitTailBufferChunkSizeMBs
             @fbitTailBufferMaxSizeMBs = fbitTailBufferMaxSizeMBs.to_i
             puts "Using config map value: tail_buf_maxsize_megabytes = #{@fbitTailBufferMaxSizeMBs}"
           else
+            # tail_buf_maxsize_megabytes has to be greater or equal to tail_buf_chunksize_megabytes
             @fbitTailBufferMaxSizeMBs = @fbitTailBufferChunkSizeMBs
-            puts "config::warn: tail_buf_maxsize_megabytes should be greater or equal to value of tail_buf_chunksize_megabytes. Using tail_buf_maxsize_megabytes = #{@fbitTailBufferMaxSizeMBs} since provided config value not valid"
+            puts "config::warn: tail_buf_maxsize_megabytes must be greater or equal to value of tail_buf_chunksize_megabytes. Using tail_buf_maxsize_megabytes = #{@fbitTailBufferMaxSizeMBs} since provided config value not valid"
           end
         end
-         # in scenario - tail_buf_maxsize_megabytes provided but not tail_buf_chunksize_megabytes and vice versa
-        if @fbitTailBufferMaxSizeMBs > 0 && @fbitTailBufferChunkSizeMBs == 0
-          @fbitTailBufferChunkSizeMBs = @fbitTailBufferMaxSizeMBs
-          puts "config::warn: since tail_buf_chunksize_megabytes not provided hence using tail_buf_chunksize_megabytes=#{@fbitTailBufferChunkSizeMBs} which is same as the value of tail_buf_maxsize_megabytes"
-        elsif  @fbitTailBufferChunkSizeMBs > 0  && @fbitTailBufferMaxSizeMBs == 0
+        # in scenario - tail_buf_chunksize_megabytes provided but not tail_buf_maxsize_megabytes to prevent fbit crash
+        if  @fbitTailBufferChunkSizeMBs > 0  && @fbitTailBufferMaxSizeMBs == 0
           @fbitTailBufferMaxSizeMBs = @fbitTailBufferChunkSizeMBs
           puts "config::warn: since tail_buf_maxsize_megabytes not provided hence using tail_buf_maxsize_megabytes=#{@fbitTailBufferMaxSizeMBs} which is same as the value of tail_buf_chunksize_megabytes"
         end 
--- a/build/linux/installer/scripts/tomlparser-agent-config.rb
+++ b/build/linux/installer/scripts/tomlparser-agent-config.rb
@@ -55,6 +55,12 @@ require_relative "ConfigParseErrorLogger"
 @podsEmitStreamBatchSizeMin = 50
 @nodesEmitStreamBatchSizeMin = 50
 
+# configmap settings related fbit config
+@fbitFlushInterval = 0
+@fbitTailBufferChunkSize = 0
+@fbitTailBufferMaxSize = 0
+
+
 def is_number?(value)
   true if Integer(value) rescue false
 end
@@ -130,6 +136,36 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           @nodesEmitStreamBatchSize = nodesEmitStreamBatchSize.to_i
           puts "Using config map value: NODES_EMIT_STREAM_BATCH_SIZE = #{@nodesEmitStreamBatchSize}"
         end
+      end
+      fbit_config = parsedConfig[:agent_settings][:fbit_config]
+      if !fbit_config.nil?
+        fbitFlushInterval = fbit_config[:FBIT_SERVICE_FLUSH_INTERVAL]
+        if !fbitFlushInterval.nil? && is_number?(fbitFlushInterval) && fbitFlushInterval.to_i > 0
+          @fbitFlushInterval = fbitFlushInterval.to_i
+          puts "Using config map value: FBIT_SERVICE_FLUSH_INTERVAL = #{@fbitFlushInterval}"
+        end
+
+        fbitTailBufferChunkSize = fbit_config[:FBIT_TAIL_BUFFER_CHUNK_SIZE]
+        if !fbitTailBufferChunkSize.nil? && is_number?(fbitTailBufferChunkSize) && fbitTailBufferChunkSize.to_i > 0
+          @fbitTailBufferChunkSize = fbitTailBufferChunkSize.to_i
+          puts "Using config map value: FBIT_TAIL_BUFFER_CHUNK_SIZE = #{@fbitTailBufferChunkSize}"
+        end
+
+        fbitTailBufferMaxSize = fbit_config[:FBIT_TAIL_BUFFER_MAX_SIZE]
+        if !fbitTailBufferMaxSize.nil? && is_number?(fbitTailBufferMaxSize) && fbitTailBufferMaxSize.to_i > 0 
+          if fbitTailBufferMaxSize.to_i >= @fbitTailBufferChunkSize
+            @fbitTailBufferMaxSize = fbitTailBufferMaxSize.to_i
+            puts "Using config map value: FBIT_TAIL_BUFFER_MAX_SIZE = #{@fbitTailBufferMaxSize}"
+          else
+            @fbitTailBufferMaxSize = @fbitTailBufferChunkSize
+            puts "config::warn: FBIT_TAIL_BUFFER_MAX_SIZE should be greater or equal to value of FBIT_TAIL_BUFFER_CHUNK_SIZE. Using FBIT_TAIL_BUFFER_MAX_SIZE = #{@fbitTailBufferMaxSize} since provodided config value not valid"
+          end
+        end
+         # in scenario - FBIT_TAIL_BUFFER_MAX_SIZE provided but not FBIT_TAIL_BUFFER_CHUNK_SIZE
+        if @fbitTailBufferMaxSize > 0 && @fbitTailBufferChunkSize == 0
+          @fbitTailBufferChunkSize = @fbitTailBufferMaxSize
+        end 
+
       end
     end
   rescue => errorStr
--- a/build/linux/installer/scripts/tomlparser-agent-config.rb
+++ b/build/linux/installer/scripts/tomlparser-agent-config.rb
@@ -162,10 +162,13 @@ def populateSettingValuesFromConfigMap(parsedConfig)
             puts "config::warn: FBIT_TAIL_BUFFER_MAX_SIZE should be greater or equal to value of FBIT_TAIL_BUFFER_CHUNK_SIZE. Using FBIT_TAIL_BUFFER_MAX_SIZE = #{@fbitTailBufferMaxSize} since provodided config value not valid"
           end
         end
-         # in scenario - FBIT_TAIL_BUFFER_MAX_SIZE provided but not FBIT_TAIL_BUFFER_CHUNK_SIZE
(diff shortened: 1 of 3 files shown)


build/linux/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -164,6 +204,16 @@ def populateSettingValuesFromConfigMap(parsedConfig)
   file.write("export HPA_CHUNK_SIZE=#{@hpaChunkSize}\n")
   file.write("export PODS_EMIT_STREAM_BATCH_SIZE=#{@podsEmitStreamBatchSize}\n")
   file.write("export NODES_EMIT_STREAM_BATCH_SIZE=#{@nodesEmitStreamBatchSize}\n")
+  # fbit settings
+  if @fbitFlushInterval > 0
+    file.write("export FBIT_SERVICE_FLUSH_INTERVAL=#{@fbitFlushInterval}\n")
```
vishiy, line 209: we support only megs & secs. so these should have explicit units.

ganga1980: > we support only megs & secs. so these should have explicit units.

For chunksize(s), "m" being appened by https://github.com/microsoft/Docker-Provider/blob/ci_dev/build/common/installer/scripts/td-agent-bit-conf-customizer.rb and for flush interval, there is no unit for the interval unit suffix. Flush interval value supported in seconds. I kept the same contract as we support for hyrbid via environment variables.

Basically, here in this file, we are parsing configmap and setting the env vars, which will be inturn parsed by td-agent-bit-conf-customizer.rb code and setting the td-agent conf file.

vishiy: External setting names must have units. How will people know what FBIT_TAIL_BUFFER_MAX_SIZE=1 mean ? 1k ? 1g? 1m?

build/linux/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -164,6 +204,16 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+  end
+  if @fbitTailBufferChunkSize > 0
+    file.write("export FBIT_TAIL_BUFFER_CHUNK_SIZE=#{@fbitTailBufferChunkSize}\n")
+  end
+  if @fbitTailBufferMaxSize > 0
+    file.write("export FBIT_TAIL_BUFFER_MAX_SIZE=#{@fbitTailBufferMaxSize}\n")
```
rashmichandrashekar, line 215: As we didnt have a dependency between any parser scripts before, we need to add a note in the main.sh file so that we don't move around the scripts and miss the ordering.

build/linux/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -131,6 +137,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+
+        fbitTailBufferChunkSize = fbit_config[:FBIT_TAIL_BUFFER_CHUNK_SIZE]
+        if !fbitTailBufferChunkSize.nil? && is_number?(fbitTailBufferChunkSize) && fbitTailBufferChunkSize.to_i > 0
+          @fbitTailBufferChunkSize = fbitTailBufferChunkSize.to_i
+          puts "Using config map value: FBIT_TAIL_BUFFER_CHUNK_SIZE = #{@fbitTailBufferChunkSize}"
+        end
+
+        fbitTailBufferMaxSize = fbit_config[:FBIT_TAIL_BUFFER_MAX_SIZE]
+        if !fbitTailBufferMaxSize.nil? && is_number?(fbitTailBufferMaxSize) && fbitTailBufferMaxSize.to_i > 0 
+          if fbitTailBufferMaxSize.to_i >= @fbitTailBufferChunkSize
```
vishiy, line 157: can you explain this optimization and where it says it cannot be > ?

build/linux/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -131,6 +137,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+            @fbitTailBufferMaxSize = fbitTailBufferMaxSize.to_i
+            puts "Using config map value: FBIT_TAIL_BUFFER_MAX_SIZE = #{@fbitTailBufferMaxSize}"
+          else
+            @fbitTailBufferMaxSize = @fbitTailBufferChunkSize
+            puts "config::warn: FBIT_TAIL_BUFFER_MAX_SIZE should be greater or equal to value of FBIT_TAIL_BUFFER_CHUNK_SIZE. Using FBIT_TAIL_BUFFER_MAX_SIZE = #{@fbitTailBufferMaxSize} since provodided config value not valid"
+          end
+        end
+         # in scenario - FBIT_TAIL_BUFFER_MAX_SIZE provided but not FBIT_TAIL_BUFFER_CHUNK_SIZE and vice versa
```
rashmichandrashekar, line 165: The current behavior is to not set the values in case they are not provided. Why are we setting chunk to max and max to chunk if either is not provided?
Essentially the customer would want to see the exact configuration settings they are providing using configmap.

rashmichandrashekar: I think the logic should remain the same as here. https://github.com/microsoft/Docker-Provider/blob/ci_dev/build/common/installer/scripts/td-agent-bit-conf-customizer.rb

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > can you explain this optimization and where it says it cannot be > ?

Look at these,
https://fluentbit.io/documentation/0.13/input/tail.html, default value of Buffer_Max_Size is the same as Buffer_Chunk_Size. 
https://github.com/fluent/fluent-bit/blob/master/plugins/in_tail/tail_config.c#L165

vishiy: if its not specified its the same. where oes it say it cannot be > ?

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > As we didnt have a dependency between any parser scripts before, we need to add a note in the main.sh file so that we don't move around the scripts and miss the ordering.

> great catch. completely make sense. let me add the comment.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > The current behavior is to not set the values in case they are not provided. Why are we setting chunk to max and max to chunk if either is not provided?
> Essentially the customer would want to see the exact configuration settings they are providing using configmap.

> Buffer_Chunk_Size and Buffer_Chunk_Max_Size are tightly coupled. as pointed out  here
https://fluentbit.io/documentation/0.13/input/tail.html, default value of Buffer_Max_Size is the same as Buffer_Chunk_Size.
https://github.com/fluent/fluent-bit/blob/master/plugins/in_tail/tail_config.c#L165
This is to ensure no undesired effects in case customer provides one value but not other value. If the customer provides both the values and those will be honored as long as meet this criterion buffer_max_size must be >= buffer_chunk.

rashmichandrashekar: 1. If the value is not set for Buffer_Max_Size fluentbit already takes the default value as the Buffer chunk size.
2. If the value is not set for Buffer_Chunk_Size fluentbit takes 32k as the default. Why should we set this to Buffer max size if it is not provided?

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > if its not specified its the same. where does it say it cannot be > ?

> here https://github.com/fluent/fluent-bit/blob/master/plugins/in_tail/tail_config.c#L165.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > 1. If the value is not set for Buffer_Max_Size fluentbit already takes the default value as the Buffer chunk size.
>   where do you see this? can you please point me the code or doc  so that i can validate before removing this additional code ? if it does what you say, then I can safely remove.

> 2. If the value is not set for Buffer_Chunk_Size fluentbit takes 32k as the default. Why should we set this to Buffer max size if it is not provided?
>  To avoid undesired effects when Buffer_Chunk_Size is larger than default max size. If you prove that Buffer_Max_Size will be set same as Buffer_Chunk_Size  and vice versa then I am happy to remove this piece of the code though it does same what you are describing here.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > External setting names must have units. How will people know what FBIT_TAIL_BUFFER_MAX_SIZE=1 mean ? 1k ? 1g? 1m?

> agree, having the units suffix in name of the config key is self explanatory.  For example FBIT_TAIL_BUFFER_MAX_SIZE_MB but changing this will not keep the same naming contract what we support in HELM chart.   In configmap, these will be 

FBIT_SERVICE_FLUSH_INTERVAL_SECS
FBIT_TAIL_BUFFER_CHUNK_SIZE_MB
FBIT_TAIL_BUFFER_MAX_SIZE_MB

but where as in helm or yaml will be  without unit suffix. since changing these will break the existing customers

FBIT_SERVICE_FLUSH_INTERVAL
FBIT_TAIL_BUFFER_CHUNK_SIZE
FBIT_TAIL_BUFFER_MAX_SIZE

ganga1980: Fixed. Please have a look a it.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: Fixed.

build/linux/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -55,6 +55,12 @@
 @podsEmitStreamBatchSizeMin = 50
 @nodesEmitStreamBatchSizeMin = 50
 
+# configmap settings related fbit config
+@fbitFlushIntervalSecsSecs = 0
```
gracewehner, line 59: Should be @fbitFlushIntervalSecs I believe

build/linux/installer/scripts/tomlparser-agent-config.rb
rashmichandrashekar: 1. It is the same documentation for tail plugin you shared earlier - https://fluentbit.io/documentation/0.13/input/tail.html
Look under the table for the settings.
2. In your code "if @fbitTailBufferMaxSize > 0 && @fbitTailBufferChunkSize == 0
          @fbitTailBufferChunkSize = @fbitTailBufferMaxSize"
You are setting the chunk size to max size if it is not set. This is also in the same tail documentation.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > Should be @fbitFlushIntervalSecs I believe

> yep, your right. :) fixed. thanks for catching.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > 1. It is the same documentation for tail plugin you shared earlier - https://fluentbit.io/documentation/0.13/input/tail.html
>    Look under the table for the settings.
> 2. In your code "if @fbitTailBufferMaxSize > 0 && @fbitTailBufferChunkSize == 0
>    @fbitTailBufferChunkSize = @fbitTailBufferMaxSize"
>    You are setting the chunk size to max size if it is not set. This is also in the same tail documentation.

>  The doc only talking out the default value. do you know if buf_max_size will be buf_chunk_size if not proided? as I pointed out here, I think, if the value of the buf_chunk_size   is greater than buf_max_size  then the fbit will fail with an error https://github.com/fluent/fluent-bit/blob/master/plugins/in_tail/tail_config.c#L165. Let me do quick validation and see what's behavior.

rashmichandrashekar: In the changes I dont see code where you are checking if chunk size is greater than max size and setting chunk size to max size. The only change I see is if chunk size is not set, it is being set to max, which is not clear why. 
And yes the table says the max is set to chunk size by default.
This is the code I am talking about

if @fbitTailBufferMaxSizeMBs > 0 && @fbitTailBufferChunkSizeMBs == 0
          @fbitTailBufferChunkSizeMBs = @fbitTailBufferMaxSizeMBs
          puts "config::warn: since tailbufchunksizemegabytes not provided hence using tailbufchunksizemegabytes=#{@fbitTailBufferChunkSizeMBs} which is same as the value of tailbufmaxsizemegabytes"
        elsif  @fbitTailBufferChunkSizeMBs > 0  && @fbitTailBufferMaxSizeMBs == 0
          @fbitTailBufferMaxSizeMBs = @fbitTailBufferChunkSizeMBs
          puts "config::warn: since tailbufmaxsizemegabytes not provided hence using tailbufmaxsizemegabytes=#{@fbitTailBufferMaxSizeMBs} which is same as the value of tailbufchunksizemegabytes"
        end

ganga1980: just verified the fbit crashes if we only set tailbufchunksizemegabytes but not max one.

replacing nodename in telegraf config
File Doesnt Exist. Creating file...
←[1mFluent Bit v1.4.2←[0m
* ←[1m←[93mCopyright (C) 2019-2020 The Fluent Bit Authors←[0m
* ←[1m←[93mCopyright (C) 2015-2018 Treasure Data←[0m
* Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
* https://fluentbit.io

[engine] caught signal (SIGSEGV)
2021-01-08T22:04:12Z I! Starting Telegraf
Telegraf unknown (git: fork 50cd124)
/opt/main.sh: line 592:   439 Aborted                 (core dumped) /opt/td-agent-bit/bin/td-agent-bit -c /etc/opt/microsoft/docker-cimprov/td-agent-bit.conf -e /opt/td-agent-bit/bin/out_oms.so

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: just verified. if this is not specified td-agent crashing so we need to handle this case to avoid case in case if user provided one value but not other.

build/linux/installer/scripts/tomlparser-agent-config.rb
ganga1980: > > 1. It is the same documentation for tail plugin you shared earlier - https://fluentbit.io/documentation/0.13/input/tail.html
> >    Look under the table for the settings.
> > 2. In your code "if @fbitTailBufferMaxSize > 0 && @fbitTailBufferChunkSize == 0
> >    @fbitTailBufferChunkSize = @fbitTailBufferMaxSize"
> >    You are setting the chunk size to max size if it is not set. This is also in the same tail documentation.
> 
> > The doc only talking out the default value. do you know if buf_max_size will be buf_chunk_size if not proided? as I pointed out here, I think, if the value of the buf_chunk_size   is greater than buf_max_size  then the fbit will fail with an error https://github.com/fluent/fluent-bit/blob/master/plugins/in_tail/tail_config.c#L165. Let me do quick validation and see what's behavior.

>  just verified. fbit crashes for example if chunksize ="20"  and similarly crashes if we specify  both chunksize=20 and maxsize =1 but not met the criterion  maxsize > chunksize .

rashmichandrashekar (approved):

PR #492, opened 2021-01-19
upgrade apt to latest version

ganga1980:
upgrade to latest version of apt as per advisory https://lists.ubuntu.com/archives/ubuntu-security-announce/2020-December/005802.html to address reported vulnerability in apt version in the linux agent image.

### DIFF
--- a/kubernetes/linux/setup.sh
+++ b/kubernetes/linux/setup.sh
@@ -2,8 +2,8 @@ TMPDIR="/opt"
 cd $TMPDIR
 
 #Download utf-8 encoding capability on the omsagent container.
-
-apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
+#upgrade apt to latest version
+apt-get update && apt-get install -y --only-upgrade apt && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
 
 sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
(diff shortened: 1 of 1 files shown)


kubernetes/linux/setup.sh
```diff
@@ -2,8 +2,8 @@ TMPDIR="/opt"
 cd $TMPDIR
 
 #Download utf-8 encoding capability on the omsagent container.
-
-apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
+#upgrade apt to latest version
+apt-get update && apt-get install -y --only-upgrade apt && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
```
vishiy, line 6: why do we need --only-upgrade for apt and not for others ?

ganga1980: good question.  advisory about the apt version hence updated only this.  I tried invoking "apt-get upgrade"  at this line but that doesnt get update apt to latest. doing "apt-get upgrade" at the end upgrades all the pinned versions like td-agent-bit which we don't want.  I think,  going selective upgrade the packages what we need to get latest would be safer option than upgrading other packages.

vishiy: apt-get update && apt-get install apt ------ will pull the latest as well and also only for apt. based on doc, it says --only-upgrade will upgrade the package and its dependencies, and will ignore requests to install new packages. So if there is a new package introduced between the version we already have v. latest version we are trying to get to, it will cause some undesirable effects.

vishiy (approved):

PR #494, opened 2021-01-20
Use cloud-specific instrumentation keys

daweim0:
The environment variable APPLICATIONINSIGHTS_AUTH can now be set to a URL. The agent will download the content at the passed URL and interpret it as a base64 encoded instrumentation key (like the previous behavior)
(for example, https://ciinstrumentationdata.blob.core.windows.net/containerinsightstelem/ContainerInsightsInstrumentationData.txt)

This is my first agent PR, please let me know if I'm breaking any conventions.


kubernetes/linux/main.sh
```diff
@@ -1,5 +1,44 @@
 #!/bin/bash
 
+# check to see if Ikey is properly formatted an instrumentation key
```
vishiy, line 3: do we need this for public cloud ? if not, where is the cloud check ?

kubernetes/linux/main.sh
```diff
@@ -1,5 +1,44 @@
...
+if [[ $APPLICATIONINSIGHTS_AUTH =~ ^[A-Za-z0-9=]+$ ]]  # regex for base64 encoded data
```
rashmichandrashekar, line 4: Will the change for fetching the key from Storage account be in another PR?

kubernetes/linux/main.sh
```diff
@@ -1,5 +1,44 @@
...
+            echo "Using cloud-specific instrumentation key"
+      else
+            # no ikey can be retrieved. Put some value in the ikey and disable telemetry
+            export APPLICATIONINSIGHTS_AUTH=NzAwZGM5OGYtYTdhZC00NThkLWI5NWMtMjA3ZjM3NmM3YmRi
+            export DISABLE_TELEMETRY=true
+            echo "Could not get cloud-specific instrumentation key (network error?). Disabling telemetry"
+      fi
+fi
+
+echo "export APPLICATIONINSIGHTS_AUTH=$APPLICATIONINSIGHTS_AUTH" >> ~/.bashrc
```
rashmichandrashekar, line 34: Looks like you are setting env variable without decoding. I think you should decode first and then set env and telemetry variable.

kubernetes/windows/main.ps1
```diff
@@ -120,9 +120,56 @@ function Set-EnvironmentVariables {
...
+            # couldn't fetch the Ikey. Use the default key and turn telemetry off
+            # this key will not work in airgapped clouds, but still set an actual value so code expecting an IKey will not crash
+            # (like a base64 decode)
+            Write-Host "Could not get cloud-specific instrumentation key (network error?). Disabling telemetry"
+            $aikey = "NzAwZGM5OGYtYTdhZC00NThkLWI5NWMtMjA3ZjM3NmM3YmRi"
+            [System.Environment]::SetEnvironmentVariable("DISABLE_TELEMETRY", "True", "Process")
+            [System.Environment]::SetEnvironmentVariable("DISABLE_TELEMETRY", "True", "Machine")
+        }
+    }
+    [System.Environment]::SetEnvironmentVariable("APPLICATIONINSIGHTS_AUTH", $aikey, "Process")
```
rashmichandrashekar, line 166: Same here

kubernetes/linux/main.sh
```diff
@@ -1,5 +1,44 @@
...
 
+# check to see if Ikey is properly formatted an instrumentation key
+if [[ $APPLICATIONINSIGHTS_AUTH =~ ^[A-Za-z0-9=]+$ ]]  # regex for base64 encoded data
+then
+      # base64 encoded ikey was passed, decode ikey and write to environment variable
+      export APPLICATIONINSIGHTS_AUTH=$(echo $APPLICATIONINSIGHTS_AUTH )
+else
+      # need to fetch ikey from storage account
+      for BACKOFF in {0..5}; do
+            KEY=$(curl -sS $APPLICATIONINSIGHTS_AUTH )
```
ganga1980, line 11: Why cant  we introduce $APPLICATIONINSIGHTS_AUTH_URL or something easy to understand (IKEY vs URL)and invoke the logic to set APPLICATIONINSIGHTS_AUTH value fetched from SA account and this way you will be covered all the clouds and not touching any existing code? For example, if ($APPLICATIONINSIGHTS_AUTH_URL) { fetch "aikey"}?

ganga1980, line 11: In hybrid environments, customer can be configured to proxy in that case, this has to be invoked with proxy and if proxy set similar to other external endpoints we do here in this file. Though JEDI not supported yet for Azure Arc, but it will be good to cover similar to other external endpoints we do in the code.

kubernetes/linux/main.sh
```diff
@@ -1,5 +1,44 @@
...
+      done
+
+      # validate that the retrieved data is an instrumentation key
+      if [[ $KEY =~ ^[A-Za-z0-9=]+$ ]]; then
+            export APPLICATIONINSIGHTS_AUTH=$(echo $KEY)
+            echo "Using cloud-specific instrumentation key"
+      else
+            # no ikey can be retrieved. Put some value in the ikey and disable telemetry
+            export APPLICATIONINSIGHTS_AUTH=NzAwZGM5OGYtYTdhZC00NThkLWI5NWMtMjA3ZjM3NmM3YmRi
+            export DISABLE_TELEMETRY=true
```
ganga1980, line 29: I didnt see that, you are writing DISABLE_TELEMETRY value to ~/.bashrc. dont we need to write so that we can disable telemetry in case where we failed to get the IKEY?

kubernetes/windows/main.ps1
```diff
@@ -120,9 +120,56 @@ function Set-EnvironmentVariables {
     }
 
     # Set environment variable for TELEMETRY_APPLICATIONINSIGHTS_KEY
-    $aiKey = [System.Text.Encoding]::UTF8.GetString([System.Convert]::FromBase64String($env:APPLICATIONINSIGHTS_AUTH))
-    [System.Environment]::SetEnvironmentVariable("TELEMETRY_APPLICATIONINSIGHTS_KEY", $aiKey, "Process")
-    [System.Environment]::SetEnvironmentVariable("TELEMETRY_APPLICATIONINSIGHTS_KEY", $aiKey, "Machine")
+    # (or fetch a custom telemetry key if set)
+    $aiKeyPassed = [System.Environment]::GetEnvironmentVariable('APPLICATIONINSIGHTS_AUTH')
```
ganga1980, line 124: similarly, here, imo, going with APPLICATIONINSIGHTS_AUTH_URL would be cleaner and easy to understand rather using same variable name for both URL and AUTH.

kubernetes/windows/main.ps1
```diff
@@ -120,9 +120,56 @@ function Set-EnvironmentVariables {
...
+        # instrumentation key directly passed, set environment variable
+        $aikey = $aiKeyPassed;
+    }
+    else {
+        $aiKeyFetched = ""
+        
+        # retry up to 5 times
+        for( $i = 1; $i -le 5; $i++) {
+            try {
+                $response = Invoke-WebRequest -uri $aiKeyPassed -UseBasicParsing -TimeoutSec 5 -ErrorAction:Stop
```
ganga1980, line 135: -ErrorAction:stop  should stop the execution in case of any error and basically, container will be crashloop back state in case of any error. I presume, we want to proceed but without telemetry and in that case, we should remove "-ErrorAction:stop" proceed.

ganga1980 (changes requested):

kubernetes/linux/main.sh
```diff
@@ -1,5 +1,44 @@
...
+      # need to fetch ikey from storage account
+      for BACKOFF in {0..5}; do
+            KEY=$(curl -sS $APPLICATIONINSIGHTS_AUTH )
+
+            # there's no easy way to get the HTTP status code from curl, so just check if the result is well formatted
+            if [[ $KEY =~ ^[A-Za-z0-9=]+$ ]]; then
+                  break
+            else
+                  # (exponential backoff)
+                  sleep $((2**$BACKOFF / 4))
```
ganga1980, line 18: is **/4** intentional here? in windows side,  i see, you have different backoff intervals.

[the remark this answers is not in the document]
daweim0: It will block up to 15.5 seconds before moving on with no ikey (https://www.wolframalpha.com/input/?i=sum+2%5En%2F4+from+n%3D1+to+n%3D5). Is this too long or not long enough?

kubernetes/linux/main.sh
daweim0: The cloud check is implicit. If we're in public (or another non-airgapped cloud) then the ikey will be passed in $APPLICATIONINSIGHTS_AUTH and no web requests will be made. If we're in an air-gapped cloud then $APPLICATIONINSIGHTS_AUTH will be set to a URL and the agent will try to fetch an ikey from it.

kubernetes/linux/main.sh
daweim0: $APPLICATIONINSIGHTS_AUTH previously held the base64 encoded ikey, and $TELEMETRY_APPLICATIONINSIGHTS_KEY held the decoded ikey, this is matching previous behavior. (do we want to change that behavior?)

kubernetes/linux/main.sh
daweim0: oops, that was not intentional. Adding /4 to windows

kubernetes/linux/main.sh
daweim0: Good point, I'll fix that

kubernetes/linux/main.sh
daweim0: The IKey URL is now passed in through a new environment variable - APPLICATIONINSIGHTS_AUTH_URL. It will only be set in private clouds, so checking if it is set serves as a cloud check

kubernetes/linux/main.sh
daweim0: Good idea, I made this change. Now if there is a custom IKey then APPLICATIONINSIGHTS_AUTH_URL will be set with the URL

ganga1980: Discussed with Arc team and @joby-thomas was also there in the meeting. Eventually, they want to enable in these clouds but not in near term. probable  you can defer this and we can enable it when we can make this change when we add support.

kubernetes/linux/main.sh
```diff
@@ -161,6 +161,39 @@ fi
 export CLOUD_ENVIRONMENT=$CLOUD_ENVIRONMENT
 echo "export CLOUD_ENVIRONMENT=$CLOUD_ENVIRONMENT" >> ~/.bashrc
 
+# Check if the instrumentation key needs to be fetched from a storage account (as in arigapped clouds)
```
ganga1980, line 164: typo :)

kubernetes/linux/main.sh
daweim0: Cool, we'll defer it then

ganga1980 (dismissed):
lgtm

kubernetes/linux/main.sh
daweim0: oops, fixed

vishiy (approved):

Issue #391, opened 2020-07-03
[Feature Request] Ability to set PriorityClass for OMSAgent in AKS


davelosert:
I am not sure if this is the correct repository to open this issue, but:

**It would be great if it was possible to somehow configure a priority-class for the `omsagent` within an AKS-Cluster.**

Currently, that does not seem possible - or at least I could find no documentation whatsoever about it.

In our scenario, we want to give the cluster monitoring a somewhat higher priority than most of the other services - so that in case of an error, we won't be flying blind.

The only workaround I can think of is to use a `global default` class - however that is not really feasible as there might be other, unimportant pods without a `PriorityClass` within the cluster that might suddenly be ranked way higher than they should be.

Is such a feature possible?

daweim0:
@davelosert does setting omsagent.priority solve your problem? [docs here](https://github.com/Michael-Sinz/Docker-Provider/blob/ci_dev/charts/azuremonitor-containers/README.md#configuration)  (sorry for bringing a year-old issue back to life, I'm doing some cleanup)

We don't support changing the priority class in managed deployments (onboarding a cluster through the azure portal or cli). However, that runs omsagent pods with the system-node-critical priority class, which should be high enough?

davelosert:
@daweim0 : Thanks for your response and the tip.

However, we switched to another monitoring solution so unfortunately I can not try this out anymore.
Mabye there is someone else who is experiencing the same problem might be able to try and give you a feedback?

PR #500, opened 2021-02-09
Add priority class to the daemonsets

Michael-Sinz:
Add a priority class for omsagent and have the daemonsets use this
to be sure to schedule the pods.

Daemonset pods are constrained in scheduling to run on specific
nodes.  This is done by the daemonset controller.  When a node shows
up it will create a pod with a strong affinity to that node.  When a
node goes away, it will delete the pod with the node affinity to that
node.

Kubernetes pod scheduling does not know it is a daemonset but it does
know it is tied to a specific node.  With default scheduling, it is
possible for the pods to be "frozen out" of a node because the node
already is full.  This can happen because "normal" pods may already
exist and are looking for a node to get scheduled on when a node is
added to the cluster.  The daemonset controller will only first
create the pod for the node at around the same time.  The kubernetes
scheduler is running async from all of this and thus there can be a
race as to who gets scheduled on the node.

The pod priority class (and thus the pod priority) is a way to indicate
that the pod has a higher scheduling priority than a default pod.

By default, all pods are at priority 0.  Higher numbers are higher
priority.  Setting the priority to something greater than zero will
allow the omsagent daemonsets to win a race against "normal" pods for
scheduled resources on a node - and will also allow for graceful
eviction in the case the node is too full.

Without this, omsagent can be left out of node in clusters that are
very busy, especially in dynamic scaling situations.

I did not test the windows pod as we have no windows clusters.

Fixes issue #391

### DIFF
--- a/charts/azuremonitor-containers/templates/omsagent-priorityclass.yaml
+++ b/charts/azuremonitor-containers/templates/omsagent-priorityclass.yaml
@@ -1,0 +1,22 @@
+{{- if and (ne .Values.omsagent.secret.key "<your_workspace_key>") (ne .Values.omsagent.secret.wsid "<your_workspace_id>") (or (ne .Values.omsagent.env.clusterName "<your_cluster_name>") (ne .Values.omsagent.env.clusterId "<your_cluster_id>") (ne .Values.Azure.Cluster.ResourceId "<your_cluster_id>") )}}
+# This pod priority class is used for daemonsets to allow them to have priority
+# over pods that can be scheduled elsewhere.  Without a priority class, it is
+# possible for a node to fill up with pods before the daemonset pods get to be
+# created for the pod or get scheduled.  Note that pods are not "daemonset"
+# pods - they are just pods created by the daemonset controller but they have
+# a specific affinity set during creation to the specific node each pod was
+# created to run on (daemonset controller takes care of that)
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: omsagent
+  # Priority classes don't have labels :-)
+  annotations:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    component: oms-agent
+value: {{ .Values.omsagent.priority }}
+globalDefault: false
+description: "This is the daemonset priority class for omsagent"
+{{- end }}
--- a/charts/azuremonitor-containers/templates/omsagent-priorityclass.yaml
+++ b/charts/azuremonitor-containers/templates/omsagent-priorityclass.yaml
@@ -2,7 +2,7 @@
 # This pod priority class is used for daemonsets to allow them to have priority
 # over pods that can be scheduled elsewhere.  Without a priority class, it is
 # possible for a node to fill up with pods before the daemonset pods get to be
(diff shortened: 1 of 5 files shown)


Michael-Sinz:
I have actually not tested this in a significant way yet due to our deployment depending on having the chart in a helm server somewhere.  I am still working on an adjustment to the deployment process to allow me to reference my local changes.

mradmila (approved):
You should probably not include formatting changes and you should link the 391 issue.

ganga1980 (dismissed):
lgtm. After the PR merged, automatically deploys to our dev clusters (which has windows nodes ) so I can help on validating on windows nodes.

charts/azuremonitor-containers/values.yaml
```diff
@@ -20,6 +20,20 @@ omsagent:
...
+  # in getting scheduled to the node.  This needs to be greater than 0 such
+  # that the daemonset pods, which can not schedule onto different nodes as
+  # they are defined to run on specific nodes, are not accidentally frozen
+  # out of a node due to other pods showing up earlier in scheduling.
+  # (DaemonSet pods by definition only are created once the node exists for
+  # them to be created for and thus it is possible to have "normal" pods
+  # already in line to run on the node before the DeamonSet controller got a
+  # chance to build pod for the node and give it to the scheduler)
+  # Should be some number greater than default (0)
+  priority: 10
```
vishiy, line 35: Why is default 10 ? Can we have 0 (no priority) if not specified thru params ?

charts/azuremonitor-containers/templates/omsagent-priorityclass.yaml
```diff
@@ -0,0 +1,22 @@
...
+kind: PriorityClass
+metadata:
+  name: omsagent
+  # Priority classes don't have labels :-)
+  annotations:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    component: oms-agent
+value: {{ .Values.omsagent.priority }}
```
vishiy, line 19: check needed for ensuring .Values.omsagent.priority is a number ?

vishiy, line 19: Can you also please add this param to the .README.MD in the repo (relative path - charts/azuremonitor-containers/README.md)

charts/azuremonitor-containers/values.yaml
Michael-Sinz: The reason for something larger than 0 is that at 0 this will end up having pods for some nodes stay pending due to the race I discussed if the cluster is dynamically scaling due to pending work pods.  If k8s had thought of this from the start, daemonset controller would have defaulted to something like 10 to begin with such that daemonsets are not frozen out.

charts/azuremonitor-containers/templates/omsagent-priorityclass.yaml
Michael-Sinz: Will add something to the readme.  Good idea.

Michael-Sinz: Helm will do those checks as part of the validation against the schema when applying to a cluster.

Michael-Sinz:
@ganga1980 - needs another review as I updated the README.md to describe the priority value.

ganga1980 (approved):

vishiy (approved):

PR #501, opened 2021-02-11
[Not for ci_dev] Moving out KubePodInventory to a separate process

gracewehner:
Team, please take a look at the code changes that David and I have been working on. This is only for the replicaset, the daemonset will still use the prod image. The in_kube_podinventory plugin is separated out from omsagent into a fluentd process that's running the latest version of fluentd. KubePodinventory data is sent through oneagent whereas the other data types collected in in_kube_podinventory plugin are still sent through the out_mdm and out_oms plugins. Everything else is still using omsagent. Telemetry is collected for the cpu usage of fluentd, omsagent, and oneagent.


rashmichandrashekar:
I have a basic question. Why do we need to duplicate all the files for the new fluent? what is the fundamental difference?

daweim0:
> I have a basic question. Why do we need to duplicate all the files for the new fluent? what is the fundamental difference?

There are some slight differences (mostly adding Fluent:: to the beginning of names). Coppying all the files seemed cleanest, but I'm open to merging the two plugin folders.

ganga1980 (commented):
Hi, @daweim0 , @gracewehner, there are few bug fixes missing in this agent, are you planning to take those fixes? For example  cadvisor bug fix  https://github.com/microsoft/Docker-Provider/pull/504
https://github.com/microsoft/Docker-Provider/pull/505

daweim0:
Thanks, I'll pull those fixes in.

daweim0:
(converting to a draft since this is not meant to be merged)

PR #505, opened 2021-02-22
Gangams/feb 2021 agent bug fix

ganga1980:
This PR addresses the tasks (9221286 & 8870309)and which will go in Feb agent release

Here is the validation of Repo, Image, ImageTag and ImageId across the agent versions
image

### DIFF
--- a/source/plugins/ruby/kubernetes_container_inventory.rb
+++ b/source/plugins/ruby/kubernetes_container_inventory.rb
@@ -45,12 +45,10 @@ class KubernetesContainerInventory
             end
             # imagedId is of the format - repo@sha256:imageid
             imageIdValue = containerStatus["imageID"]
-            imageRepo = ""
             if !imageIdValue.nil? && !imageIdValue.empty?
               atLocation = imageIdValue.index("@")
               if !atLocation.nil?
                 containerInventoryRecord["ImageId"] = imageIdValue[(atLocation + 1)..-1]
-                imageRepo = imageIdValue[0..(atLocation - 1)]
               end
             end            
             containerInventoryRecord["ExitCode"] = 0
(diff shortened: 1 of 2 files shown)


source/plugins/ruby/kubernetes_container_inventory.rb
```diff
@@ -107,6 +84,51 @@ def getContainerInventoryRecords(podItem, batchTime, clusterCollectEnvironmentVa
...
+                  containerInventoryRecord["Image"] = imageValue
+                else
+                  # repo/image
+                  containerInventoryRecord["Repository"] = imageValue[0..(slashLocation - 1)]
+                  containerInventoryRecord["Image"] = imageValue[(slashLocation + 1)..-1]
+                end 
+                # if no tag specified, k8s assumes latest as imagetag and this is same behavior from docker API and from status.
+                # Ref - https://kubernetes.io/docs/concepts/containers/images/#image-names
+                if isDigestSpecified == false 
+                  containerInventoryRecord["ImageTag"] = "latest"
```
vishiy, line 127: Can we remove this? If its not specified, we don't have image tag, rather then filling in with 'latest'.  k8s might assume latest, but we are showing customers what they specified.

source/plugins/ruby/kubernetes_container_inventory.rb
```diff
@@ -165,6 +187,7 @@ def getContainersInfoMap(podItem, isWindows)
             podContainers.each do |container|
               containerInfoMap = {}
               containerName = container["name"]
+              containerInfoMap["image"] = container["image"]
```
vishiy, line 190: why is this added as part of this change ?

source/plugins/ruby/kubernetes_container_inventory.rb
ganga1980: > Can we remove this? If its not specified, we don't have image tag, rather then filling in with 'latest'. k8s might assume latest, but we are showing customers what they specified.

This is same behavior we have in previous agents (docker & kubelet APIs). removing this would break the customer scenarios if they have any queries or alerts on "latest" image tag.

source/plugins/ruby/kubernetes_container_inventory.rb
ganga1980: > why is this added as part of this change ?

> This is getting image from "spec" .

source/plugins/ruby/kubernetes_container_inventory.rb
vishiy: i dont remember seeing hardcoding image to 'latest' anywhere before this change. Where is it hardcoded before to break compatibility ?

vishiy: are you saying previousy we were getting from runtime and it was filling in with latest ? and now we are getting from spec ?

ganga1980: Yes, "latest" was coming from the status and same behavior we had with Docker API as well.
For example
# 1. create nginx pod
 kubectl run nginxpod --image=nginx
# 2. observe "latest" tag under status
  containerStatuses:
  - containerID: containerd://93762c31560ecdd8994351d0884b30b8a7d534956c62994b7b44ac38f4266dd1
    image: docker.io/library/nginx:latest

vishiy (approved):

PR #508, opened 2021-02-24
The metric kubelet_running_pods_count was renamed in 1.19

gracewehner:
Fix for the kubelet workbook issue found by Ganga where kubelet_running_pods_count metrics were missing (ICM: https://portal.microsofticm.com/imp/v3/incidents/details/229404590/home)

The metric `kubelet_running_pods_count` was renamed to` kubelet_running_pods` starting with kubernetes 1.19 (https://github.com/kubernetes/kubernetes/pull/92407). We will need to scrape both names to support all kubernetes versions.

### DIFF
--- a/build/linux/installer/conf/telegraf.conf
+++ b/build/linux/installer/conf/telegraf.conf
@@ -675,7 +675,9 @@
   ## An array of urls to scrape metrics from.
   urls = ["$CADVISOR_METRICS_URL"]
 
-  fieldpass = ["kubelet_running_pod_count","volume_manager_total_volumes", "kubelet_node_config_error", "process_resident_memory_bytes", "process_cpu_seconds_total"]
+  # <= 1.18: metric name is kubelet_running_pod_count
+  # >= 1.19: metric name changed to kubelet_running_pods
+  fieldpass = ["kubelet_running_pod_count","kubelet_running_pods","volume_manager_total_volumes", "kubelet_node_config_error", "process_resident_memory_bytes", "process_cpu_seconds_total"]
 
   metric_version = 2
   url_tag = "scrapeUrl"


build/linux/installer/conf/telegraf.conf
```diff
@@ -675,7 +675,9 @@
   ## An array of urls to scrape metrics from.
   urls = ["$CADVISOR_METRICS_URL"]
 
-  fieldpass = ["kubelet_running_pod_count","volume_manager_total_volumes", "kubelet_node_config_error", "process_resident_memory_bytes", "process_cpu_seconds_total"]
+  # <= 1.18: metric name is kubelet_running_pod_count
+  # >= 1.19: metric name changed to kubelet_running_pods
+  fieldpass = ["kubelet_running_pod_count","kubelet_running_pods","volume_manager_total_volumes", "kubelet_node_config_error", "process_resident_memory_bytes", "process_cpu_seconds_total"]
```
vishiy, line 680: Isn't the new ones 	'running_pods' & 'running_containers' ?

vishiy, line 680: If its either/or (meaning clusters >= 1.19.x will have the new metric, and clusters < 1.19.x will have old metric), we should fix the workbook query to include both the metrics so account for all the cluster versions.

gracewehner: Yes, I have created a task for this on the board

gracewehner: Could you point me to where you see new names as 'running_pods' & 'running_containers? I have tried with 'kubelet_running_pods' and see the data in LA

vishiy: here - https://github.com/kubernetes/kubernetes/blob/a7fe0bf8e593decabba030f771ea8dbb34d22710/pkg/kubelet/metrics/metrics.go#L56

Metric is named without 'kubelet' but looks its prefixing with subsystem (kubelet in this case). If you see the metric flowing, we are good.

vishiy (approved):

PR #511, opened 2021-03-09
Gangams/arc proxy contract and token renewal updates

ganga1980:
This PR has following updates ..
1.   Accepting ProxySettings and contract from Arc K8s RP which are being set by the customer as part of the Arc K8s cluster. This makes user doesnt need to specify for each Arc K8s extension
2.  Fix for token renewal issue and handling of re-creating crd  in case user accidently deleted

### DIFF
--- a/scripts/onboarding/templates/arc-k8s-extension/existingClusterOnboarding.json
+++ b/scripts/onboarding/templates/arc-k8s-extension/existingClusterOnboarding.json
@@ -13,14 +13,7 @@
       "metadata": {
         "description": "Location of the Azure Arc Connected Cluster Resource e.g. \"eastus\""
       }
-    },
-    "proxyEndpointUrl": {
-      "type": "string",
-      "defaultValue": "",
-      "metadata": {
-        "description": "If the cluster behind forward proxy, then specify Proxy Endpoint URL in this format: http(s)://<user>:<password>@<proxyhost>:<port>"
-      }
-    },
+    },  
     "workspaceResourceId": {
       "type": "string",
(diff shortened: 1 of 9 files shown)


scripts/onboarding/templates/arc-k8s-extension/existingClusterOnboarding.json
```diff
@@ -13,14 +13,7 @@
...
         "description": "Location of the Azure Arc Connected Cluster Resource e.g. \"eastus\""
       }
-    },
-    "proxyEndpointUrl": {
-      "type": "string",
-      "defaultValue": "",
-      "metadata": {
-        "description": "If the cluster behind forward proxy, then specify Proxy Endpoint URL in this format: http(s)://<user>:<password>@<proxyhost>:<port>"
-      }
-    },
```
gracewehner, line 23: Just want to confirm these settings in existingClusterOnboarding.json and existingClusterParam.json can be removed because if the proxy settings are needed, they _have_ to be settings for Arc K8s RP?

ganga1980: > Just want to confirm these settings in existingClusterOnboarding.json and existingClusterParam.json can be removed because if the proxy settings are needed, they _have_ to be settings for Arc K8s RP?

> Yes, @gracewehner,  proxy settings are being populated by Arc K8s RP as part of our chart deployment. Before it was passed by user. Since this being passed by RP so this parameter not required in the template.

gracewehner (approved):

Issue #409, opened 2020-07-29
enable-monitoring.sh fails if az output format not set to json


jamurtag:
The bash script to [enable monitoring for Arc clusters](https://docs.microsoft.com/en-us/azure/azure-monitor/insights/container-insights-enable-arc-enabled-clusters#enable-using-bash-script) fails if the output format is not json.

An example when output format is set to `table`:

```bash
$ bash enable-monitoring.sh --resource-id $azureArcClusterResourceId --kube-context $kubeContext --workspace-id $logAnalyticsWorkspaceResourceId
...
validating cluster identity
cluster identity type: result -------------- systemassigned
-e only supported cluster identity is systemassigned for Azure ARC K8s cluster type
```

The az command formatting only works for single line output.

In addition, a script does not satisfy our automation workflows, it would be useful to see RM/Terraform code to configure the workspace then instructions to implement the Helm chart natively.

ganga1980:
Hi, @jamurtag,  We are working on moving to extension model to enable monitoring for Arc K8s cluster(s) and with extension model, there is simple integrated CLI command and also can be enabled with RM & Policy etc. This script will be removed at that time. We are targetting extension model to release in couple of weeks.

nyuen:
@ganga1980 Should we close the issue following the merge #513 ?

PR #513, opened 2021-03-15
Adding explicit json output to az commands #409

nyuen:
Cluster onboarding fails if the AZ cli is not configured as json output. 
Adding explicit -o json parameter to az cli where no output format is specified #409

### DIFF
--- a/scripts/onboarding/managed/disable-monitoring.sh
+++ b/scripts/onboarding/managed/disable-monitoring.sh
@@ -127,7 +127,7 @@ remove_monitoring_tags()
 
   # validate cluster identity for Azure Arc enabled Kubernetes cluster
   if [ "$isArcK8sCluster" = true ] ; then
-   identitytype=$(az resource show -g ${clusterResourceGroup} -n ${clusterName} --resource-type $resourceProvider --query identity.type)
+   identitytype=$(az resource show -g ${clusterResourceGroup} -n ${clusterName} --resource-type $resourceProvider --query identity.type -o json)
    identitytype=$(echo $identitytype | tr "[:upper:]" "[:lower:]" | tr -d '"')
    echo "cluster identity type:" $identitytype
     if [[ "$identitytype" != "systemassigned" ]]; then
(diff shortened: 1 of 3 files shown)


scripts/onboarding/managed/disable-monitoring.sh
```diff
@@ -127,7 +127,7 @@ remove_monitoring_tags()
 
   # validate cluster identity for Azure Arc enabled Kubernetes cluster
   if [ "$isArcK8sCluster" = true ] ; then
-   identitytype=$(az resource show -g ${clusterResourceGroup} -n ${clusterName} --resource-type $resourceProvider --query identity.type)
+   identitytype=$(az resource show -g ${clusterResourceGroup} -n ${clusterName} --resource-type $resourceProvider --query identity.type -o json)
```
ganga1980, line 130: Thanks @nyuen  for making PR on this.  looks like some weird issue with az cli handling quotes.
Can you please sharing the version of the Az CLI and platform you are able to repro this issue? and also, just wanted to let you know that, we will be extension in few extensions and at that point of time, script will not be used any more.

nyuen: Tested with when the default output is configured to **table** as opposed to **json** :
Azure-cli    Azure-cli-core    Azure-cli-telemetry
-----------  ----------------  ---------------------
2.17.1       2.17.1            1.0.6

I've also tested it with the latest version:
Azure-cli    Azure-cli-core    Azure-cli-telemetry
-----------  ----------------  ---------------------
2.20.0       2.20.0            1.0.6


This is the error first error I get in both cases:
cluster identity type: result **-------------- systemassigned**
-e only supported cluster identity is **systemassigned** for Azure Arc enabled Kubernetes cluster type

With az output set to table you get:
az resource show -g ${rgName} -n ${clusterName} --resource-type $resourceProvider --query identity.type
**Result
--------------
SystemAssigned**

with json output you get
az resource show -g ${rgName} -n ${clusterName} --resource-type $resourceProvider --query identity.type -o json
**"SystemAssigned"**

We need to explicitly set the output to json or tsv on every single calls in case the user has a different configuration:
az resource show -g ${rgName} -n ${clusterName} --resource-type $resourceProvider --query identity.type -o tsv
**SystemAssigned**

@ganga1980 thanks for the update regarding the script, makes sense to use the embedded CLI :-)

ganga1980: Thanks for this PR and validated, looks good to me.

ganga1980 (approved):
lgtm

PR #515, opened 2021-03-17
Prometheus scraping from sidecar and OSM changes

rashmichandrashekar:
Changelist-
1. Move custom prometheus to sidecar when environment variable is set.
2. Provide support for OSM metric scraping
3. Removed wireserver calls to get certs
4. Addition of sidecar container
5. Providing Label and Field selector options for custom prometheus scraping
6. Latest telegraf build

PR #517, opened 2021-03-19
added liveness and telemetry for telegraf

daweim0:
Liveness probe will now fail if telegraf is not running. Additionally, a message will be printed to the agent's stdout (which will be picked up in the traces table).

I don't think this will interact with the [OSM & Prometheus Scraping PR](https://github.com/microsoft/Docker-Provider/pull/515), but I'll double check after the OSM PR is merged.

This PR was inspired by [this ICM](https://portal.microsofticm.com/imp/v3/incidents/details/231222858/home), telegraf was dying in their replicaset but the agent didn't restart so metrics stopped flowing.


build/linux/installer/scripts/livenessprobe.sh
```diff
@@ -26,6 +26,15 @@ then
  exit 1
 fi
 
+#test to exit non zero value if telegraf is not running
+(ps -ef | grep telegraf | grep -v "grep")
+if [ $? -ne 0 ]
+then
+ echo "Telegraf is not running" > /dev/termination-log
+ echo "Telegraf is not running" > /dev/write-to-traces  # this file is tailed and printed to stdout
```
vishiy, line 34: If you write to stderr for our container, shouldn't we already be picking that up ?

daweim0: Liveness probe output is not written to container stdout :(

vishiy: ok makes sense. Can you also include the changes for tailing this file ?

daweim0: Tailing the file is included, see main.sh line 630

vishiy: so this will go to our stdout and we will pick it up, which looks good. But only thing is if cusotmer turns on kube-system log, it will also end up in customer's workspace. Rather if you just tail this file thru fluentbit, and route to our telemetry it will be quiet for customers. But either way is fine.

vishiy (approved):

rashmichandrashekar:
Please make this change for windows as well.

rashmichandrashekar (approved):

build/windows/installer/scripts/livenessprobe.cmd
```diff
@@ -7,6 +7,13 @@ IF ERRORLEVEL 1 (
     exit /b 1
 )
 
+tasklist /fi "imagename eq telegraf.exe" /fo "table"  | findstr telegraf
+
+IF ERRORLEVEL 1 (
+    echo "Telegraf is not running"
```
gracewehner, line 13: Will this also be sent to our traces telemetry, the same as for linux?

rashmichandrashekar:
> Please make this change for windows as well.

As we discussed offline, this can be added later. Else, it will restart windows omsagent constantly until the sidecar changes rollout. You can remove it for now.

build/linux/installer/scripts/livenessprobe.sh
daweim0: done, the liveness probe failure message is no longer sent to the container stdout

build/windows/installer/scripts/livenessprobe.cmd
daweim0: Got rid of the windows componenet, right now only linux. Rashmi and I decided to add telegraf to the windows liveness probe later.

gracewehner (approved):

vishiy (approved):

PR #526, opened 2021-04-05
Fix conflicting logrotate settings

gracewehner:
The node and the omsagent container both have a cron.daily file to rotate certain logs daily. These settings are the same for some files in /var/log (mounted from the node with read/write access), causing the rotation to fail when both try to rotate at the same time. So then the /var/log/*.1 file is written to forever. Since these files are always written to and never rotated, it causes high memory usage on the node after a while.

This fix removes the container logrotate settings for /var/log, which the container does not write to.


kubernetes/linux/setup.sh
```diff
@@ -85,3 +85,7 @@ rm -f $TMPDIR/docker-cimprov*.sh
 rm -f $TMPDIR/azure-mdsd*.deb
 rm -f $TMPDIR/mdsd.xml
 rm -f $TMPDIR/envmdsd
+
+# Remove settings for cron.daily that conflict with the node's cron.daily. Since both are trying to rotate the same files
+# in /var/log at the same time, the rotation doesn't happen correctly and then the *.1 file is forever logged to.
+rm /etc/logrotate.d/alternatives /etc/logrotate.d/apt /etc/logrotate.d/azure-mdsd /etc/logrotate.d/rsyslog
```
vishiy, line 91: Grace, please create a workitem in the april release, to make /var/log readonly for a future fix.

vishiy, line 91: please add an empty line in the end as this is a shell script.

vishiy (approved):

PR #530, opened 2021-04-07
Windows metric fix

daweim0:
Fixing the metrics cpuUsagePercentage and memoryWorkingSetPercentage. I added telemetry to catch percentage metrics greater than 100% or less than 0%, let me know if we should keep it or take it out.


source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -9,7 +9,6 @@ class Kube_nodeInventory_Input < Input
     @@MDMKubeNodeInventoryTag = "mdm.kubenodeinventory"
     @@configMapMountPath = "/etc/config/settings/log-data-collection-settings"
     @@promConfigMountPath = "/etc/config/settings/prometheus-data-collection-settings"
-    @@osmConfigMountPath = "/etc/config/osm-settings/osm-metric-collection-configuration"
```
rashmichandrashekar, line 12: please get the latest ci_dev changes

source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -157,6 +155,19 @@ def parse_and_emit_records(nodeInventory, batchTime = Time.utc.iso8601)
         insightsMetricsEventStream = MultiEventStream.new
         kubePerfEventStream = MultiEventStream.new
         @@istestvar = ENV["ISTEST"]
+        
+        begin
+          # first refresh the node cpu and memory capacity cache
```
ganga1980, line 160: This will be invoked for every min and its not first refresh, right?

source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -496,4 +501,98 @@ def getNodeTelemetryProps(item)
...
+          # this is always run even though there was a return earlier
+          @lock.unlock
+        end
+      end
+  
+      def set_capacity_all(capacity_json)
+        # clean the cache here so that calling code doesn't have to remember to clean it. The disadvantage to this 
+        # approach is that the cache will never be cleaned if new nodes are never added, (but then the cache also 
+        # can't grow and our main concern is unbounded cache growth)
+        if DateTime.now.to_time.to_i - @lastCacheClearTime > @@RECORD_TIME_TO_LIVE
```
ganga1980, line 538: You are invoking set_capacity_all for every minute with list of nodes available atm,. in that case, just taking that snapshot override the existing cache with new list isnt that good enough? why we need to maintain cache for non-existent nodes?

source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -157,6 +155,19 @@ def parse_and_emit_records(nodeInventory, batchTime = Time.utc.iso8601)
         insightsMetricsEventStream = MultiEventStream.new
         kubePerfEventStream = MultiEventStream.new
         @@istestvar = ENV["ISTEST"]
+        
+        begin
+          # first refresh the node cpu and memory capacity cache
+          cpu_capacity_json = KubernetesApiClient.parseNodeLimits(nodeInventory, "capacity", "cpu", "cpuCapacityNanoCores")
+          @NodeCache.cpu.set_capacity_all(cpu_capacity_json)
```
ganga1980, line 162: so the cache for all the nodes and not just windows nodes, is that right? if yes, why do need for Linux nodes?

source/plugins/ruby/filter_cadvisor2mdm.rb
```diff
@@ -161,19 +163,30 @@ def filter(tag, time, record)
             if counter_name == Constants::CPU_USAGE_NANO_CORES
               metric_name = Constants::CPU_USAGE_MILLI_CORES
               metric_value /= 1000000 #cadvisor record is in nanocores. Convert to mc
-              @log.info "Metric_value: #{metric_value} CPU Capacity #{@cpu_capacity}"
-              if @cpu_capacity != 0.0
-                percentage_metric_value = (metric_value) * 100 / @cpu_capacity
+              target_node_cpu_capacity_nc = @NodeCache.cpu.get_capacity(record["DataItems"][0]["Host"])
```
ganga1980, line 166: we are invoking getNodesResourceUri in ensure_cpu_memory_capacity_set when controller_type.downcase == "replicaset". is this additional call not required any more since we will be using from cache?

source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -496,4 +501,98 @@ def getNodeTelemetryProps(item)
       return properties
     end
   end # Kube_Node_Input
+
+
+  class NodeStatsCache
+    # inner class for caching implementation (CPU and memory caching is handled the exact same way, so logic to do so is moved to a private inner class)
```
ganga1980, line 507: Agree, we should move this logic to separate file.

source/plugins/ruby/in_kube_nodes.rb
daweim0: Correct, it will be invoked every minute (we don't know when new nodes will be added, so I think it needs to be refreshed every minute)

source/plugins/ruby/in_kube_nodes.rb
daweim0: oops, done

ganga1980 (commented):
lgtm

ganga1980 (approved):

vishiy (approved):

PR #531, opened 2021-04-08
Adding MDM metrics for threshold violation

rashmichandrashekar:
Adding 4 new metrics for threshold violation so that nulls can be treated as zeroes and alerts can go back resolved state.

### DIFF
--- a/source/plugins/ruby/MdmMetricsGenerator.rb
+++ b/source/plugins/ruby/MdmMetricsGenerator.rb
@@ -169,43 +169,59 @@ class MdmMetricsGenerator
 
         metric_threshold_hash = getContainerResourceUtilizationThresholds
         container_zero_fill_dims = [Constants::OMSAGENT_ZERO_FILL, Constants::OMSAGENT_ZERO_FILL, Constants::OMSAGENT_ZERO_FILL, Constants::KUBESYSTEM_NAMESPACE_ZERO_FILL].join("~~")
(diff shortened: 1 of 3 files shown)


source/plugins/ruby/MdmMetricsGenerator.rb
```diff
@@ -276,14 +307,27 @@ def getContainerResourceUtilMetricRecords(recordTimeStamp, metricName, percentag
...
+
+        # Adding another metric for threshold violation
+        resourceThresholdViolatedRecord = MdmAlertTemplates::Container_resource_threshold_violation_template % {
+          timestamp: recordTimeStamp,
+          metricName: @@container_metric_name_metric_threshold_violated_hash[metricName],
+          containerNameDimValue: containerName,
+          podNameDimValue: podName,
+          controllerNameDimValue: controllerName,
+          namespaceDimValue: podNamespace,
+          containerResourceThresholdViolated: isZeroFill ? 0 : 1,
```
vishiy, line 319: Do these metrics get manufactured only when threshold is violated? Thats why you are zero filling them to generate 'violated' metric ?

vishiy (approved):

source/plugins/ruby/MdmMetricsGenerator.rb
rashmichandrashekar: Yes. That is right

PR #532, opened 2021-04-12
Collector and ME telemetry for error logs, # metrics scraped

gracewehner:
- Install and run fluentbit. Re-uses the telemetry functions of out_oms.go plugin
- Log only collector errors, tail and send to traces telemetry table
- Still log ME at info level
  - Tail and filter by error/fatal logs and send to traces telemetry table
  - Tail and filter by ME log line that has 1) count of metrics processed that minute, 2) count of metrics sent to publication that minute, and 3) account name. Send to customMetrics telemetry table


otelcollector/fluent-bit/fluent-bit.conf
```diff
@@ -0,0 +1,54 @@
...
+    Flush         15
+    HTTP_Server   Off
+    Daemon        Off
+    storage.path  /var/opt/microsoft/docker-cimprov/state/flbstore/
+    storage.sync  normal
+    storage.checksum off
+    storage.backlog.mem_limit 10M
+    Log_Level     info
+    Parsers_File  /opt/fluent-bit/fluent-bit-parsers.conf
+    Log_File      /var/opt/microsoft/docker-cimprov/log/fluent-bit.log
```
vishiy, line 11: please change it to /opt/fluentbit

otelcollector/fluent-bit/src/Makefile
```diff
@@ -0,0 +1,22 @@
...
+	#export BUILDDATE=$(CONTAINER_BUILDVERSION_DATE)
+	#@echo $(BUILDVERSION)
+	#@echo $(BUILD_DATE)
+	@echo "========================= cleanup existing .so and .h file  ========================="
+	rm -rf *.so *.h *~
+	@echo "========================= go get  ========================="
+	go get
+	@echo "========================= go build  ========================="
+	#go build -ldflags "-X 'main.revision=$(BUILDVERSION)' -X 'main.builddate=$(BUILDDATE)'" -buildmode=c-shared -o out_oms.so .
+	go build -buildmode=c-shared -o out_oms.so .
```
vishiy, line 16: change out_oms to out_appinsights ? :)

otelcollector/fluent-bit/src/out_oms.h
```diff
@@ -0,0 +1,83 @@
+/* Code generated by cmd/cgo; DO NOT EDIT. */
```
vishiy, line 1: these header files are generated. not needed to be source controlled.

otelcollector/fluent-bit/src/Makefile
```diff
@@ -0,0 +1,22 @@
+BASE_DIR := $(subst /build/linux,,$(PWD))
```
vishiy, line 1: does this make file gets picked up & built when we do make from otelcollector folder ?

otelcollector/scripts/setup.sh
```diff
@@ -51,15 +51,25 @@ cp -f $TMPDIR/prometheus-2.25.2.linux-amd64/promtool /opt/promtool
...
+wget -qO - https://packages.fluentbit.io/fluentbit.key | sudo apt-key add -
+sudo echo "deb https://packages.fluentbit.io/ubuntu/xenial xenial main" >> /etc/apt/sources.list
+sudo echo "deb http://security.ubuntu.com/ubuntu bionic-security main" >> /etc/apt/sources.list.d/bionic.list
+sudo apt-get update
+sudo apt-get install td-agent-bit=1.6.8 -y
+
 sudo apt --fix-broken install -y
 sudo apt-get install inotify-tools -y
 
+# Some dependencies were fixed with sudo apt --fix-broken, try installing td-agent-bit again
```
vishiy, line 69: do you know why is this ? Can you install fluent bit just once after -fix-broken, and if you do that the issue goes away ?

otelcollector/fluent-bit/src/Makefile
gracewehner: I have added this in now

otelcollector/scripts/setup.sh
gracewehner: Yes this fixes it. I believe this is because we upgraded to ubuntu 20, but are keeping the same fluent-bit version and one dependency needed is a lower version than what comes with ubuntu 20

vishiy (approved):

PR #538, opened 2021-04-20
Rashmi/april agent 2021

rashmichandrashekar:
1. Making Job completion metric configurable
2. Telemetry for omsagent requests and limits


source/plugins/ruby/MdmMetricsGenerator.rb
```diff
@@ -490,6 +508,12 @@ def getContainerResourceUtilizationThresholds
           pvUsagePercentageThresholdFloat = (pvUsagePercentageThreshold.to_f).round(2)
           metric_threshold_hash[Constants::PV_USED_BYTES] = pvUsagePercentageThresholdFloat
         end
+
+        jobCompletionTimeThreshold = ENV["AZMON_ALERT_JOB_COMPLETION_TIME_THRESHOLD"]
```
vishiy, line 512: So Rashmi, i see you have made the threshold configurable, and also have kept the previous default value (6 hours), so we don't create any breaking changes, correct ?

build/linux/installer/scripts/tomlparser-mdm-metrics-config.rb
```diff
@@ -101,6 +102,25 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+    begin
+      jobCompletion = parsedConfig[:alertable_metrics_configuration_settings][:job_completion_threshold]
+      if !jobCompletion.nil?
+        jobCompletionThreshold = jobCompletion[:job_completion_threshold_time_minutes]
+        jobCompletionThresholdInt = jobCompletionThreshold.to_i
+        if jobCompletionThresholdInt.kind_of? Integer
+          @jobCompletionThresholdMinutes = jobCompletionThresholdInt
+        else
+          puts "config::Non interger value or value not convertible to integer specified for job completion threshold, using default "
+          @jobCompletionThresholdMinutes = Constants::DEFAULT_MDM_JOB_COMPLETED_TIME_THRESHOLD_MINUTES
```
vishiy, line 116: Add this to our telemetry as well, so we know what people configure this as ?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -456,6 +461,33 @@ def getContainerResourceRequestsAndLimits(pod, metricCategory, metricNameToColle
               metricProps["Collections"].push(metricCollections)
               metricItem["DataItems"].push(metricProps)
               metricItems.push(metricItem)
+              #Telemetry about omsagent requests and limits
```
vishiy, line 464: Thanks for adding this. This works and also you are not adding additional calls. For future reference, you ca easily pass this to us thru downward API. Please see here for reference - https://github.com/microsoft/Docker-Provider/blob/08a6c6ae1c0bf31b3bb10b07b9b2b0042a9978a5/otelcollector/deploy/prometheus-collector.yaml#L101

build/linux/installer/scripts/tomlparser-mdm-metrics-config.rb
rashmichandrashekar: good point.. will add it.

source/plugins/ruby/MdmMetricsGenerator.rb
rashmichandrashekar: yes thats right. the default will still be 6 hours.

build/linux/installer/scripts/tomlparser-mdm-metrics-config.rb
rashmichandrashekar: Double checked this. I had already added this when I first made the change. So it is taken care of.

vishiy (approved):

PR #539, opened 2021-04-20
add Read_from_Head config for all fluentbit tail plugins

gracewehner:
See the commit message of: https://github.com/fluent/fluent-bit/commit/70e33fa2618227882d48faf690848bb4117cdb3e
for details explaining the fluentbit change and what Read_from_Head does when set to true.

Setting Read_from_Head = true fixes the issues where:
- only the last lines of the omsagent startup logs are sent to traces
- when fluentbit sees new files (new container logs) that don't have an offset in the database, it starts tailing the file from the end and the beginning logs are missed

I tested these scenarios:
- full omsagent logs are now sent to traces (whether it's logs from starting up, restarting, shutting down, or just running)
- full logs are now sent to ContainerLogs for a new pod/container that's created after omsagent startup
- known log files are still tailed from the offset in the database after omsagent restart (they aren't affected by the Read_from_Head setting)
- /dev/write-to-traces for telegraf liveness still works as expected

### DIFF
--- a/build/linux/installer/conf/td-agent-bit-prom-side-car.conf
+++ b/build/linux/installer/conf/td-agent-bit-prom-side-car.conf
@@ -15,6 +15,7 @@
     Name tail
     Tag oms.container.log.flbplugin.terminationlog.*
     Path /dev/write-to-traces
+    Read_from_Head true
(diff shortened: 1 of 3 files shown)


vishiy:
Thanks for the test cases.

vishiy (approved):

PR #541, opened 2021-04-21
Update sidecar mem limits 

rashmichandrashekar:
1. Update side car memory limits
2. Fix job olderThanHours to precision 2


build/linux/installer/conf/td-agent-bit-prom-side-car.conf
```diff
@@ -31,7 +31,7 @@
     Port        25229
     Chunk_Size  1m
     Buffer_Size 1m
-    Mem_Buf_Limit 20m
+    Mem_Buf_Limit 200m
```
ganga1980, line 34: what's the rationale behind increasing this limit? It will be really helpful if you can provide more context in the PR description.

rashmichandrashekar: @ganga1980 - This was a change made as per our discussion in scrum for the prometheus scraping incident - 
https://portal.microsofticm.com/imp/v3/incidents/details/231222858/home

ganga1980 (approved):

PR #553, opened 2021-04-28
Update enable-monitoring.sh

AshWilliams:
The version 2.8.3 doesn't exist and the whole script fails at the end. Check the following tag list
https://mcr.microsoft.com/v2/azuremonitor/containerinsights/preview/azuremonitor-containers/tags/list

### DIFF
--- a/scripts/onboarding/managed/enable-monitoring.sh
+++ b/scripts/onboarding/managed/enable-monitoring.sh
@@ -44,7 +44,7 @@ defaultAzureCloud="AzureCloud"
 omsAgentDomainName="opinsights.azure.com"
 
 # released chart version in mcr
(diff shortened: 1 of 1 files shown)


vishiy:
Thank you for the PR. Its not released yet. So please continue to use 2.8.2 locally until 2.8.3 gets pushed. we need to fix this process. Apologies for inconvenience. We will post back here when 2.8.3 gets pushed out.

gracewehner:
@AshWilliams the chart with version 2.8.3 has been pushed now, so the script should now work again.

ganga1980:
Hi, @AshWilliams , we have removed the version dependency and next release will address this.

PR #560, opened 2021-05-18
Add wait time for telegraf and also force mdm egress to use tls 1.2

vishiy:


rashmichandrashekar (dismissed):

source/plugins/ruby/out_mdm.rb
```diff
@@ -102,6 +102,7 @@ def start
           else
             @http_client = Net::HTTP.new(@post_request_uri.host, @post_request_uri.port)
           end
+          @http_client.instance_eval { @ssl_context = OpenSSL::SSL::SSLContext.new(:TLSv1_2) }
```
ganga1980, line 105: This (meaning negotiation) can fail if the metric endpoint enforces higher version of the TLS at some point, for example TLSv1_3.  I think, you have mentioned there are no server side logs,  but curios, you were able to get clues from client, what version of TLS version being used the client? probably, picking the higher supported TLS version can make both the client & server agree lower version so that we dont need to worry about  TLS enforcement at least for available TLS versions.

ganga1980, line 105: I think, you have mentioned  ciphers being picked up randomly. is the cipher not issue?

ganga1980, line 105: dont we need to build the ruby with openssl to use OpenSSL? the reason I ask is to make sure the changes I am working on work with your updates.

vishiy: These are the TLS versions supported by ingestion. Its literally tlsv1_2 alone. so what you do want to make agree by whom ?
image.

vishiy: ciphers are picked on based on tls protocol. If you see the error, in our case client is picking ssl_v3 and server doesn;t support, and we see the error on the client side.

vishiy: ruby is already built with later version of openssl. If mdm is ingestion working in your new bits, then tlsv1_2 is also working. Is that not the case ? Why do we want to build with a different version of openssl ?

rashmichandrashekar (approved):

source/plugins/ruby/out_mdm.rb
ganga1980: what I am saying, if ingestion endpoint makes TLS1.3 as supported version, then this code will break because the server wants to TLS 1.3 where as client hard coded for TLS1.2.

vishiy: yes, thats eactly what i said in the scrum yesterday. There is no negotiation is happening. we are forcing it to use v12

ganga1980: OK, mdm ingestion works with new bits. IIRC, i remembered you were mentioning about this in scrum so asking whether we would need anything in ruby & openssl combo.

ganga1980: negotiation still happens but client saying it can only support TLS12 or lower.  https://datatracker.ietf.org/doc/html/rfc5246#appendix-E.1
https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.1.2 
If the server upgrades to new version and then connection closed . To be future proof, if we set TLS1.3 as client version, then it should work future & all existing scenarios.

vishiy: currently tls 1.3 is not supported by ingestion. see pic above.

vishiy: Also, how do we know they will go to tls13 and not sslv3?

ganga1980: because tls13 will be new version and they cant go back to deprecated sslv3 version.

vishiy: you are predicting the future. they can go to a totally different ssl version like 14 when it becomes available.

PR #568, opened 2021-05-25
Send perf metrics to MDM from windows daemonset

rashmichandrashekar:
Send perf metrics to MDM from windows daemonset


vishiy (approved):

build/windows/installer/conf/fluent.conf
```diff
@@ -46,6 +61,23 @@
...
 </filter>
 
+<match mdm.cadvisorperf**>
+  @type mdm
+  @log_level debug
+  <buffer>
+    @type file
+    path /etc/omsagentwindows/out_mdm_cdvisorperf.buffer
+    overflow_action drop_oldest_chunk
+    chunk_limit_size 4m
```
ganga1980, line 71: I think, we should add queue_limit_length 20 similar to all other plugins. This will help us to buffer upto 80MB in case of transient failures.

rashmichandrashekar: I have reused the configuration we have in Linux. Is there any reason why this was not in Linux and should be included in Windows?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -21,10 +21,15 @@ class ApplicationInsightsUtility
   @@EnvApplicationInsightsEndpoint = "APPLICATIONINSIGHTS_ENDPOINT"
   @@EnvControllerType = "CONTROLLER_TYPE"
   @@EnvContainerRuntime = "CONTAINER_RUNTIME"
-
+  @@isWindows = false
+  @@hostName = (OMS::Common.get_hostname)
+  @@os_type = ENV["OS_TYPE"]
+  if !@@os_type.nil? && !@@os_type.empty? && @@os_type.strip.casecmp("windows") == 0
+    @@isWindows = true
+    @@hostName = ENV["HOSTNAME"]
```
ganga1980, line 29: Looks like this doesnt give windows host name, right? since AI telemetry code running on rs node?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -21,10 +21,15 @@ class ApplicationInsightsUtility
   @@EnvApplicationInsightsEndpoint = "APPLICATIONINSIGHTS_ENDPOINT"
   @@EnvControllerType = "CONTROLLER_TYPE"
   @@EnvContainerRuntime = "CONTAINER_RUNTIME"
-
+  @@isWindows = false
+  @@hostName = (OMS::Common.get_hostname)
```
ganga1980, line 25: I think, we should avoid running this code - OMS::Common.get_hostname on windows since this linux specific implementation.

rashmichandrashekar: We set in the yaml for windows - https://github.com/microsoft/Docker-Provider/blob/959b455d5ab873b6fa5ed7445bd61dc847ec2c08/kubernetes/omsagent.yaml#L779

source/plugins/ruby/ApplicationInsightsUtility.rb
rashmichandrashekar: This is set in the yaml - https://github.com/microsoft/Docker-Provider/blob/959b455d5ab873b6fa5ed7445bd61dc847ec2c08/kubernetes/omsagent.yaml#L779

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -133,16 +138,23 @@ def initializeUtility()
...
-        # DockerVersion field holds either containerRuntime for non-docker or Dockerversion if its docker
-        @@CustomProperties["DockerVersion"] = containerRuntime
-        if containerRuntime.casecmp("docker") == 0
-          dockerInfo = DockerApiClient.dockerInfo
-          if (!dockerInfo.nil? && !dockerInfo.empty?)
-            @@CustomProperties["DockerVersion"] = dockerInfo["Version"]
+      begin
+        # Not doing this for windows since docker is being deprecated soon and we dont want to bring in the socket dependency.
+        if !@@isWindows.nil? && @@isWindows == false
+          containerRuntime = ENV[@@EnvContainerRuntime]
```
ganga1980, line 144: CONTAINER_RUNTIME available on windows. only thing we need to avoid talking docker sock as its specific to linux and it will be deprecated soon with containerd.

source/plugins/ruby/ApplicationInsightsUtility.rb
rashmichandrashekar: > I think, we should avoid running this code - OMS::Common.get_hostname on windows since this linux specific implementation.

This works in windows as well.

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -25,7 +25,12 @@ class KubernetesApiClient
   #@@IsValidRunningNode = nil
   #@@IsLinuxCluster = nil
   @@KubeSystemNamespace = "kube-system"
-  @LogPath = "/var/opt/microsoft/docker-cimprov/log/kubernetes_client_log.txt"
+  @os_type = ENV["OS_TYPE"]
+  if !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp("windows") == 0
```
ganga1980, line 29: wrapping this to helper function is good option to consider to avoid same code multiple places.

rashmichandrashekar: Since this is across multiple files, I think wrapper function would be an overkill. Agree this would have helped if it was in the same file.

source/plugins/ruby/ApplicationInsightsUtility.rb
rashmichandrashekar: I will fix this.

source/plugins/ruby/ApplicationInsightsUtility.rb
ganga1980: How does the get_hostname method which has dependency "/var/opt/microsoft/docker-cimprov/state/containerhostname" path would work on windows? are you saying this returns just empty?  if its working then why need to read from environment variable?

source/plugins/ruby/KubernetesApiClient.rb
ganga1980: helper function would be useful when that piece of code being used multiple places regardless whether thats same file or different file. it makes more readble and easy to manage as well.

source/plugins/ruby/ApplicationInsightsUtility.rb
rashmichandrashekar: It returns the container host name, which is the pod name which is incorrect for our telemetry. It does not return empty.

source/plugins/ruby/KubernetesApiClient.rb
rashmichandrashekar: I think a better way to solve this problem is to keep it consistent between linux and windows. This is a workaround since we dont have the same path. When we moved to oneagent, I think one of the optimizations we can make is to use the same path.

PR #571, opened 2021-05-28
trying again to fix the json gem

daweim0:
Removing the manual json gem update

Tuns out we already had the fix to CVE-2020-10663, the installed patch level is 142
https://github.com/ruby/ruby/commit/36e9ed7fef6eb2d14becf6c52452e4ab16e4bf01

### DIFF
--- a/kubernetes/linux/setup.sh
+++ b/kubernetes/linux/setup.sh
@@ -50,7 +50,7 @@ sudo apt-add-repository ppa:brightbox/ruby-ng -y
 sudo apt-get update
 sudo apt-get install ruby2.6 ruby2.6-dev gcc make -y
 # to fix CVE-2020-10663
-gem update json -v 2.5.1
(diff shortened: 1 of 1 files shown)


kubernetes/linux/setup.sh
```diff
@@ -50,7 +50,7 @@ sudo apt-add-repository ppa:brightbox/ruby-ng -y
 sudo apt-get update
 sudo apt-get install ruby2.6 ruby2.6-dev gcc make -y
 # to fix CVE-2020-10663
-gem update json -v 2.5.1
+gem install json -v 2.5.1
```
ganga1980, line 53: For CVE, we should upgrade to 2.5.1. I think, this makes both 2.1.0 & 2.5.1 co-exist which doesnt address cve.

ganga1980 (approved):

PR #573, opened 2021-06-04
Make sidecar fluentbit chunk size configurable

rashmichandrashekar:


build/linux/installer/conf/td-agent-bit-prom-side-car.conf
```diff
@@ -29,9 +29,9 @@
     Tag         oms.container.perf.telegraf.*
     Listen      0.0.0.0
     Port        25229
-    Chunk_Size  10m
-    Buffer_Size 10m
-    Mem_Buf_Limit 200m
+    Chunk_Size  ${AZMON_SIDECAR_FBIT_CHUNK_SIZE}
```
vishiy, line 32: Rashmi, Can we just test for the highest scale and hard code for that config , rather than making this configurable ? Any side effects due to that ?

rashmichandrashekar: Discussed offline with Vishwa. I will document that the recommended max size for chunk size and buffer size would be 4M as discussed with the ODS team. And they can increase the mem buf limit for higher scales, if they increase the limits for sidecar using helm chart. Making this configurable as requested by a customer through an icm.

vishiy (approved):

PR #582, opened 2021-06-15
Windows build optimization

rashmichandrashekar:


kubernetes/windows/Dockerfile-dev-base-image
```diff
@@ -0,0 +1,43 @@
+FROM mcr.microsoft.com/windows/servercore:ltsc2019
+MAINTAINER OMSContainers@microsoft.com
+LABEL vendor=Microsoft\ Corp \
+    com.microsoft.product="Azure Monitor for containers"
+
+# Do not split this into multiple RUN!
+# Docker creates a layer for every RUN-Statement
+RUN powershell -Command "Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))"
+# Fluentd depends on cool.io whose fat gem is only available for Ruby < 2.5, so need to specify --platform ruby when install Ruby > 2.5 and install msys2 to get dev tools
+RUN choco install -y ruby --version 2.6.5.1 --params "'/InstallDir:C:\ruby26'" \
```
ganga1980, line 10: imo, this can go out of sync, when we upgrade ruby and fluentd versions, any dependencies. are you planning to apply these changes to main one so that we can use the same dependencies for dev test and release?

rashmichandrashekar: Once this is used by people across the team, we can make further optimizations to this. 
But since the dependencies are not changes often, we would need to make changes in both places until then.

vishiy (approved):

PR #584, opened 2021-06-15
feat: add new metrics to MDM for allocatable % calculation of cpu and memory usage

bragi92:


source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -81,6 +81,11 @@ def getSummaryStatsFromCAdvisor(winNode)
       return getResponse(winNode, relativeUri)
     end
 
+    def getCongifzCAdvisor(winNode)
+      relativeUri = "/configz"
```
rashmichandrashekar, line 85: Does this endpoint work in windows?

bragi92: yes. I've tested that scenario. Windows daemonset also sends the proper values.

bragi92: sent a DM with my cluster running the images from this branch

rashmichandrashekar: Cool.. thanks!

bragi92: I'm resolving this comment since I've answered it and we need to resolve all comments before merging

ganga1980, line 85: do we have any restriction support of this endpoint on  k8s version? or will this api work on k8s version 1.17 or up?

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -41,6 +41,108 @@ def get_node_capacity
...
+          end
+          @log.info "get_node_allocatable::kubereserved_cpu  #{kubereserved_cpu}"
+        rescue => errorStr
+          @log.error "Error in get_node_allocatable::kubereserved_cpu: #{errorStr}"
+          kubereserved_cpu = "0"
+          ApplicationInsightsUtility.sendExceptionTelemetry("Error in get_node_allocatable::kubereserved_cpu: #{errorStr}")
+        end 
+
+        begin
+          kubereserved_memory = JSON.parse(allocatable_response.body)["kubeletconfig"]["kubeReserved"]["memory"]
```
ganga1980, line 81: instead of multiple JSON.parse, optimizing to single JSON parse would help on the perf side.

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -41,6 +41,108 @@ def get_node_capacity
...
+            kubereserved_memory = "0"
+          end
+          @log.info "get_node_allocatable::kubereserved_memory #{kubereserved_memory}"
+        rescue => errorStr
+          @log.error "Error in get_node_allocatable::kubereserved_memory: #{errorStr}"
+          kubereserved_memory = "0"
+          ApplicationInsightsUtility.sendExceptionTelemetry("Error in get_node_allocatable::kubereserved_cpu: #{errorStr}")
+        end 
+        begin
+          systemReserved_cpu = JSON.parse(allocatable_response.body)["kubeletconfig"]["systemReserved"]["cpu"]
```
ganga1980, line 92: similarly here JSON.parse. doing one time JSON.parse and reusing the parsed object would help on perf side.

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -41,6 +41,108 @@ def get_node_capacity
...
+          end
+          @log.info "get_node_allocatable::systemReserved_cpu  #{systemReserved_cpu}"
+        rescue => errorStr
+          # this will likely always reach this condition for AKS ~ change logic....
+          @log.error "Error in get_node_allocatable::systemReserved_cpu: #{errorStr}"
+          systemReserved_cpu = "0"
+          ApplicationInsightsUtility.sendExceptionTelemetry("Error in get_node_allocatable::kubereserved_cpu: #{errorStr}")
+        end 
+        begin
+           systemReserved_memory = JSON.parse(allocatable_response.body)["kubeletconfig"]["systemReserved"]["memory"]
```
ganga1980, line 104: same as above.

source/plugins/ruby/filter_cadvisor2mdm.rb
```diff
@@ -12,6 +12,7 @@ module Fluent::Plugin
   require_relative "kubelet_utils"
   require_relative "MdmMetricsGenerator"
   require_relative "in_kube_nodes"
+  require_relative "KubernetesApiClient"
```
ganga1980, line 15: looks like this is not required, right?

source/plugins/ruby/filter_cadvisor2mdm.rb
```diff
@@ -345,6 +379,7 @@ def ensure_cpu_memory_capacity_set
         end
       elsif @@controller_type.downcase == "daemonset"
         capacity_from_kubelet = KubeletUtils.get_node_capacity
+        allocatable_from_kubelet = KubeletUtils.get_node_allocatable(@@isWindows)
```
ganga1980, line 382: two questions 1) should  be winNode instead @@isWindows here 2) i see that, you are passing windows to the allocatable but not capacity. this will be an issue on windows nodes, right?

source/plugins/ruby/filter_cadvisor2mdm.rb
bragi92: removed

source/plugins/ruby/kubelet_utils.rb
bragi92: updated

source/plugins/ruby/kubelet_utils.rb
bragi92: updated

source/plugins/ruby/kubelet_utils.rb
bragi92: updated

source/plugins/ruby/filter_cadvisor2mdm.rb
bragi92: I checked the windows container. I see that we actually have a node ip environment variable defined for windows containers too and the get_node_capacity function is also using it. ([linked here](https://github.com/microsoft/Docker-Provider/blob/ci_dev/source/plugins/ruby/CAdvisorMetricsAPIClient.rb#L106))

I'm updating my code to use this too now similar to capacity.

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -41,6 +41,111 @@ def get_node_capacity
...
+        cpu_allocatable = cpu_allocatable / (1000.0 ** 2)
+        @log.info "CPU Allocatable #{cpu_allocatable}"
+
+        memory_allocatable = memory_capacity - (KubernetesApiClient.getMetricNumericValue("memory", kubereserved_memory) + KubernetesApiClient.getMetricNumericValue("memory", systemReserved_memory) + KubernetesApiClient.getMetricNumericValue("memory", evictionHard_memory))
+        @log.info "Memory Allocatable #{memory_allocatable}"
+
+        # cpu_allocatable = BigDecimal(cpu_allocatable.to_s).to_f
+        # memory_allocatable = BigDecimal(memory_allocatable.to_s).to_f
+
+        return [cpu_allocatable, memory_allocatable]
```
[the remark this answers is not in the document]
vishiy: Yeah, this will be an issue in hybrid as people can customize the node cpu in a way they want. Currently there is no way in managed kubernetes to set aside reserved cpus. But it could be part of future node customization that we need to be aware..
Since MDM metrics doesn;t apply to un-managed now, this seems ok?

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -41,6 +41,105 @@ def get_node_capacity
...
+          @log.error "Error in get_node_allocatable::evictionHard_memory: #{errorStr}"
+          evictionHard_memory = "0"
+          ApplicationInsightsUtility.sendExceptionTelemetry("Error in get_node_allocatable::kubereserved_cpu: #{errorStr}")
+        end 
+
+        # do calculation in nanocore since that's what KubernetesApiClient.getMetricNumericValue expects
+        cpu_capacity_number = cpu_capacity.to_i * 1000.0 ** 2
+        # subtract to get allocatable. Formula : Allocatable = Capacity - ( kube reserved + system reserved + eviction threshold )
+        # https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable
+        cpu_allocatable  = cpu_capacity_number - (KubernetesApiClient.getMetricNumericValue("cpu", kubereserved_cpu) + KubernetesApiClient.getMetricNumericValue("cpu", systemReserved_cpu))
```
vishiy, line 128: kaveesh, it works for all units (like Mi, Gi, cpu=2, 100mc etc..) ?

vishiy (dismissed):

source/plugins/ruby/MdmMetricsGenerator.rb
```diff
@@ -37,6 +37,12 @@ class MdmMetricsGenerator
     Constants::MEMORY_WORKING_SET_BYTES => Constants::MDM_NODE_MEMORY_WORKING_SET_PERCENTAGE,
   }
 
+  @@node_metric_name_metric_allocatable_percentage_name_hash = {
+    Constants::CPU_USAGE_MILLI_CORES => Constants::MDM_NODE_CPU_USAGE_ALLOCATABLE_PERCENTAGE,
+    Constants::MEMORY_RSS_BYTES => Constants::MDM_NODE_MEMORY_RSS_ALLOCATABLE_PERCENTAGE,
```
vishiy, line 42: you also need to make a change on metrics rp for hardcoding the metric metadata (units & allowed aggs etc...) as first step before you push this change out.

source/plugins/ruby/kubelet_utils.rb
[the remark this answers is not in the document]
ganga1980: We do support MDM metrics for hybrid via azure arc. If the reserved-cpus takes precedence, is there any reason not considering that?

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
rashmichandrashekar: For secure cadvisor, will this require adding configz resource for omsagent-reasder resource?

source/plugins/ruby/kubelet_utils.rb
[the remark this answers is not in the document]
bragi92: I see, updating the code to take this into account now that I know its being used in azure arc.

source/plugins/ruby/MdmMetricsGenerator.rb
bragi92: ok, will ping you again once I have that change ready

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -41,6 +41,123 @@ def get_node_capacity
...
+          end
+          @log.info "get_node_allocatable::systemReserved_cpu  #{systemReserved_cpu}"
+        rescue => errorStr
+          # this will likely always reach this condition for AKS ~ only applicable for hyrid + MDM combination
+          @log.error "Error in get_node_allocatable::systemReserved_cpu: #{errorStr}"
+          systemReserved_cpu = "0"
+          ApplicationInsightsUtility.sendExceptionTelemetry("Error in get_node_allocatable::kubereserved_cpu: #{errorStr}")
+        end 
+
+        begin
```
bragi92, line 101: find better way of doing this ~ right now it logs too much telemetry.

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
bragi92: tested it in a cluster where isSecure is set to true and it works. resolving conversation. Please reopen it if needed.

source/plugins/ruby/MdmMetricsGenerator.rb
bragi92: metrics rp changes have been made and are rolling out. resolving converstaion.

source/plugins/ruby/kubelet_utils.rb
bragi92: actually this is fine, its only happening once when the container starts up and that's exactly the number of times I want to log this.

vishiy (approved):

PR #588, opened 2021-06-21
Exception stack multiline

gracewehner:
I will wait for AAD changes to merge and release before merging this.

This is the exception stack multiline feature using Fluent-Bit [DockerMode](https://docs.fluentbit.io/manual/pipeline/inputs/tail#docker_mode) and [Multiline](https://docs.fluentbit.io/manual/pipeline/inputs/tail#multiline) settings with a regex which looks for timestamps at the start of log lines and appends lines without that regex to the starting multiline one.
It's enabled by default when using ContainerLogV2 schema, with a hidden configmap setting to disable.

For windows, it uses fluent-bit for container logs only when containerlogv2 and multiline are enabled. The config files for fluent-bit and fluentd are changed to handle which is collecting the container logs.

Also, fluent-bit filepath storage isn't supported in windows fluent-bit until 1.7. I upgraded to 1.7.8 and did the same for linux to keep parity. The docs did not mention any breaking changes and everything looked ok to me, but more testing may be needed for the newer fluent-bit version before release.


build/windows/installer/conf/fluent-bit.conf
```diff
@@ -2,15 +2,15 @@
...
     Log_Level    info
-    Log_File      /etc/fluent-bit/fluent-bit.log
+    Log_File     /etc/fluent-bit/fluent-bit.log
 
 [INPUT]
-    Name            forward
-    Listen          127.0.0.1
-    Port            25230
-    Mem_Buf_Limit   10m
-    Chunk_Size      32
```
gracewehner, line 12: The naming and units for these are different and were causing errors for the newer fluent-bit version: https://docs.fluentbit.io/manual/pipeline/inputs/forward

kubernetes/linux/main.sh
```diff
@@ -38,6 +38,68 @@ waitforlisteneronTCPport() {
       fi
 }
 
+#Setting environment variable for CAdvisor metrics to use port 10255/10250 based on curl request
```
gracewehner, line 41: This is necessary to move earlier in main.sh to know the container runtime to change the fluent-bit conf file before starting fluent-bit

build/common/installer/scripts/tomlparser.rb
```diff
@@ -242,12 +261,20 @@ def get_command_windows(env_variable_name, env_variable_value)
   file = File.open("setenv.ps1", "w")
 
   if !file.nil?
+
+    # This path format is necessary for fluent-bit in windows
+    if @logStitchMultiline
```
ganga1980, line 266: I think, this will override this  overrides  @logTailPath = "/opt/nolog*.log" which being set when the customer disabled both stderr & stdout logs.  please double check on that.

gracewehner: Thanks Ganga, great catch, I will fix this

build/common/installer/scripts/tomlparser.rb
```diff
@@ -21,6 +21,7 @@
 @collectClusterEnvVariables = true
 @logTailPath = "/var/log/containers/*.log"
 @logExclusionRegexPattern = "(^((?!stdout|stderr).)*$)"
+@logStitchMultiline = false
```
ganga1980, line 24: Hi, @gracewehner , I am planning to add the automation script to deploy all the workloads and scenarios yamls to the scale cluster so that we can use them to scale test and also potentially for any dev/test cluster to validate, can you please add the test yamls related to multiline logging so that we can use them to test on regular basis?

build/windows/installer/conf/azm-containers-parser.conf
```diff
@@ -0,0 +1,35 @@
...
+    Regex ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<log>.*)$
+    Time_Key    time
+    Time_Format %Y-%m-%dT%H:%M:%S.%L
+    Time_Keep    On
+
+
+[PARSER]
+    Name containerd-multiline
+    Format regex
+    Regex ^(?<time>[^ ]+) (?<stream>stdout|stderr) (?<logtag>[^ ]*) (?<log> *(\\u001b\[\d{1,2}m)* *(\[.*\])? *(((Mon|Tue|Wed|Thu|Fri|Sat|Sun|M|T|W|R|F|S|U|Monday|Tuesday|Wednesday|Thursday|Friday|Saturday|Sunday) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{1,2} \d{1,2})|(\d{4}-\d{2}-\d{2})|(\d{2}:\d{2}:\d{2}((\.|,)\d{1,10})?)|(\d{8}T\d{6}((\.|,)\d{1,10})?)|(\d{10,17})|(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{0,9}Z)|(((\d{1,2} +(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) +\d{4,9})|((Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) +\d{1,2},? +\d{4,9})) +((\d{1,2}:\d{1,2}:\d{1,2}((\.|,)\d{1,10})?)))).*)$
```
ganga1980, line 26: is it possible to add the rubular linke like this CRI parser one  # http://rubular.com/r/tjUt3Awgg4 for these complex regexes?

source/plugins/go/src/oms.go
```diff
@@ -1049,6 +1053,25 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+
+		if (stripMultilineHeaders) {
+			logEntry = multilineContainerdRegex.ReplaceAllString(logEntry, "\n")
+
+			// If there's a single line log when multiline is enabled, it won't be parsed by the parser
+			// and the whole line with the containerd header will be the log entry.
+			// Parse the timestamp, source, and log entry, so the log isn't dropped in the LA pipeline.
+			if len(logEntryTimeStamp) == 0 {
+				var containerdHeaderRegex = regexp.MustCompile(`^([^ ]+) (stdout|stderr) ([^ ]*) ( *(.*))$`)
+				groupMatches := containerdHeaderRegex.FindStringSubmatch(logEntry)
```
ganga1980, line 1067: FindStringSubmatch can return "nil"- https://golang.org/src/regexp/regexp.go. From the reading the golang docs, len of ni slice is 0 - https://tour.golang.org/moretypes/12. I am wondering, we should add "nil" check to make sure we dont crash if the implementation of this function changes with go version.  This is not critical but i am thinking as safetynet.

build/windows/installer/conf/fluent-multiline.conf
```diff
@@ -0,0 +1,92 @@
...
+</source>
+
+<source>
+  @type cadvisor_perf
+  tag oms.api.cadvisorperf
+  run_interval 60
+  @log_level debug
+</source>
+
+<source>
```
ganga1980, line 14: if I understand correctly, with multi-line, we will be using Fluent bit tail plugin not the Fluent-D,right? in that case, why do we need FluentD tail plugin?

PR #589, opened 2021-06-22
Gangams/remove chart version dependency

ganga1980:
Removed chart version dependency in onboarding scripts  which will address issues of pulling pre-released chart version. Chart repo used same as one we have here - https://docs.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-hybrid-setup#install-the-helm-chart

### DIFF
--- a/scripts/onboarding/managed/enable-monitoring.ps1
+++ b/scripts/onboarding/managed/enable-monitoring.ps1
@@ -62,11 +62,10 @@ $isArcK8sCluster = $false
 $isAksCluster = $false
 $isUsingServicePrincipal = $false
 
-# released chart version in mcr
-$mcr = "mcr.microsoft.com"
-$mcrChartVersion = "2.8.3"
-$mcrChartRepoPath = "azuremonitor/containerinsights/preview/azuremonitor-containers"
-$helmLocalRepoName = "."
+# microsoft helm chart repo
+$microsoftHelmRepo="https://microsoft.github.io/charts/repo"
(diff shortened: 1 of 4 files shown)


scripts/onboarding/managed/enable-monitoring.ps1
```diff
@@ -62,11 +62,10 @@ $isArcK8sCluster = $false
 $isAksCluster = $false
 $isUsingServicePrincipal = $false
 
-# released chart version in mcr
-$mcr = "mcr.microsoft.com"
-$mcrChartVersion = "2.8.3"
-$mcrChartRepoPath = "azuremonitor/containerinsights/preview/azuremonitor-containers"
-$helmLocalRepoName = "."
+# microsoft helm chart repo
+$microsoftHelmRepo="https://microsoft.github.io/charts/repo"
```
gracewehner, line 66: Are we already pushing the helm chart here in our release process? Will we not push the helm chart to the mcr.microsoft.com anymore then?

ganga1980: Yes, the same repo - https://microsoft.github.io/charts/repo used for on-prem k8s which we already have the release process. I have added note in the release process doc to indicate the same repo being used for AROv4 as well.  Yes, we dont use mcr.microsoft.com or our ACR as chart repo either for On-prem K8s or AROv4.

gracewehner (approved):

PR #598, opened 2021-07-02
Linux build speedup

daweim0:
Similar to the windows build speed improvements, package fetching can now be cached. Multiple `apt-get update` calls were also removed

I'm taking two strategies here:
1. install all dependencies before installing the docker-provider bundle. That way the dependency installation steps can be cached (much like Rashmi's Windows build speedup). This requires moving some things out of setup.sh
2. Run apt-get update as few times as practically possible. There's no reason to do it more than twice (once to install wget so that other package repositories can be added, then once again to get packages from those repositories). We previously ran apt-get update 5 times)
3. Do things in parallel if possible. I think the only easy parallelization we can do is installing telegraf while doing apt operations  (since telegraf is not installed through apt)


kubernetes/linux/setup.sh
```diff
@@ -1,9 +1,27 @@
...
+    wget https://dl.influxdata.com/telegraf/releases/telegraf-1.18.0_linux_amd64.tar.gz
+    tar -zxvf telegraf-1.18.0_linux_amd64.tar.gz
+    mv /opt/telegraf-1.18.0/usr/bin/telegraf /opt/telegraf
+    chmod 777 /opt/telegraf
+) &
 
 #Download utf-8 encoding capability on the omsagent container.
 #upgrade apt to latest version
-apt-get update && apt-get install -y apt && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
+apt-get install -y apt && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
```
vishiy, line 24: how is this related to speedier builds ? also, we need to update the already existing apt package manager. Will install on top existing update to latest version as well ?

kubernetes/linux/setup.sh
```diff
@@ -12,12 +30,11 @@ sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
 #install oneagent - Official bits (05/17/2021)
 wget https://github.com/microsoft/Docker-Provider/releases/download/05172021-oneagent/azure-mdsd_1.10.1-build.master.213_x86_64.deb
 
-/usr/bin/dpkg -i $TMPDIR/azure-mdsd*.deb
-cp -f $TMPDIR/mdsd.xml /etc/mdsd.d
-cp -f $TMPDIR/envmdsd /etc/mdsd.d
+/usr/bin/dpkg -i /opt/azure-mdsd*.deb
```
vishiy, line 33: why change this ?

kubernetes/linux/setup.sh
```diff
@@ -1,9 +1,27 @@
...
+sudo apt-add-repository ppa:brightbox/ruby-ng -y
+# set up apt repo for fluent-bit(td-agent-bit)
+wget -qO - https://packages.fluentbit.io/fluentbit.key | sudo apt-key add -
+sudo echo "deb https://packages.fluentbit.io/ubuntu/xenial xenial main" >> /etc/apt/sources.list
+
+sudo apt-get update
+
+apt-get install -y libc-bin wget openssl curl sudo python-ctypes init-system-helpers  net-tools rsyslog cron vim dmidecode apt-transport-https gnupg
+
+# install telegraf in parallel with everything else to save build time
```
vishiy, line 14: do you know how much time this saves ?

kubernetes/linux/setup.sh
```diff
@@ -70,3 +70,5 @@ if [  -e "/var/lib/gems/2.6.0/gems/http_parser.rb-0.6.0/Gemfile.lock" ]; then
       echo "Renaming unused gemfile.lock for http_parser 0.6.0"
       mv /var/lib/gems/2.6.0/gems/http_parser.rb-0.6.0/Gemfile.lock /var/lib/gems/2.6.0/gems/http_parser.rb-0.6.0/renamed_Gemfile_lock.renamed
 fi
+
+wait  # in case telegraf hasn't finished installing yet
```
vishiy, line 74: this wait call syncs with telegraf install ? if there is more installed in parallel , this will wait on everything (all children) i assume ?

kubernetes/linux/setup.sh
daweim0: I have no idea why we install apt. I didn't change that (the change on this line is removing `apt-get update`

kubernetes/linux/setup.sh
daweim0: Yes it will wait on all children. Telegraf is the only thing being installed in parallel (since everything else is installed through apt)

kubernetes/linux/setup.sh
daweim0: good point, I changed that back

kubernetes/linux/setup.sh
daweim0: Quite a bit after the initial build. 

current ci_dev fresh build:
- docker build took 406.7s
- pushing docker image required uploading 595MB + 31MB + 60.5kB

current ci_dev rebuild after one line golang change:
- docker build took 398.8s
- pushing docker image required uploading 595MB + 31.5MB + 60.5kB

this PR fresh build:
- docker build took 373.4s
- pushing docker image required uploading 645MB + 33.4kB + 31.5MB + 49.6MB

this PR rebuild after one line golang change:
- docker build took 5.7s
- pushing docker image required uploading 31.5MB + 49.6MB

(make and the bundler still run on every build, so builds are longer than my posted times. I didn't include them because they aren't changed by this PR)

PR #599, opened 2021-07-08
Don't crash on incomplete node data

daweim0:
Adds some protection against the Kubernetes API server returning incomplete node data. Normally this won't happen, but it can if users manually update node objects.

The diff is hard to read, but this basically adds a begin & rescue around some telemetry code.

### DIFF
--- a/source/plugins/ruby/in_kube_nodes.rb
+++ b/source/plugins/ruby/in_kube_nodes.rb
@@ -280,48 +280,54 @@ module Fluent::Plugin
           timeDifference = (DateTime.now.to_time.to_i - @@nodeTelemetryTimeTracker).abs
           timeDifferenceInMinutes = timeDifference / 60
           if (timeDifferenceInMinutes >= Constants::TELEMETRY_FLUSH_INTERVAL_IN_MINUTES)
-            properties = getNodeTelemetryProps(item)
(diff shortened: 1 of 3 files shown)


ganga1980 (approved):

source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -280,48 +280,54 @@ def parse_and_emit_records(nodeInventory, batchTime = Time.utc.iso8601)
...
+                properties["rsPromMonPodsLabelSelectorLength"] = @@rsPromMonitorPodsLabelSelectorLength
+                properties["rsPromMonPodsFieldSelectorLength"] = @@rsPromMonitorPodsFieldSelectorLength
               end
+              # telemetry about osm metric settings for replicaset
+              if (File.file?(@@osmConfigMountPath))
+                properties["osmNamespaceCount"] = @@osmNamespaceCount
+              end
+              ApplicationInsightsUtility.sendMetricTelemetry("NodeCoreCapacity", capacityInfo["cpu"], properties)
+              telemetrySent = true
             rescue => errorStr
```
vishiy, line 326: did you test what happens if any code in the block got rescued and rest  (partial) of the  code (after rescue) runs after the rescue ? Also whats the advantage of this rescue block ? This isn't a crash, correct? There are global rescues in outer methods (enumerate & runperiodic).

PR #606, opened 2021-07-15
Update container-azm-ms-agentconfig.yaml

tsubasaxZZZ:
Add tail_mem_buf_limit_megabytes describe


kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -144,7 +144,9 @@ data:
       tcp_listener_chunk_size = 10
       tcp_listener_buffer_size = 10
       tcp_listener_mem_buf_limit = 200
-
+    # tail buffer setting for large amounts of container stdout.
```
vishiy, line 147: hi, we dont want to set this as default, for everyone using the configmap. please add as an example here and comment it out, so people who wants it can uncomment and use it. we also need to update our docs for this and other fbit setting we have.

PR #614, opened 2021-07-23
remove un-used output plugin

vishiy:


rashmichandrashekar (approved):

build/linux/installer/conf/telegraf-rs.conf
```diff
@@ -124,26 +124,6 @@
   namedrop = ["agent_telemetry", "file"]
   #tagdrop = ["AgentVersion","AKS_RESOURCE_ID", "ACS_RESOURCE_NAME", "Region","ClusterName","ClusterType", "Computer", "ControllerType"]
 
-[[outputs.application_insights]]
```
ganga1980, line 127: curios, are these unused plugins causing the cpu spike?

vishiy: No. These are generating logs and reported by Nuance.

PR #615, opened 2021-07-28
Updating omsagent yaml to have parity with omsagent yaml file in AKS RP

rashmichandrashekar:


kubernetes/omsagent.yaml
```diff
@@ -379,13 +379,19 @@ spec:
               memory: 225Mi
           env:
             # azure devops pipeline uses AKS_RESOURCE_ID and AKS_REGION hence ensure to uncomment these
+            - name: AKS_CLUSTER_NAME
```
vishiy, line 382: do you know where we use this and why this is needed ?

kubernetes/omsagent.yaml
```diff
@@ -379,13 +379,19 @@ spec:
...
+              value: "VALUE_AKS_CLUSTER_NAME"
             - name: AKS_RESOURCE_ID
               value: "VALUE_AKS_RESOURCE_ID_VALUE"
             - name: AKS_REGION
               value: "VALUE_AKS_RESOURCE_REGION_VALUE"
+            - name: FBIT_SERVICE_FLUSH_INTERVAL
```
vishiy, line 388: wouldn't it make sense to ad all the settings here and make then the default value ? But then i am assuming if customer over-rides them with a configmap, that will win over the defaults ?

ganga1980: Since we have reverted to default FBIT settings - https://github.com/microsoft/Docker-Provider/commit/cd2275354aafb588f0ed74cf8d747e40226e4974 and this will be removed, and there is a task on July release story to remove this from AKS RP yamls.

kubernetes/omsagent.yaml
```diff
@@ -379,13 +379,19 @@ spec:
...
+              value: "15"
+            - name: AKS_NODE_RESOURCE_GROUP
+              value: "VALUE_AKS_NODE_RESOURCE_GROUP"
             # this used for e2e test and setting this just emits some additional log statements which used for the e2e tests
-            - name: ISTEST  
-              value: "true"
+            #- name: ISTEST
```
ganga1980, line 393: what's the reason for commenting this ? This used for e2e test - https://github.com/microsoft/Docker-Provider#e2e-tests. If you comments these, e2e test will fail.

kubernetes/omsagent.yaml
```diff
@@ -379,13 +379,19 @@ spec:
...
             # azure devops pipeline uses AKS_RESOURCE_ID and AKS_REGION hence ensure to uncomment these
+            - name: AKS_CLUSTER_NAME
+              value: "VALUE_AKS_CLUSTER_NAME"
             - name: AKS_RESOURCE_ID
               value: "VALUE_AKS_RESOURCE_ID_VALUE"
             - name: AKS_REGION
               value: "VALUE_AKS_RESOURCE_REGION_VALUE"
+            - name: FBIT_SERVICE_FLUSH_INTERVAL
+              value: "15"
+            - name: AKS_NODE_RESOURCE_GROUP
```
ganga1980, line 390: I dont think we use this anywhere in our code. why do we need that?

kubernetes/omsagent.yaml
```diff
@@ -445,59 +453,65 @@ spec:
...
-          livenessProbe:
-            exec:
-              command:
-                - /bin/bash
-                - -c
-                - /opt/livenessprobe.sh
-            initialDelaySeconds: 60
-            periodSeconds: 60
-            timeoutSeconds: 15
+        # - name: omsagent-prometheus
```
ganga1980, line 456: since this can be controlled via SIDECAR_SCRAPING_ENABLED, why do we need to comment this?

rashmichandrashekar: This creates a sidecar container. the goal is to keep it consistent with AKS-RP. Commenting it now, will be uncommenting it when we start rolling out to regions,

kubernetes/omsagent.yaml
rashmichandrashekar: Double checking this, looks like this was added long ago and the history until june2020 on this file doesnt tell me who added this. In June, this file was moved and has existed even before that. But looks like we are not using this, so removing this env var.

kubernetes/omsagent.yaml
rashmichandrashekar: Removed this, since we will be reverting.

kubernetes/omsagent.yaml
rashmichandrashekar: Removed this

kubernetes/omsagent.yaml
rashmichandrashekar: Uncommenting this,

kubernetes/omsagent.yaml
```diff
@@ -789,13 +799,13 @@ spec:
...
           - name: SIDECAR_SCRAPING_ENABLED
-            value: "true"
+            value: "false"
           # Update this with the user assigned msi client id for omsagent
           - name: USER_ASSIGNED_IDENTITY_CLIENT_ID
             value: ""
           # Add this only for clouds that require cert bootstrapping
-          - name: REQUIRES_CERT_BOOTSTRAP
-            value: "true"
+          # - name: REQUIRES_CERT_BOOTSTRAP
```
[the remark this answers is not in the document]
rashmichandrashekar: I dont see it in windows ds in aks-rp yaml here - https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=%2Fccp%2Fcharts%2Fkube-control-plane%2Fcharts%2Fkube-addons%2Ftemplates%2Faddons%2Fpartner-addons%2F_omsagent.yaml
Could you give me the line number?

vishiy (approved):

kubernetes/omsagent.yaml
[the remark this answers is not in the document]
rashmichandrashekar: I dont see it in the aks-rp yaml as it was removed. Since this is no longer relevant, resolving the comment.

PR #618, opened 2021-08-02
emit log entries when more than 0 after exclusion

ganga1980:
If the namespace excluded in both stdout & stderr streams, then no issue since we don’t tail the logs of the container of that namespace completely.

But, If the namespace excluded either to stdout or stderr, then we drop messages of excluded stream after receiving log lines.

This is depends on the timing but If the batch, we received from the fluent bit tail plugin contains the **all log lines of excluded stream** then instead of dropping them, because of the bug in ciprod06112021, agent try to ingest incorrect path which will fail with 403 (which is expected) but those log lines keep retried.

This will not issue of the number of dropped messages but if the dropped messages are very high and that will cause, the number of log lines are retrying are very high, then that will impact the memory usage of td-agent-bit (aka Fluent bit) and eventually agent will OOM if the memory usage keep grows because of the retries .

In this customer case - https://github.com/Azure/AKS/issues/2457, the namespace excluded “baapi2-test” excluded for stdout but not stderr, the containers in this namespace generating very high number of stdout logs which are keep retrying incorrect path instead of dropping those log lines.

### DIFF
--- a/source/plugins/go/src/oms.go
+++ b/source/plugins/go/src/oms.go
@@ -1357,7 +1357,7 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 		numContainerLogRecords = len(dataItemsADX)
 		Log("Success::ADX::Successfully wrote %d container log records to ADX in %s", numContainerLogRecords, elapsed)
 
-	} else { //ODS
+	} else if (ContainerLogSchemaV2 == true && len(dataItemsLAv2) > 0) || len(dataItemsLAv1) > 0) { //ODS
 		var logEntry interface{}
 		recordType := ""
 		loglinesCount := 0
@@ -1437,7 +1437,7 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 		numContainerLogRecords = loglinesCount
 		Log("PostDataHelper::Info::Successfully flushed %d %s records to ODS in %s", numContainerLogRecords, recordType, elapsed)
 
-		}
+	}
 
 	ContainerLogTelemetryMutex.Lock()
 	defer ContainerLogTelemetryMutex.Unlock()
--- a/source/plugins/go/src/oms.go
+++ b/source/plugins/go/src/oms.go
@@ -165,17 +165,17 @@ var (
 	// ADX tenantID
 	AdxTenantID string
 	//ADX client secret
-	AdxClientSecret string	
+	AdxClientSecret string
 	// container log or container log v2 tag name for oneagent route
-	MdsdContainerLogTagName string 
+	MdsdContainerLogTagName string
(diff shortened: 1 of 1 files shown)


source/plugins/go/src/oms.go
```diff
@@ -1357,7 +1357,7 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 		numContainerLogRecords = len(dataItemsADX)
 		Log("Success::ADX::Successfully wrote %d container log records to ADX in %s", numContainerLogRecords, elapsed)
 
-	} else { //ODS
+	} else if ((ContainerLogSchemaV2 == true && len(dataItemsLAv2) > 0) || len(dataItemsLAv1) > 0) { //ODS
```
vishiy, line 1360: can we remove the ods code path ? If you had removed the path in the first place, we wouldn't have encountered this issue. Yes, we will flush an empty array to mdsd, but the behavior will be the same as before without triggering retries.

ganga1980: We cant remove ODS code path since this being used for the Windows agent as we dont have Windows AMA integration.  Until we integrate with Windows AMA, we have to keep this code path.

vishiy: sorry, dont agree. we can certainly remove ods code path in linux completely.

ganga1980: ODS code path used for windows and for Linux, as backup option, where we can suggest customer configures the "v1" route via config map in case if we ended getting any incidents related to container log missing with mdsd in high container log volume customers.  Certainly, we can remove ods path for linux completly in august or later release. lets keep this backup option as is. we can discuss more if you have concern on this.

vishiy: where did we discuss building a feature like this to revert back to v1 ?

ganga1980: > where did we discuss building a feature like this to revert back to v1 ?

> Please refer the add stage2 PR where I have mentioned and also responded to similar comment - https://github.com/microsoft/Docker-Provider/pull/559

vishiy: given we had this bug, what you had in the old PR doesn't matter. Just check that for linux node this path is completely mooted. We don't want a 'hidden' unused feature introducing bugs, just like it did. Previously this was needed, now we dont. Also, for comments in PR that you made no changes, please ensure the person putting the comment agree to your theories. You cannot just resolve them by yourselves, and ask someone else to approve the PR and merge.

ganga1980: created a task to address this august agent release to moot the ODS code path completely.

ganga1980:
since changes in this PR covered by cherr-pick from prod PR so closing this PR.

PR #620, opened 2021-08-05
Planned Prod Agent release fixes 

ganga1980:
PR has following changes

1.	Fix for the CPU spike which caused by unattended auto upgrades of the packages 
2.	MDSD fixes on top of drop what we released - https://github.com/microsoft/Docker-Provider/blob/ci_prod/kubernetes/linux/setup.sh#L13 for following issues
-	Fix for the core dump issue
-	Reduce the verbosity of the error logs. Only logged after 3 retry attempts. No error logs if API called failed within in retry threshold
-	Timeout of the OMS homing service increased to 60s from 30s as we see ~1288 clusters are seeing these calls are getting timedout         
3.	Fix for the AKS Github issue - https://github.com/Azure/AKS/issues/2457
4.	Tailing of mdsd.err log file in replicaset for the agent telemetry

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -17,9 +17,9 @@ Note : The agent version(s) below has dates (ciprod<mmddyyyy>), which indicate t
 - Linux Agent
   - Fix for CPU spike which occurrs at around 6.30am UTC on every day because of unattended package upgrades
   - Update MDSD build which has fixes for the following issues
-     -	Undeterministic Core dump issue because of the non 200 status code and runtime exception stack unwindings
-     -	Reduce the verbosity of the error logs for OMS & ODS code paths.
-     - 	Increase Timeout for OMS Homing service API calls from 30s to 60s
+     - Undeterministic Core dump issue because of the non 200 status code and runtime exception stack unwindings
+     - Reduce the verbosity of the error logs for OMS & ODS code paths.
+     - Increase Timeout for OMS Homing service API calls from 30s to 60s
    - Fix for https://github.com/Azure/AKS/issues/2457
    - In replicaset, tailing of the mdsd.err log file to agent telemetry
 
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -11,6 +11,19 @@ additional questions or comments.
 
 Note : The agent version(s) below has dates (ciprod<mmddyyyy>), which indicate the agent build dates (not release dates)
 
+### 08/05/2021 -
+##### Version microsoft/oms:ciprod08052021 Version mcr.microsoft.com/azuremonitor/containerinsights/ciprod:ciprod08052021 (linux)
+##### Code change log
+- Linux Agent
+  - Fix for CPU spike which occurrs at around 6.30am UTC on every day because of unattended package upgrades
+  - Update MDSD build which has fixes for the following issues
+     -	Undeterministic Core dump issue because of the non 200 status code and runtime exception stack unwindings
+     -	Reduce the verbosity of the error logs for OMS & ODS code paths.
+     - 	Increase Timeout for OMS Homing service API calls from 30s to 60s
+   - Fix for https://github.com/Azure/AKS/issues/2457
+   - In replicaset, tailing of the mdsd.err log file to agent telemetry
+
+
 ### 07/13/2021 -
 ##### Version microsoft/oms:win-ciprod06112021-2 Version mcr.microsoft.com/azuremonitor/containerinsights/ciprod:win-ciprod06112021-2 (windows)
 ##### Code change log
--- a/kubernetes/linux/setup.sh
+++ b/kubernetes/linux/setup.sh
@@ -9,8 +9,8 @@ sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
     dpkg-reconfigure --frontend=noninteractive locales && \
     update-locale LANG=en_US.UTF-8
 
-#install oneagent - Official bits (05/17/2021)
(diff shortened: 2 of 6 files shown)


kubernetes/linux/setup.sh
```diff
@@ -44,8 +44,8 @@ sudo apt-get update
 sudo apt-get install td-agent-bit=1.6.8 -y
 
 # install ruby2.6
-sudo apt-get install software-properties-common -y
-sudo apt-add-repository ppa:brightbox/ruby-ng -y
+sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys F5DA5F09C3173AA6
```
vishiy, line 47: looks ok to me. Only question is if we can get the keys from keyserver (jusu like we do for fluentbit).

kubernetes/linux/setup.sh
```diff
@@ -44,8 +44,8 @@ sudo apt-get update
...
+sudo echo "deb http://ppa.launchpad.net/brightbox/ruby-ng/ubuntu bionic main" >> /etc/apt/sources.list
 sudo apt-get update
 sudo apt-get install ruby2.6 ruby2.6-dev gcc make -y
```
vishiy, line 50: on related note: why do we need gcc, make, ruby-dev (which also puts ruby docs) packages?

kubernetes/linux/setup.sh
ganga1980: Yesterday, I have spent good amount of time to figure out if we can use the key file like we did FBIT but no luck on that. Hence, the proceed with approaching of getting key from ubuntu key server and which is other good alternate option.  AFIK, this seems better available option but if you have way to locate the key file and i am happy to change to that.

kubernetes/linux/setup.sh
ganga1980: > on related note: why do we need gcc, make, ruby-dev (which also puts ruby docs) packages?
these required to build the ruby c extensions and without these, we cant install fluentd-gem and these are minimal dependencies we would need to install for the fluentd-gem.  did very quick test, but didnt see ruby docs installed with ruby-dev.

vishiy: hmmm. i dont see any gcc or make dependencies (ruby depends on rake and not make).
Also, i dont fully understand why we need to install ruby this way.

'apt install ruby'  --> this just installs latest ruby runtime, and doesn't install ruby-docs, ruby-dev, gcc, make etc... i can install gems as well.

ganga1980: in 18.04, "apt install ruby" install 2.5 which is included in the default ubuntu repo.  and ruby 2.6 is not part of the native ubuntu repo. Are you saying, you can install **ruby 2.6 wit apt install on 1804**? when I brought ruby 2.6 & fluentd gem updates, i picked the critical dependencies which are required but I can double check and see if we can get rid of any unneeded dependencies.

ganga1980: double checked, we would need  all these dependencies ruby2.6-dev gcc make. gem part of the ruby2.6-dev and make and gcc required to build native extension.   I have image without ruby & fluentd - cidev:norubyandfluentd in case you want try out.
Error I was referring this
**Building native extensions. This could take a while...
ERROR:  Error installing fluentd:
        ERROR: Failed to build gem native extension.**

    current directory: /var/lib/gems/2.6.0/gems/msgpack-1.4.2/ext/msgpack
/usr/bin/ruby2.6 -I /usr/lib/ruby/2.6.0 -r ./siteconf20210805-901-10soqqt.rb extconf.rb
mkmf.rb can't find header files for ruby at /usr/lib/ruby/include/ruby.h

vishiy: so these are more fluentd dependencies than ruby  then ? also all these are install dependencies, After installing everything, we can remove all these dev tools, as we dont need them for runtime.

ganga1980: > so these are more fluentd dependencies than ruby then ? also all these are install dependencies, After installing everything, we can remove all these dev tools, as we dont need them for runtime.

> yes, as I called out in my initial response and these are dependencies to install fluentd gem.  will have to play removing all these dependencies  after installing fluentd gem and see fluentd gem has any runtime dependencies coming from these dpendencies.

vishiy: i dont think there are any runtime dependencies. what you are doing is 'ruby-full'. what we need at runtime is 'ruby'

ganga1980: looks like fluentd doesnt these dependencies for runtime so removed these build dependencies.

vishiy: Also, Ganga, when we remove packages, follow that with apt autoremove (as remove will only remove the packages asked to be removed and not its dependencies). So when you go apt autoremove does it remove more in this case ?

ReleaseNotes.md
```diff
@@ -11,6 +11,19 @@ additional questions or comments.
...
+##### Code change log
+- Linux Agent
+  - Fix for CPU spike which occurrs at around 6.30am UTC on every day because of unattended package upgrades
+  - Update MDSD build which has fixes for the following issues
+     - Undeterministic Core dump issue because of the non 200 status code and runtime exception stack unwindings
+     - Reduce the verbosity of the error logs for OMS & ODS code paths.
+     - Increase Timeout for OMS Homing service API calls from 30s to 60s
+   - Fix for https://github.com/Azure/AKS/issues/2457
+   - In replicaset, tailing of the mdsd.err log file to agent telemetry
+
```
vishiy, line 25: put a note to increase mem request & limit for ds? you can skip specifying values in readme here.

vishiy (approved):

ReleaseNotes.md
ganga1980: Sure, i will update the readme once req & limits in early next week.

PR #623, opened 2021-08-09
Fix out_oms.go dependency vulnerabilities

gracewehner:
Upgrade k8s.io/client-go and k8s.io/apimachinery used for container log enrichment
Upgrade ADX packages


.github/workflows/pr-checker.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           format: 'table'
           severity: 'CRITICAL,HIGH'
           vuln-type: 'os,library'
-          skip-dirs: 'opt/telegraf'
+          skip-dirs: 'opt/telegraf,usr/sbin/telegraf'
```
vishiy, line 59: Grace, so i am assuming that we are not taking in a newer telegraf as part of this? I remember you said that you tested the newer version, so just checking. I understand the newer version has a few vulnerabilities as well..

gracewehner: Right we can't really take a newer version since any newer version has the `monitor_kubernetes_pods=true` feature broken. I have created an [issue](https://github.com/influxdata/telegraf/issues/9600) and am about to make a PR for it. Or we would need to use our own custom build

vishiy (approved):

source/plugins/go/src/go.mod
```diff
@@ -3,33 +3,17 @@ module Docker-Provider/source/plugins/go/src
...
-	golang.org/x/net v0.0.0-20200421231249-e086a090c8fd // indirect
-	golang.org/x/time v0.0.0-20161028155119-f51c12702a4d // indirect
-	gopkg.in/inf.v0 v0.9.0 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0-20170531160350-a96e63847dc3
-	k8s.io/api v0.0.0-20180628040859-072894a440bd // indirect
-	k8s.io/apimachinery v0.0.0-20180621070125-103fd098999d
-	k8s.io/client-go v8.0.0+incompatible
-	golang.org/x/crypto v0.0.0-20201216223049-8b5274cf687f
+	k8s.io/apimachinery v0.21.0
+	k8s.io/client-go v0.21.0
```
ganga1980, line 18: @gracewehner, because of this update, did you find any compatibility issues with k8s versions < 1.17?  is this upgrade agnostic to k8s version?

gracewehner: @ganga1980 do we need compatibility for < 1.17? I thought we only support the same Kubernetes versions as AKS, even for non-AKS: https://docs.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-onboard#supported-configurations

ganga1980: @gracewehner , yes, our support matrix same as as what you have pointed out. I am trying to understand the impact because of this and based on that, we can decide whether we would need to cover in docs or not as part of this release. From our telemetry in last 24hr, i see we have 5K+ clusters with 1.14, 1.15 & 1.16 together. do you know what will be broken with versions < 1.17 version ?

vishiy: i dont understand where the "1.17" is coming from ? Where do you see this is not compatible with < 1.17 specifically? why not 1.18 ? The fix here is we were using clint go version that was deprecated with k8s 1.11 in 2019. So she is taking the latest version of client go. The only place this is used is when enrichment is turned ON.

gracewehner: I see 57 clusters with version < 1.17 and enrichment enabled. The client-go documentation is a bit vague about compatibility: https://github.com/kubernetes/client-go#compatibility-client-go---kubernetes-clusters

ganga1980: yes, that's what I was referring..

https://github.com/kubernetes/client-go#client-go
https://github.com/kubernetes/client-go/blob/master/INSTALL.md#using-a-specific-version
because of this client-go ( to tag v0.x.y) update, no issue to the < 1.17 clusters then we are good.  Since the documentation is very vague so I think, better to validate.

gracewehner: Thanks Ganga, I have now validated with 1.16 and 1.17 clusters and checked that there aren't errors in `fluent-bit-out-oms-runtime.log` for these clusters

ganga1980 (approved):

PR #627, opened 2021-08-12
increase ds memory requests and limis due to additional hop from mdsd

ganga1980:
analyzed 30+ clusters which are DS oom kill. The pattern of  increase in the memory usage not consistent. This could because of the telemetry sample interval what we collect.  Based on the analysis, 25% increase should give pretty good coverage. We can watch this agent rollout and see whether we would further need to adjust.

telemetry analysis details are here - OOM-killed-ds-clusters.xlsx in aka.ms/inin under containerinsights->agent directory.

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -22,6 +22,7 @@ Note : The agent version(s) below has dates (ciprod<mmddyyyy>), which indicate t
      - Increase Timeout for OMS Homing service API calls from 30s to 60s
    - Fix for https://github.com/Azure/AKS/issues/2457
    - In replicaset, tailing of the mdsd.err log file to agent telemetry
+   - Daemonset memory requests and limits increased by 25% due to additional hop with MDSD
 
 
 ### 07/13/2021 -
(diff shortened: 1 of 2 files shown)


ReleaseNotes.md
```diff
@@ -22,6 +22,7 @@ Note : The agent version(s) below has dates (ciprod<mmddyyyy>), which indicate t
      - Increase Timeout for OMS Homing service API calls from 30s to 60s
    - Fix for https://github.com/Azure/AKS/issues/2457
    - In replicaset, tailing of the mdsd.err log file to agent telemetry
+   - Daemonset memory requests and limits increased by 25% due to additional hop with MDSD
```
vishiy, line 25: I looked at the xls and dont see where you are seeing 25% correlation. I see some have doubled in memory usage.

ganga1980: > I looked at the xls and dont see where you are seeing 25% correlation. I see some have doubled in memory usage.
> yes, there is no consistent pattern and some cases, I see 20% and some cases I see 40% , and some cases it got doubled.  in one of the sample clusters, i see better memory usage with mdsd than the previous agent. It could be related to our telemetry sample interval and current thinking is increase 25% and watch the release, and see if we need to adjust further.

PR #629, opened 2021-08-17
check onboarding status

ganga1980:
This PR has following changes 
  1.  onboarding success status for both legacy and MSI auth and which will be used by AKS Black box telemetry to track onboarding status
  2.  MDSD update with recent MDSD fixes related OMS homing service  and reduce verbosity for the logs) on top of MDSD official version with MSI auth changes.

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -38,6 +38,51 @@ waitforlisteneronTCPport() {
       fi
 }
 
+checkAgentOnboardingStatus() {
+      local sleepdurationsecs=1
+      local totalsleptsecs=0
+      local isaadmsiauthmode=$1
+      local waittimesecs=$2
+      local numeric='^[0-9]+$'
+
+      if [ -z "$1" ] || [ -z "$2" ]; then
+            echo "${FUNCNAME[0]} called with incorrect arguments<$1 , $2>. Required arguments <#isaadmsiauthmode, #wait-time-in-seconds>"
+            return -1
+      else
+
+            if [[ $waittimesecs =~ $numeric ]]; then
+                  successMessage="Onboarding success"
+                  failureMessage="Failed to register certificate with OMS Homing service, giving up"
+                  if [ "${isaadmsiauthmode}" == "true" ]; then
+                        successMessage="Loaded data sources"
+                        failureMessage="Failed to load data sources into config"
+                  fi
+                  while true
+                  do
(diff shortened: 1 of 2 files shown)


kubernetes/linux/main.sh
```diff
@@ -38,6 +38,51 @@ waitforlisteneronTCPport() {
...
+      else
+
+            if [[ $waittimesecs =~ $numeric ]]; then
+                  successMessage="Onboarding success"
+                  failureMessage="Failed to register certificate with OMS Homing service, giving up"
+                  if [ "${isaadmsiauthmode}" == "true" ]; then
+                        successMessage="Loaded data sources"
+                        failureMessage="Failed to load data sources into config"
+                  fi
+                  while true
```
rashmichandrashekar, line 60: Is this going to be retried once every second for 30 seconds?

ganga1980: MDSD onboards the agent as soon as it starts. if it succeeds, mdsd.info will have success message, else if it fails, failed message in the mdsd.err logs. I havent seen case where neither succeed nor failure. Retry only happen if ever MDSD get into state where onboarding didnt triggered and in that case, retry happens every 1s for 30s. **This similar pattern what we are doing for telegraf.**

rashmichandrashekar: If there is no case where MDSD doesnt get into success or failure state why do we need the retries?

ganga1980: good question. we can certainly remove next agent release based on the telemetry and i dont see any harm having the retries.

rashmichandrashekar: I would suggest keeping it consistent with oneagent behavior. Based on offline discussion with Sid, looks like there will not be any delays since they dont have retries currently. So the aggressive 1s will not be a scenario we need to handle. And once they implement retries, the behavior is retry once every 30s for 3 minutes. So the current retry mechanism will not help.

rashmichandrashekar (approved):

kubernetes/linux/main.sh
rashmichandrashekar: Please create a task on onboard to track and address this later.

ganga1980: planning to watch telemetry and adjust this based on MDSD retry change which I am working with Sid to get that change.  30s should work even with retry buts lets watch telemetry.

PR #630, opened 2021-08-19
upgrade golang version for windows in pipeline and locally

gracewehner:
Upgrade to fix this error from the pipeline build and local build:
```
C:\Program Files\Go\pkg\tool\windows_amd64\link.exe: running gcc failed: exit status 1
C:/TDM-GCC-64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: error: export ordinal too large: 69814
```
See https://github.com/golang/go/issues/40795 and https://github.com/golang/go/issues/30674 (which confirms others reproducing with 1.14). This was patched for > 1.15.10 but not patched for 1.14

Will wait to merge until the image is pushed to cdpxwin1809.azurecr.io.


ganga1980 (approved):
lgtm

PR #636, opened 2021-09-01
Adding script to collect logs (for troubleshooting)

daweim0:
Adding a script to pull logs from a daemonset and replicaset agent pod. I frequently ask customers to collect these logs, making a script to grab and zip them should make the process faster for everyone involved.

The one way this script needs to be improved is that it doesn't collect any logs under /etc/omsagentwindows (kubectl cp was returning permission errors for that directory). I don't have time to investigate why now, and this script is still useful without it.

### DIFF
--- a/scripts/troubleshoot/collect_logs.sh
+++ b/scripts/troubleshoot/collect_logs.sh
@@ -1,21 +1,32 @@
 #!/bin/bash
 
+
+# This script pulls logs from the replicaset agent pod and a random daemonset pod. This script is to make troubleshooting faster
+
 export ds_pod=$(kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name | grep -E omsagent-[a-z0-9]{5} | head -n 1)
+export ds_win_pod=$(kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name | grep -E omsagent-win-[a-z0-9]{5} | head -n 1)
 export rs_pod=$(kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name | grep -E omsagent-rs-[a-z0-9]{5} | head -n 1)
 
-echo "collecting logs from ${ds_pod} and ${rs_pod}"
+echo "collecting logs from ${ds_pod}, ${ds_win_pod}, and ${rs_pod}"
+echo "    note: some erros are expected for clusters without windows nodes, they can safely be disregarded (filespec must match the canonical format:, zip warning: name not matched: omsagent-win-daemonset-fbit)"
 
-kubectl cp ${ds_pod}:/var/opt/microsoft/docker-cimprov/log omsagent-daemonset-log --namespace=kube-system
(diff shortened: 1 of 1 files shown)


scripts/troubleshoot/collect_logs.sh
```diff
@@ -0,0 +1,32 @@
...
+# This script pulls logs from the replicaset agent pod and a random daemonset pod. This script is to make troubleshooting faster
+
+export ds_pod=$(kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name | grep -E omsagent-[a-z0-9]{5} | head -n 1)
+export ds_win_pod=$(kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name | grep -E omsagent-win-[a-z0-9]{5} | head -n 1)
+export rs_pod=$(kubectl get pods -n kube-system -o custom-columns=NAME:.metadata.name | grep -E omsagent-rs-[a-z0-9]{5} | head -n 1)
+
+echo "collecting logs from ${ds_pod}, ${ds_win_pod}, and ${rs_pod}"
+echo "    note: some erros are expected for clusters without windows nodes, they can safely be disregarded (filespec must match the canonical format:, zip warning: name not matched: omsagent-win-daemonset-fbit)"
+
+kubectl cp ${ds_pod}:/var/opt/microsoft/docker-cimprov/log omsagent-daemonset --namespace=kube-system --container omsagent
```
vishiy, line 13: Thanks David. also would be good to add kubectl describe pod for each and dump to a file. For liveness failures, it will tell what caused them, and also how frequently its happening.

daweim0: Good idea, will add

scripts/troubleshoot/collect_logs.sh
```diff
@@ -0,0 +1,32 @@
...
+kubectl cp ${ds_pod}:/var/opt/microsoft/linuxmonagent/log omsagent-daemonset-mdsd --namespace=kube-system --container omsagent
+
+kubectl cp ${ds_pod}:/var/opt/microsoft/docker-cimprov/log omsagent-prom-daemonset --namespace=kube-system --container omsagent-prometheus
+kubectl cp ${ds_pod}:/var/opt/microsoft/linuxmonagent/log omsagent-prom-daemonset-mdsd --namespace=kube-system --container omsagent-prometheus
```
gracewehner, line 17: Thanks for adding this script, this is super useful. For the sidecar logs, would it be better to check if this container exists before trying to copy the logs, or should we just mention that it's ok if there are some errors such as "container omsagent-prometheus is not valid for pod omsagent-*"?

gracewehner (approved):

scripts/troubleshoot/collect_logs.sh
daweim0: My original thought was to add a note saying "some erros are expected don't pannic", but not printing errors is probably better. Now it only collects sidecar and windows logs if those containers exist

PR #637, opened 2021-09-02
bumping image tag to 08052021-1

daweim0:
Bumping the image tag so that we can toggle clusters back to the prod image (for clusters with a different image under the same tag as the prod image)


ganga1980 (approved):
lgtm

PR #638, opened 2021-09-07
documenting fbit tail plugin configmap settings.

daweim0:
A customer asked for this to be documented. I think I added enough of a disclaimer (that uncommenting the settings is not supported).

### DIFF
--- a/kubernetes/container-azm-ms-agentconfig.yaml
+++ b/kubernetes/container-azm-ms-agentconfig.yaml
@@ -150,7 +150,7 @@ data:
     # [agent_settings.fbit_config]
     #   log_flush_interval_secs = "15"
     #   tail_mem_buf_limit_megabytes = "10"
-    #   tail_buf_chunksize_megabytes= "1"
+    #   tail_buf_chunksize_megabytes = "1"
     #   tail_buf_maxsize_megabytes = "1"
 
 metadata:
--- a/kubernetes/container-azm-ms-agentconfig.yaml
+++ b/kubernetes/container-azm-ms-agentconfig.yaml
@@ -151,6 +151,7 @@ data:
     #   log_flush_interval_secs = "1"
     #   tail_buf_chunksize_megabytes= "1"
     #   tail_buf_maxsize_megabytes = "1"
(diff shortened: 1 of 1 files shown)


kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -144,6 +144,13 @@ data:
       tcp_listener_chunk_size = 10
       tcp_listener_buffer_size = 10
       tcp_listener_mem_buf_limit = 200
+    
+    # The following settings are "undocumented", we don't recommend uncommenting them unless directed by Microsoft.
+    # They increase the maximum stdout/stderr log collection rate but will also cause higher cpu/memory usage.
+    # [agent_settings.fbit_config]
+    #   log_flush_interval_secs = "1"
+    #   tail_buf_chunksize_megabytes= "1"
+    #   tail_buf_maxsize_megabytes = "1"
```
vishiy, line 153: how about mem_buf_limit ?

daweim0: good point, I added tail_mem_buf_limit_megabytes

vishiy: As comment, please put default values above each of these, so people know how much will they be increasing/decreasing to.

daweim0: I set log_flush_interval_secs and tail_mem_buf_limit_megabytes to their defaults, but that isn't possible for tail_buf_chunksize_megabytes and tail_buf_maxsize_megabytes (because we measure them in mb but the default value for both is 32k). Should I add a note that it's impossible to set those values to their default or leave them at 1mb?

daweim0: I added comments for tail_buf_chunksize_megabytes and tail_buf_maxsize_megabytes since their default is 32kb (and the configmap can only set them to integer megabytes)

vishiy: Thanks David.

vishiy (approved):

PR #653, opened 2021-09-27
Gangams/optimize win livenessprobe

ganga1980:
optimize liveness probe for the windows agent.  The existing liveness probe invokes multiple processes and creating process heavy in windows. The optimized liveness probe invokes native win32 APIs within in single process. Based on the testing, avg time taken for the liveness probe in msec and max is 4 sec on the busy node.  This change should address liveness probe issue and we can monitor once this deployed.

Perf data
image


daweim0:
(just curious, not requesting a change) why c++ instead of go? This adds another set of dependencies (the c++ distribution) to manage.

daweim0 (approved):
Looks good to me, just wondering why c++ instead of go

vishiy:
1) To check if a file exists, process exists & service is running, we need a seperate c++ client, is it ? How did you choose c++?
2) Do we know the root cause for inefficiency here ? Is it scquery for fd (o)r tasklist for fbit ? Simply put, if its scquery, we can just run fd as process and avoid scquery

daweim0:
One more thought, c++ will probably be harder than go to cross-compile for ARM (whenever we start supporting arm)

ganga1980:
> Looks good to me, just wondering why c++ instead of go
>  explored this path as well but go is heavy as we need to load additional go packages which requires additional dlls. 
cpp makes lightweight as we can native win32 apis which are available kernel32 dll.

ganga1980:
> 1. To check if a file exists, process exists & service is running, we need a seperate c++ client, is it ? How did you choose c++?
> 2. Do we know the root cause for inefficiency here ? Is it scquery for fd (o)r tasklist for fbit ? Simply put, if its scquery, we can just run fd as process and avoid scquery
> both tasklist and sc query are taking time which contributing over 15s(timeout) on busy nodes. with cpp, we can invoke native 32 apis directly without any additional overhead of loading dlls.

ganga1980:
> (just curious, not requesting a change) why c++ instead of go? This adds another set of dependencies (the c++ distribution) to manage.
> we already have build dependencies and there is no additional dependencies required.

> One more thought, c++ will probably be harder than go to cross-compile for ARM (whenever we start supporting arm)
> for ARM, afik, we just need building image for multi-arch, right? how the cross compiler playing role here?

vishiy:
> > 1. To check if a file exists, process exists & service is running, we need a seperate c++ client, is it ? How did you choose c++?
> > 2. Do we know the root cause for inefficiency here ? Is it scquery for fd (o)r tasklist for fbit ? Simply put, if its scquery, we can just run fd as process and avoid scquery
> >    both tasklist and sc query are taking time which contributing over 15s(timeout) on busy nodes. with cpp, we can invoke native 32 apis directly without any additional overhead of loading dlls.

Did we measure for tasklist & scquery individually?  not convinced that this is the right thing to do, by writing a c++ executable baking our entire liveness into it. If you want to design this well, the best thing to do would be wrap scquery as an executable and tasklist as another executable that we can use to compose our liveness instead of tasklist & scquery.

ganga1980:
> > > 1. To check if a file exists, process exists & service is running, we need a seperate c++ client, is it ? How did you choose c++?
> > > 2. Do we know the root cause for inefficiency here ? Is it scquery for fd (o)r tasklist for fbit ? Simply put, if its scquery, we can just run fd as process and avoid scquery
> > >    both tasklist and sc query are taking time which contributing over 15s(timeout) on busy nodes. with cpp, we can invoke native 32 apis directly without any additional overhead of loading dlls.
> 
> Did we measure for tasklist & scquery individually? not convinced that this is the right thing to do, by writing a c++ executable baking our entire liveness into it. If you want to design this well, the best thing to do would be wrap scquery as an executable and tasklist as another executable that we can use to compose our liveness instead of tasklist & scquery.
> Yes,  measured tasklist and scquery separately and both are contributing timeouts on busy nodes. wrapping them into separate executable will be better than released  prod version of the liveness probe but its not optimal as we will be creating 2 processes instead of one. with CPP code and single exe, we are getting optimal results as we are invoking lower level native 32 apis without additional overhead creating multiple processes and loading of any dependent dlls. Working with windows team, we have explored various ways to get the optimal one.

PR #656, opened 2021-09-28
Gangams/addon token adapter image tag to telemetry

ganga1980:
Add addon-token-adapter image tag to the telemetry so that we can use for the telemetry and track the version which would helpful for us to investigate any issues related.  I will adding dashboard tile to track this version .

### DIFF
--- a/source/plugins/ruby/in_containerinventory.rb
+++ b/source/plugins/ruby/in_containerinventory.rb
@@ -128,7 +128,7 @@ module Fluent::Plugin
           telemetryProperties["Computer"] = hostName
           telemetryProperties["ContainerCount"] = containerInventory.length
           if !addonTokenAdapterImageTag.empty?
-            telemetryProperties["addon-token-adapter"] = addonTokenAdapterImageTag
+            telemetryProperties["addonTokenAdapterImageTag"] = addonTokenAdapterImageTag
           end
           ApplicationInsightsUtility.sendTelemetry(@@PluginName, telemetryProperties)
         end
--- a/source/plugins/ruby/in_containerinventory.rb
+++ b/source/plugins/ruby/in_containerinventory.rb
@@ -57,6 +57,7 @@ module Fluent::Plugin
       containerInventory = Array.new
       eventStream = Fluent::MultiEventStream.new
       hostName = ""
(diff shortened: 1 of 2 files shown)


daweim0 (approved):

source/plugins/ruby/in_containerinventory.rb
```diff
@@ -82,6 +83,15 @@ def enumerate
                   if hostName.empty? && !containerRecord["Computer"].empty?
                     hostName = containerRecord["Computer"]
                   end
+                  if addonTokenAdapterImageTag.empty? && ExtensionUtils.isAADMSIAuthMode()
```
vishiy, line 86: aadonTokenAdapter is a local variable in enumerate which is being intialized to empty, every time, which means this will be evaluated for all containers in MSI mode every minute, until this container is found. This should be a global, that  should be set the first time, and then just use it as the container image for the sidecar wont change during our lifetime until we are redeployed/updated/restarted.

source/plugins/ruby/in_containerinventory.rb
```diff
@@ -82,6 +83,15 @@ def enumerate
...
+                     if !containerRecord["ElementName"].nil? && !containerRecord["ElementName"].empty? &&
+                      containerRecord["ElementName"].include?("kube-system") &&
```
vishiy, line 88: so if i have a kube-system-2 or my-kube-system namespace with a pod having same container name, we will pick that image ?

source/plugins/ruby/in_containerinventory.rb
```diff
@@ -82,6 +83,15 @@ def enumerate
                   if hostName.empty? && !containerRecord["Computer"].empty?
                     hostName = containerRecord["Computer"]
                   end
+                  if addonTokenAdapterImageTag.empty? && ExtensionUtils.isAADMSIAuthMode()
+                     if !containerRecord["ElementName"].nil? && !containerRecord["ElementName"].empty? &&
```
vishiy, line 87: I also disagree with this approach. Much neater way to do this would be, pass this image as an env variable from yaml. Define a helm variable for sidecar container, and use that for the spec.container.image as well pass the same as env variable to be picked up by our telemetry

PR #686, opened 2021-11-03
fix: subtract number instead of string + update fluentd version 1.14.2 to fix security vulnerability

bragi92:
Things fixed in this PR
- This is a fix for the `undefined method '-' for "0":String` error.
- fluentd minor version upgrade from 1.12.2 to 1.14.2
- tomplparser.rb syntax error

Fluentd security vulnerability : https://avd.aquasec.com/nvd/cve-2021-41186/
image

### DIFF
--- a/source/plugins/ruby/kubelet_utils.rb
+++ b/source/plugins/ruby/kubelet_utils.rb
@@ -47,8 +47,8 @@ class KubeletUtils
           @log.error "kubelet_utils.rb::get_node_allocatble - cpu_capacity or memory_capacity values not set. Hence we cannot calculate allocatable values"
         end
 
-        cpu_capacity = BigDecimal(cpu_capacity).to_f
-        memory_capacity = BigDecimal(memory_capacity).to_f
+        cpu_capacity = BigDecimal(cpu_capacity, 2).to_f
+        memory_capacity = BigDecimal(memory_capacity, 2).to_f
 
         cpu_allocatable = 1.0
         memory_allocatable = 1.0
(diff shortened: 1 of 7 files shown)


rashmichandrashekar (dismissed):

source/plugins/ruby/kubelet_utils.rb
```diff
@@ -56,74 +59,68 @@ def get_node_allocatable(cpu_capacity, memory_capacity)
...
           evictionHard_memory = parsed_response["kubeletconfig"]["evictionHard"]["memory.available"]
           if evictionHard_memory.nil? || evictionHard_memory == ""
-            evictionHard_memory = "0"
+            evictionHard_memory = "0.0"
           end
           @log.info "get_node_allocatable::evictionHard_memory #{evictionHard_memory}"
         rescue => errorStr
           @log.error "Error in get_node_allocatable::evictionHard_memory: #{errorStr}"
-          evictionHard_memory = "0"
-          ApplicationInsightsUtility.sendExceptionTelemetry("Error in get_node_allocatable::kubereserved_cpu: #{errorStr}")
```
daweim0, line 126: why remove the exception telemetry?

bragi92: I feel like this is logging too much telemetry in our appinsights resource. Right now its being logged in the filter_cadvisor2mdm.log file maintained on the node via the @log.error line and there is no action item to take from this telemetry point as I'm setting the default values to 0.0 if they don't exist in the response. 

I can add it back in again if you think its adding valuable information.

daweim0: I think it's good to keep. Since the input data comes from CAdvisor (which we don't control) this might alert us to an actual problem.

daweim0 (approved):

source/plugins/ruby/kubelet_utils.rb
bragi92: I've added it back in again, thanks. Resolving this comment now.

daweim0 (approved):

daweim0:
I'm merging this PR so other PRs can merge again. Kaveesh said that it was ready for merge in Teams

PR #687, opened 2021-11-04
Faster Linux builds (part 1)

daweim0:
Since the Docker build arg IMAGE_TAG is used very early in the Linux Dockerfile, changing it invalidates the local docker build cache. Moving the arg lower in the docker file lets Docker cache some apt installs and speeds up subsequent builds. (this optimization shaves two minutes off build times on my dev machine)

I don't think the arg is used during image builds (just runtime), but I didn't move it all the way down to the bottom of the dockerfile just to be sure nothing breaks.

### DIFF
--- a/kubernetes/linux/Dockerfile
+++ b/kubernetes/linux/Dockerfile
@@ -17,7 +17,7 @@ ENV RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR 0.9
 RUN /usr/bin/apt-get update && /usr/bin/apt-get install -y libc-bin wget openssl curl sudo python-ctypes init-system-helpers  net-tools rsyslog cron vim dmidecode apt-transport-https gnupg && rm -rf /var/lib/apt/lists/*
 COPY setup.sh main.sh defaultpromenvvariables defaultpromenvvariables-rs defaultpromenvvariables-sidecar mdsd.xml envmdsd logrotate.conf $tmpdir/
 
-ARG IMAGE_TAG=ciprod08052021
+ARG IMAGE_TAG=ciprod10132021
 ENV AGENT_VERSION ${IMAGE_TAG}
 
 WORKDIR ${tmpdir}
--- a/kubernetes/linux/Dockerfile
+++ b/kubernetes/linux/Dockerfile
@@ -2,8 +2,6 @@ FROM ubuntu:18.04
 MAINTAINER OMSContainers@microsoft.com
 LABEL vendor=Microsoft\ Corp \
     com.microsoft.product="Azure Monitor for containers"
-ARG IMAGE_TAG=ciprod10132021
-ENV AGENT_VERSION ${IMAGE_TAG}
 ENV tmpdir /opt
 ENV APPLICATIONINSIGHTS_AUTH NzAwZGM5OGYtYTdhZC00NThkLWI5NWMtMjA3ZjM3NmM3YmRi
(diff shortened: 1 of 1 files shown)


kubernetes/linux/Dockerfile
```diff
@@ -18,6 +16,10 @@ ENV KUBE_CLIENT_BACKOFF_DURATION 0
 ENV RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR 0.9
 RUN /usr/bin/apt-get update && /usr/bin/apt-get install -y libc-bin wget openssl curl sudo python-ctypes init-system-helpers  net-tools rsyslog cron vim dmidecode apt-transport-https gnupg && rm -rf /var/lib/apt/lists/*
 COPY setup.sh main.sh defaultpromenvvariables defaultpromenvvariables-rs defaultpromenvvariables-sidecar mdsd.xml envmdsd logrotate.conf $tmpdir/
+
+ARG IMAGE_TAG=ciprod08052021
```
sarahpeiffer, line 20: image tag should still be latest ciprod10132021

ganga1980 (approved):
lgtm.  good to know that, moving these helping on the perf.

kubernetes/linux/Dockerfile
daweim0: oops, should have caught that

kubernetes/linux/Dockerfile
```diff
@@ -18,6 +16,10 @@ ENV KUBE_CLIENT_BACKOFF_DURATION 0
 ENV RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR 0.9
 RUN /usr/bin/apt-get update && /usr/bin/apt-get install -y libc-bin wget openssl curl sudo python-ctypes init-system-helpers  net-tools rsyslog cron vim dmidecode apt-transport-https gnupg && rm -rf /var/lib/apt/lists/*
 COPY setup.sh main.sh defaultpromenvvariables defaultpromenvvariables-rs defaultpromenvvariables-sidecar mdsd.xml envmdsd logrotate.conf $tmpdir/
+
```
vishiy, line 19: Moving these 2 to higher layers makes it faster ?

daweim0: yeah, assuming you use a new image tag every time you build an image (which at least I always do so I can be sure I'm testing my latest build).  The problem is that whenever AGENT_TAG changes it invalidates all cached layers before it. By moving the IMAGE_TAG arg as far down as possible it will invalidate as few cached layers as possible

vishiy: that makes sense, but then to capitalize on that you need to be building multiple builds/images in the same node (like you dev node). It wont make much difference on build containers as these layers are missing there every single time we build .

daweim0: I completely agree, it only helps for repeated builds (like on my dev machine). It really helps when repeatedly building images though. 

For example, yesterday I built ~10 dev images. Each one took about 5 minutes, so that's a lot of wasted time. This change would have brought each local build down to threeish minutes.

vishiy: ok. thanks David. Also there are ways to remove this image creation time tag passing to be used for our telemetry, which will completely remove these 2 (ARG & ENV). I will resolve this comment.

ganga1980 (approved):

PR #693, opened 2021-12-16
Detect log loss

daweim0:
This PR adds detection for when fluent bit dropps logs. There are two components to do this:
Job 1: out_oms counts the volume of container logs (in bytes) which fluent bit passes to it
Job 2: out_oms separately watches the size of container log files on disk (/var/logs/pod/xxx/xxx) and keeps a count of bytes written to them
every 5 minutes, the two counts are compared. If more logs were written to disk than made it to out_oms, then fluent bit must have dropped them

I put a lot of effort into never blocking Process_log(), it will never wait to aquire a mutex as a result of this PR. The only synchronization between the main log-processing thread and other threads is through a flag variable (written and read with atomic memory operations) and a channel (which is never blocked on).

# Usage

The feature is off by default, add this configmap setting to turn it on.
```
log-data-collection-settings: |-
  [log_collection_settings]
    [log_collection_settings.track_dropped_logs]
      enabled = true
```

# Scale testing results:
I did a scale test with 28 10 minute trials (it stopped early because by computer restarted, probably Windows Update). 26 trials used this PR and 22 used the current production agent code (re-built on my machine for a more even comparison). They were all done on the same single-node cluster. VM size was Standard_D8s_v3. The log load was generated by 40 pods each logging 5 mb per minute to stdout in 50 byte lines.

Each one followed this procedure (which was automated):
1. All previous omsagent and log-generator pods were deleted and re-created
2. Wait for all pods to become ready (including log-generator pods)
3. Over 10 minutes: record the overall pod CPU and memory usage, cpu & memory usage for fluent-bit, mdsd, telegraf, and fluentd 
4. After 10 minutes, query Kusto for the total number of log bytes which made it to Kusto and the total number of bytes dropped before reaching Kusto (calculated using sequence numbers at the beginning of each line logged by the log-generator pods)

I found that this PR decreased log collection rate by 0.064%. This PR decreased allup agent memory usage by about 3 mb (633.9mb vs 630.4mb). I strongly suspect neither of these two differences are statistically significant (the differences are very small and it doesn't make sense that this PR decrased memory usage). The difference in CPU usage was not significant since the agent was being CPU throttled.

Since the graphs below show slightly higher memory usage by this PR, I also did a scale test with 8 3-hour runs (four with this PR and four with the currently released agent). I don't have numbers on how much data made it to LA (the queries were too mighty for LA to handle), but it showed average memory usage went up by 4mb from ci_dev to this PR (not significant). interestingly, Fbit's average memory usage specifically went down by 500kb, most of the increase came from mdsd (3mb) which was not modified by this PR.

# Sample performance graphs:
These graphs were not from the scale test, they show performance over a longer time period. The two agents (ci_dev and this PR) were run on two different nodes in the same cluster. Which agents switched nodes between the first and second runs.

## Run 1:
CPU usage
Mem usage (working set)
Logs collected

## Run 2:
CPU usage
Mem usage (working set)
Logs collected

I was suprised by the memory usage difference in these graphs, so I ran an additional scale test with 3 hour trials. It only showed 4mb difference in average memory usage, but large standard deviation (which explains these graphs)


source/plugins/go/src/log_loss_monitoring.go
```diff
@@ -0,0 +1,277 @@
...
+
+	if enabled {
+		log_loss_logger = createLogger("container-log-counts.log")
+
+		for _, excluded_namespace := range strings.Split(os.Getenv("AZMON_STDERR_EXCLUDED_NAMESPACES"), ",") {
+			disabled_namespaces[excluded_namespace] = true
+		}
+
+		for _, excluded_namespace := range strings.Split(os.Getenv("AZMON_STDOUT_EXCLUDED_NAMESPACES"), ",") {
+			disabled_namespaces[excluded_namespace] = true
```
gracewehner, line 96: What happens when a namespace is excluded for only stdout or stderr but not both?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -43,6 +43,11 @@ data:
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
 
+        [log_collection_settings.track_dropped_logs]
+          # Collect statistics about wether all container logs make it to Log Analytics (logs can be dropped if the data rate is too high)
+          enabled = true
+          # When this is disabled (enabled = false), statistics about log collection throttling will not be kept
```
gracewehner, line 49: We might want to explain the stats are only for our internal telemetry

source/plugins/go/src/log_loss_monitoring.go
daweim0: I'm completely ignoring namespaces with any customization on them (including stdout or stderr disabled), no data is recorded from them

kubernetes/container-azm-ms-agentconfig.yaml
daweim0: Good point, I wasn't sure if I should add it to the configmap at all (if it's not there then we'll have a hard time referencing it later)

source/plugins/go/src/oms.go
```diff
@@ -410,6 +412,20 @@ func createLogger() *log.Logger {
 	})
 
 	logger.SetFlags(log.Ltime | log.Lshortfile | log.LstdFlags)
+
+	// flush fluent-bit-out-oms-runtime.log at a reasonable rate (by default it's flushed like once a minute?)
+	TimeTicker := time.NewTicker(5 * time.Second)
```
daweim0, line 417: This change is entirely unrelated to detecting log loss, I'm happy to get rid of it. By default, some of the fbit log files are only flushed every few minutes which can make debugging a lot slower (since you have to wait for the log file to be flushed). This speeds that up

kubernetes/container-azm-ms-agentconfig.yaml
daweim0: (I'm open to suggestions either way)

build/common/installer/scripts/tomlparser.rb
```diff
@@ -191,6 +192,18 @@ def populateSettingValuesFromConfigMap(parsedConfig)
     rescue => errorStr
       ConfigParseErrorLogger.logError("Exception while reading config map settings for adx database name - #{errorStr}, using default #{@adxDatabaseName}, please check config map for errors")
     end
+
+    #Get log loss tracking disablement setting
+    begin
+      if !parsedConfig[:log_collection_settings][:track_lost_logs].nil? && !parsedConfig[:log_collection_settings][:track_lost_logs][:enabled].nil?
+        if !parsedConfig[:log_collection_settings][:track_lost_logs][:enabled].empty?
+           @disableLogLossTracking = parsedConfig[:log_collection_settings][:track_dropped_logs][:enabled]
```
ganga1980, line 200: what's the difference between track_dropped_logs vs track_lost_logs? i see the top level if checks for track_lost_logs but where as value of track_dropped_logs  being assigned at line # 200?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -34,14 +34,19 @@ data:
...
+          # When this is enabled (enabled = true), every container log entry (both stdout & stderr) will be enriched with container Name & container Image
        [log_collection_settings.collect_all_kube_events]
-          # In the absense of this configmap, default value for collect_all_kube_events is false
+          # In the absense of this configmap, default value for collect_all_kube_events is false
           # When the setting is set to false, only the kube events with !normal event type will be collected
           enabled = false
-          # When this is enabled (enabled = true), all kube events including normal events will be collected
+          # When this is enabled (enabled = true), all kube events including normal events will be collected
+
+        [log_collection_settings.track_dropped_logs]
```
ganga1980, line 46: whats the default value in the absence of this configmap?

source/plugins/go/src/utils.go
```diff
@@ -259,3 +261,119 @@ func convertMsgPackEntriesToMsgpBytes(fluentForwardTag string, msgPackEntries []
...
+	output_map := make(map[string]int64)
+	getSizeOfAllFilesInDirImpl(&root_dir, "", &output_map)
+	return output_map
+}
+
+func getSizeOfAllFilesInDirImpl(root_dir *string, preceeding_dir string, storage_dict *map[string]int64) {
+	// container_full_path := filepath.Join(preceeding_dir_segments)
+	files_and_folders, err := ioutil.ReadDir(filepath.Join(*root_dir, preceeding_dir))
+	if err != nil {
+		Log("ERROR: reading dir " + err.Error())
```
ganga1980, line 278: we can just return if there is any error for ioutil.ReadDir rather proceeding further, right?

source/plugins/go/src/log_loss_monitoring.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		write_counts_ticker := time.NewTicker(5 * time.Minute)
+		go write_telemetry(write_counts_ticker.C)
+
+		//TODO: turn this frequency down
+		track_rotations_ticker := time.NewTicker(10 * time.Second)
```
ganga1980, line 129: Make duration values  5 and 10 as constants so that it can be tuned easily if required.

kubernetes/container-azm-ms-agentconfig.yaml
daweim0: Log loss tracking is enabled by default

source/plugins/go/src/utils.go
daweim0: Good point, fixing that

source/plugins/go/src/log_loss_monitoring.go
```diff
@@ -0,0 +1,303 @@
...
+			// this branch only executes when a new container is seen, so it doesn't have to be as optimized
+			process_logs_mut.Lock() // only grab this lock when a new container is seen. Slices are thread safe so we don't need to lock it
+			defer process_logs_mut.Unlock()
+
+			// do garbage collection
+			if float32(atomic.LoadInt32(&num_containers_on_disk)) > float32(m_bytes_logged_storage.len())*3 {
+				// send some containerIDs to the FW thread to check if they are deleted from disk
+			write_to_chan:
+				for i := 0; i < 5; i++ {
+					deletion_query_index := (m_deletion_query_index + 1) % m_bytes_logged_storage.len()
```
ganga1980, line 151: I see, m_deletion_query_index value is 0 in all instances, which means deletion_query_index always same,  then how do you ensure your picking random containerids?

source/plugins/go/src/log_loss_monitoring_test.go
```diff
@@ -0,0 +1,139 @@
...
+	logEntry := "1234567890"
+	logTime := "1970/01/01T01:01:01.00001Z"
+
+	enabled = true
+
+	identifier := k8sNamespace + "_" + k8sPodName + "_" + containerName
+	if len(m_bytes_logged_storage.container_identifiers) != 0 {
+		t.Error("m_bytes_logged_storage didn't start empty")
+	}
+
```
ganga1980, line 27: are not these doing the same check? looks like duplicate?

source/plugins/go/src/log_loss_monitoring.go
```diff
@@ -0,0 +1,303 @@
...
+	m_deletion_query_index = 0
+
+	FW_records = make(map[string]FwRecord) // identifier -> file names
+	// var FW_bytes_logged_rotated = make(map[string]int64)   // identifier -> count
+	// var FW_bytes_logged_unrotated = make(map[string]int64) // identifier -> count
+	// var current_iteration_count int64 = 0
+	num_containers_on_disk = 0
+
+	// Shared objects
+	deleted_containers_query = make(chan string, 5)    // There might be a lot more than 5 containers  of 10 chosen arbitrairly
```
ganga1980, line 94: probable having constant for the capacity would make easy to change if needed.

source/plugins/go/src/log_loss_monitoring.go
```diff
@@ -0,0 +1,303 @@
...
+			if len(filepath_parts) != 3 {
+				Log("illegal file found in pod log dir: %v", filepath)
+				continue
+			}
+			pod_folder_parts := strings.Split(filepath_parts[0], "_") // this is safe because kubernetes objects can't have _ in their name
+			namespace_folder := pod_folder_parts[0]
+			pod_folder := pod_folder_parts[1]
+			container_folder := filepath_parts[1]
+			log_file_name := filepath_parts[2]
+			container_identifier := namespace_folder + "_" + pod_folder + "_" + container_folder // should be of the format namespace_podname-garbage_containername
```
ganga1980, line 247: format is namespace_<podname>_containerid.  podname is podname-<hashvalue>

build/common/installer/scripts/tomlparser.rb
daweim0: That's an embarrasing bug, fixed.

kubernetes/container-azm-ms-agentconfig.yaml
daweim0: It is disabled by default

source/plugins/go/src/log_loss_monitoring.go
daweim0: That's another embarrasing bug, there's now a unit test for it

source/plugins/go/src/log_loss_monitoring_test.go
daweim0: embarasing bug #3, now fixed

source/plugins/go/src/log_loss_monitoring.go
daweim0: done

source/plugins/go/src/log_loss_monitoring.go
daweim0: The format is pod-name_pod-uuid/container-name/log-file.log since we're looking at the pods log folder instead of the containers log folder

build/linux/installer/scripts/livenessprobe.sh
```diff
@@ -4,7 +4,7 @@
 (ps -ef | grep "mdsd" | grep -v "grep")
 if [ $? -ne 0 ]
 then
-  echo "mdsd is not running" > /dev/termination-log
+  echo "mdsd is not running" >> /dev/termination-log
```
daweim0, line 7: I'm adding the termination log to telemetry traces

build/common/installer/scripts/tomlparser.rb
```diff
@@ -285,6 +300,10 @@ def get_command_windows(env_variable_name, env_variable_value)
     file.write(commands)
     commands = get_command_windows('AZMON_ADX_DATABASE_NAME', @adxDatabaseName)
     file.write(commands)
+    commands = get_command_windows('AZMON_ENABLE_LOG_LOSS_TRACKING', @disableLogLossTracking)
```
ganga1980, line 303: log loss detection logic only scoped to linux, right? if yes, do we need these environment variables for windows?

daweim0: I removed the env vars for windows, there should now be no Windows changes.

daweim0:
bump

daweim0:
bump

vishiy:
whats up david ?

daweim0:
Hey Vishwa, sorry I never responded (better 9 months late than never?) I'm doing well, all things considered. A lot of my org was laid off in January, some teams are mostly gone. Was Azure Monitor hit?

PR #696, opened 2022-01-05
turning off all processes in the prometheus sidecar when there are no metrics to scrape (and some main.sh cleanup)

daweim0:
First: this PR keeps telegraf, mdsd, and fluent-bit from starting in the prometheus sidecar if it has no work to do (monitor_kubernetes_pods = false and no OSM namespaces to scrape). This part is just a resource-usage optimization.

Second: this PR cleans up main.sh. Previously, this pattern was used a lot in main.sh:
```
 export customRegion=$AKS_REGION
 echo "export customRegion=$AKS_REGION" >> ~/.bashrc
 source ~/.bashrc
```
This PR moves that logic to a helper function (setGlobalEnvVar). Additionally, instead of writing the variables to bashrc, they get written to /opt/env_vars. Sourcing bashrc in main.sh does nothing because bashrc immediately exits if it's started in a non-interractive shell (which main.sh runs in). By putting the variable exports in /opt/env_vars, main.sh and other scripts running in non-interractive shells (like the liveness probe) can get the correct set of environment variables and they will be the same as if someone `kubectl exec`(s) into a omsagent pod.

Third, setup.sh now deletes telegraf-1.18.0_linux_amd64.tar.gz after telegraf is extracted. There's no reason to keep that file in our final image, removing it makes our final image 25mb smaller. (running apt autoremove at the end of setup.sh would save even more space, but I don't want to thoroughly test that now)

Fourth: I removed some special characters from container-azm-ms-agentconfig.yaml. (0xC2xA0 (non-breaking space) was used for spaces instead of 0x20 (askii space)).

### DIFF
--- a/build/common/installer/scripts/tomlparser-prom-customconfig.rb
+++ b/build/common/installer/scripts/tomlparser-prom-customconfig.rb
@@ -323,7 +323,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
 
             File.open(file_name, "w") { |file| file.puts new_contents }
             puts "config::Successfully substituted the placeholders in telegraf conf file for custom prometheus scraping"
-            #Set environment variables for telemetry in the sidecar container
+            #Set environment variables in the sidecar container
             if (!@containerType.nil? && @containerType.casecmp(@promSideCar) == 0)
               file = File.open("prom_config_env_var", "w")
               if !file.nil?
--- a/build/common/installer/scripts/tomlparser-prom-customconfig.rb
+++ b/build/common/installer/scripts/tomlparser-prom-customconfig.rb
@@ -323,7 +323,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
 
             File.open(file_name, "w") { |file| file.puts new_contents }
             puts "config::Successfully substituted the placeholders in telegraf conf file for custom prometheus scraping"
-            #Set environment variables in the sidecar container
+            #Set environment variables for configuration and telemetry in the sidecar container
             if (!@containerType.nil? && @containerType.casecmp(@promSideCar) == 0)
               file = File.open("prom_config_env_var", "w")
               if !file.nil?
--- a/build/common/installer/scripts/tomlparser-prom-customconfig.rb
+++ b/build/common/installer/scripts/tomlparser-prom-customconfig.rb
@@ -223,7 +223,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
             File.open(file_name, "w") { |file| file.puts new_contents }
             puts "config::Successfully substituted the placeholders in telegraf conf file for replicaset"
             #Set environment variables for telemetry
-            file = File.open("telemetry_prom_config_env_var", "w")
+            file = File.open("prom_config_env_var", "w")
             if !file.nil?
               file.write("export TELEMETRY_RS_PROM_INTERVAL=\"#{interval}\"\n")
               #Setting array lengths as environment variables for telemetry purposes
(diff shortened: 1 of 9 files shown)


kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -34,14 +34,14 @@ data:
           # In the absense of this configmap, default value for enabled is true
           enabled = true
        [log_collection_settings.enrich_container_logs]
-          # In the absense of this configmap, default value for enrich_container_logs is false
```
daweim0, line 37: This file had a bunch of invisible unicode characters (not sure what they were, probably some fancy formatting or something). I just removed them

rashmichandrashekar:
Please make sure you test out all environment variable changes thoroughly. I have had issues in the past where the variable not set through bashrc were not picked up.

kubernetes/container-azm-ms-agentconfig.yaml
rashmichandrashekar: Could you check when you do a kubectl edit on the configmap it shows up in readable format after this change?

kubernetes/linux/main.sh
```diff
@@ -368,12 +375,12 @@ else
...
-#Sourcing telemetry environment variable file if it exists
-if [ -e "telemetry_prom_config_env_var" ]; then
-      cat telemetry_prom_config_env_var | while read line; do
-            echo $line >> ~/.bashrc
+#Sourcing environment variable file if it exists. This file has telemetry and whether kubernetes pods are monitored
+if [ -e "prom_config_env_var" ]; then
+      cat prom_config_env_var | while read line; do
+            echo $line >> /opt/env_vars
       done
-      source telemetry_prom_config_env_var
```
rashmichandrashekar, line 376: Why is telemetry prefix removed from these files? They help differetiate telemetry from non tel

daweim0:
> Please make sure you test out all environment variable changes thoroughly. I have had issues in the past where the variable not set through bashrc were not picked up.

I think I've tested all the combinations of settings, but I would feel a lot better if someone else tested it too. (since I've missed scinerios when testing before)

kubernetes/linux/main.sh
daweim0: I'm using an environment variable from that file (CUSTOM_PROM_MONITOR_PODS) to help determine if the sidecar should be turned off or not. I changed the file to not be named telementry since it isn't just telemetry any more. 

I could write that data to a new file instead if that's more clear? (I'm open to suggestions)

kubernetes/linux/main.sh
```diff
@@ -417,12 +424,21 @@ if [[ ( ( ! -e "/etc/config/kube.conf" ) && ( "${CONTAINER_TYPE}" == "Prometheus
 
       if [ -e "integration_osm_config_env_var" ]; then
             cat integration_osm_config_env_var | while read line; do
-                  echo $line >> ~/.bashrc
+                  echo $line >> /opt/env_vars
```
rashmichandrashekar, line 427: From what I see, /opt/env_vars is not sourced after writing to it. Will this work?

daweim0: Yeah, it works because right after the loop `source integration_osm_config_env_var` is called (like the other loops around it). This should be slightly more efficient than `source /opt/env_vars` because integration_osm_config_env_var will be a smaller file

kubernetes/linux/main.sh
daweim0: Discussed offline, naming will stay the same

kubernetes/container-azm-ms-agentconfig.yaml
daweim0: Checked, the format isn't changed with this change (some parts of the file are non-human-readable, no idea why)

kubernetes/linux/main.sh
rashmichandrashekar: So then why do we have to write it to /opt/env_vars?

build/common/installer/scripts/tomlparser-prom-customconfig.rb
```diff
@@ -325,10 +325,10 @@ def populateSettingValuesFromConfigMap(parsedConfig)
             puts "config::Successfully substituted the placeholders in telegraf conf file for custom prometheus scraping"
             #Set environment variables for telemetry in the sidecar container
             if (!@containerType.nil? && @containerType.casecmp(@promSideCar) == 0)
-              file = File.open("telemetry_prom_config_env_var", "w")
+              file = File.open("prom_config_env_var", "w")
               if !file.nil?
                 #Setting array lengths as environment variables for telemetry purposes
```
rashmichandrashekar, line 330: Could you pls update this comment since it is being used for purposes other than telemetry?

daweim0: Done

kubernetes/linux/main.sh
daweim0: So any other processes in the container will also get the change (liveness probe and when user execs into a container). (I guess for the same reason environment variables were previously written to ~/.bashrc)

rashmichandrashekar: But we are not sourcing after writing to this file (/opt/env_vars)

PR #697, opened 2022-01-20
Updating pod annotation for latest agent version

rashmichandrashekar:
1. Updated agent version from 1.10.0-1 to azure-mdsd-1.14.2
2. Made the change for windows ds as well since it had the agent version even though agent is not used in windows since 1.10.0-1 may create confusion regarding the omi issue even though we dont use it.

### DIFF
--- a/kubernetes/omsagent.yaml
+++ b/kubernetes/omsagent.yaml
@@ -357,7 +357,7 @@ spec:
         component: oms-agent
         tier: node
       annotations:
-        agentVersion: "1.10.0.1"
+        agentVersion: "azure-mdsd-1.14.2"
         dockerProviderVersion: "16.0.0-0"
         schema-versions: "v1"
     spec:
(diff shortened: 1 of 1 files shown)


kubernetes/omsagent.yaml
```diff
@@ -765,7 +765,7 @@ spec:
         component: oms-agent-win
         tier: node-win
       annotations:
-        agentVersion: "1.10.0.1"
+        agentVersion: "azure-mdsd-1.14.2"
```
ganga1980, line 768: we dont have AMA dependency yet for windows agent so i think this can be removed and added back when we get that integration. most likely version will be different for windows and linux.

rashmichandrashekar: I am not sure what the impact would be if we removed this annotation key from the aks-rp chart. At the same time if it has the 1.10.0.1 version it would create confusion. So, I updated it to the new one. We can update this to the version based on what we use for windows later.

ganga1980 (approved):

PR #698, opened 2022-01-21
Jan agent tasks

gracewehner:
- Upgrade telegraf and re-enable /opt path for trivy scan (1.21.x has a bug with the inputs.disk plugin, so using 1.20.3, but vulnerability scan is clear)
- Ignore new disk path that comes from containerd starting with k8s version >= 1.19.x, which was adding unnecessary InsightsMetrics logs and increasing cost
- Upgrade windows daemonset CPU request and limit to 500mc to fix issues with continuous restarts because of fluentd/fluent-bit not starting up in time

(This is pending AKS approval for windows CPU increase ~~and me currently testing the new telegraf bits on windows~~)

### DIFF
--- a/.github/workflows/pr-checker.yml
+++ b/.github/workflows/pr-checker.yml
@@ -1,5 +1,8 @@
 name: pullrequest-build-and-scan
 on:
+  push:
+    branches:
+      - grace/upgrade-telegraf
   pull_request:
     types: [opened, synchronize, reopened]
     branches:
@@ -56,7 +59,7 @@ jobs:
           format: 'table'
           severity: 'CRITICAL,HIGH'
           vuln-type: 'os,library'
-          skip-dirs: '/opt,/usr/sbin'
+          skip-dirs: ''
           exit-code: '1'
           timeout: '5m0s'
   WINDOWS-build:
--- a/.github/workflows/pr-checker.yml
+++ b/.github/workflows/pr-checker.yml
@@ -1,8 +1,5 @@
 name: pullrequest-build-and-scan
 on:
-  push:
-    branches:
-      - grace/upgrade-telegraf
   pull_request:
     types: [opened, synchronize, reopened]
     branches:
--- a/.github/workflows/pr-checker.yml
+++ b/.github/workflows/pr-checker.yml
@@ -56,7 +56,7 @@ jobs:
           format: 'table'
           severity: 'CRITICAL,HIGH'
           vuln-type: 'os,library'
-          skip-dirs: 'usr/sbin'
+          skip-dirs: '/usr/sbin'
           exit-code: '1'
           timeout: '5m0s'
   WINDOWS-build:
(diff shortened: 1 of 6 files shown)


vishiy (approved):

.github/workflows/pr-checker.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           format: 'table'
           severity: 'CRITICAL,HIGH'
           vuln-type: 'os,library'
-          skip-dirs: '/opt,/usr/sbin'
+          skip-dirs: 'usr/sbin'
```
ganga1980, line 59: isnt that be /usr/sbin?

kubernetes/linux/setup.sh
```diff
@@ -30,10 +30,10 @@ sudo apt-get install jq=1.5+dfsg-2 -y
 #used to setcaps for ruby process to read /proc/env
 sudo apt-get install libcap2-bin -y
 
-wget https://dl.influxdata.com/telegraf/releases/telegraf-1.18.0_linux_amd64.tar.gz
-tar -zxvf telegraf-1.18.0_linux_amd64.tar.gz
+wget https://dl.influxdata.com/telegraf/releases/telegraf-1.20.3_linux_amd64.tar.gz
```
ganga1980, line 33: can we purge this tar.gz file after untar and extracting this so that we can optimized the image size?

kubernetes/windows/setup.ps1
```diff
@@ -31,7 +31,7 @@ Write-Host ('Finished Installing Fluentbit')
 
 Write-Host ('Installing Telegraf');
 try {
-    $telegrafUri='https://dl.influxdata.com/telegraf/releases/telegraf-1.18.0_windows_amd64.zip'
+    $telegrafUri='https://dl.influxdata.com/telegraf/releases/telegraf-1.20.3_windows_amd64.zip'
```
ganga1980, line 34: same here. i think, we just should cleanup this zip file also.

.github/workflows/pr-checker.yml
gracewehner: Yes, fixed

kubernetes/linux/setup.sh
gracewehner: Good point, added this

kubernetes/windows/setup.ps1
gracewehner: This gets added to /installation
```
Invoke-WebRequest -Uri $telegrafUri -OutFile /installation/telegraf.zip
Expand-Archive -Path /installation/telegraf.zip -Destination /installation/telegraf
```
and then that directory is removed later with `Remove-Item /installation -Recurse` so I think for windows it's ok. I double-checked and didn't see it in the container files

ganga1980 (approved):

PR #699, opened 2022-01-21
default to port 10250 & containerd for linux agent

ganga1980:
default to port 10250 & containerd for Linux agent since these are common default across AKS and non-AKS .

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -465,7 +465,7 @@ if [ ! -z "$podWithValidContainerId" ]; then
       # convert to lower case so that everywhere else can be used in lowercase
       containerRuntime=$(echo $containerRuntime | tr "[:upper:]" "[:lower:]")
       nodeName=$(echo $nodeName | tr "[:upper:]" "[:lower:]")
-      # update runtime only if its not empty, not null and not startswith docker
+      # use default container runtime if obtained runtime value is either empty or null
       if [ -z "$containerRuntime" -o "$containerRuntime" == null  ]; then
             echo "using default container runtime as $CONTAINER_RUNTIME since got containeRuntime as empty or null"
       else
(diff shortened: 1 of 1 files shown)


kubernetes/linux/main.sh
```diff
@@ -463,7 +468,7 @@ if [ ! -z "$podWithValidContainerId" ]; then
       # update runtime only if its not empty, not null and not startswith docker
```
vishiy, line 468: this comment is not valid anymore.

kubernetes/linux/main.sh
```diff
@@ -463,7 +468,7 @@ if [ ! -z "$podWithValidContainerId" ]; then
...
       if [ -z "$containerRuntime" -o "$containerRuntime" == null  ]; then
             echo "using default container runtime as $CONTAINER_RUNTIME since got containeRuntime as empty or null"
-      elif [[ $containerRuntime != docker* ]]; then
+      else
             export CONTAINER_RUNTIME=$containerRuntime
```
vishiy, line 472: what will happen if the runtime is neither docker or containerd for a hybrid cluster ?

ganga1980: > what will happen if the runtime is neither docker or containerd for a hybrid cluster ?
> container runtime will be whatever detected for example CRIO. For switching of parser, the existing logic continue to work, i.e. if the container runtime docker is use docker parse else for all the runtimes, it will be "CRI" .

kubernetes/linux/main.sh
ganga1980: > this comment is not valid anymore.

Thanks for catching this. Addressed this.

kubernetes/linux/main.sh
vishiy: ok. so the only place we use runtime string is to choose the right parser for the logs? As long as we have telemetry (which i think we have) on how many clusters we have running with neither docker nor containerd, we are ok i guess.

ganga1980: > ok. so the only place we use runtime string is to choose the right parser for the logs? As long as we have telemetry (which i think we have) on how many clusters we have running with neither docker nor containerd, we are ok i guess.
>  Yes, for parser detection is "docker" vs non-docker and detected runtime used for the telemetry, and there are no changes to either of  these in this PR. Yes,  I think, we are good on this.

vishiy (approved):

PR #701, opened 2022-01-25
Gangams/jan 2022 release tasks 2

ganga1980:
This PR has following changes
-  write the AI ruby sdk logs to file instead of STDOUT for both windows and linux agent. this would help private link customers not getting the AI telemetry specific errors getting ingested when kube-system enabled.


build/linux/installer/conf/telegraf-prom-side-car.conf
```diff
@@ -166,3 +186,20 @@ $AZMON_TELEGRAF_CUSTOM_PROM_PLUGINS_WITH_NAMESPACE_FILTER
 
 ## OSM Prometheus configuration
 $AZMON_TELEGRAF_OSM_PROM_PLUGINS
+
+[[inputs.procstat]]
+ #name_prefix="t.azm.ms/"
+ exe = "mdsd"
+ interval = "60s"
```
gracewehner, line 193: For the metrics agent we are collecting at a higher frequency and then using the aggregator plugin to get the p50 and p95 usage every 5 minutes: 

```
[[aggregators.quantile]]
  period = "5m"
  drop_original = true
  quantiles = [0.50,0.95]
  algorithm = "t-digest"
  compression = 100.0
  namepass = ["agent_telemetry"]
```

Just a suggestion though if we want to reduce telemetry volume; otherwise we can keep this as is

ganga1980: > For the metrics agent we are collecting at a higher frequency and then using the aggregator plugin to get the p50 and p95 usage every 5 minutes:
> 
> ```
> [[aggregators.quantile]]
>   period = "5m"
>   drop_original = true
>   quantiles = [0.50,0.95]
>   algorithm = "t-digest"
>   compression = 100.0
>   namepass = ["agent_telemetry"]
> ```
> 
> Just a suggestion though if we want to reduce telemetry volume; otherwise we can keep this as is

changed to 30m aggregation and only P95 based on discussion with Vishwa.

gracewehner (approved):

gracewehner (approved):
LGTM

PR #702, opened 2022-01-26
Gangams/jan 2022 release tasks 3

ganga1980:
This PR has following changes
  - Add explicit exit code 1 across all the PS scripts
  - Add telemetry for nodeCount, containerCount
  - Add below telemetry to gather if any of the existing windows customers using 64KB field size for the variable fields - EnvVar, Command, Ports and Tags field
  -  !windows check in in_cadvisor_perf  when AADMSI enabled to avoid making call to mdsd socket

### DIFF
--- a/source/plugins/ruby/in_kube_podinventory.rb
+++ b/source/plugins/ruby/in_kube_podinventory.rb
@@ -288,16 +288,16 @@ module Fluent::Plugin
                   containerInventoryStream.add(emitTime, cirecord) if cirecord
                   ciRecordSize = cirecord.to_s.length
                   @winContainerInventoryTotalSizeBytes += ciRecordSize
-                  if ciRecordSize >= 65536 # 64KB
+                  if ciRecordSize >= Constants::MAX_RECORD_OR_FIELD_SIZE_FOR_TELEMETRY
                     @winContainerCountWithInventoryRecordSize64KBOrMore += 1
                   end
-                  if !cirecord["EnvironmentVar"].nil? && !cirecord["EnvironmentVar"].empty? && cirecord["EnvironmentVar"].length >= 65536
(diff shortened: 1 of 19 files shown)


source/plugins/ruby/in_kube_podinventory.rb
```diff
@@ -258,13 +286,27 @@ def parse_and_emit_records(podInventory, serviceRecords, continuationToken, batc
...
+                  if ciRecordSize >= 65536 # 64KB
+                    @winContainerCountWithInventoryRecordSize64KBOrMore += 1
+                  end
+                  if !cirecord["EnvironmentVar"].nil? && !cirecord["EnvironmentVar"].empty? && cirecord["EnvironmentVar"].length >= 65536
+                    @winContainerCountWithEnvVarSize64KBOrMore += 1
+                  end
+                  if !cirecord["Ports"].nil? && !cirecord["Ports"].empty? && cirecord["Ports"].length >= 65536
+                    @winContainerCountWithPortsSize64KBOrMore += 1
+                  end
+                  if !cirecord["Command"].nil? && !cirecord["Command"].empty? && cirecord["Command"].length >= 65536
```
gracewehner, line 300: Maybe the 65536 should be changed to be (64 * 1024) or made a constant just to be more clear about where the number is coming from

gracewehner (approved):

source/plugins/ruby/in_kube_podinventory.rb
ganga1980: > Maybe the 65536 should be changed to be (64 * 1024) or made a constant just to be more clear about where the number is coming from

> fixed pr feedback.

gracewehner (approved):

PR #706, opened 2022-01-28
Fixing telegraf bug for placeholder name

rashmichandrashekar:


kubernetes/windows/main.ps1
```diff
@@ -490,6 +496,11 @@ function Start-Telegraf {
         Write-Host "Failed to set environment variable NODE_IP for target 'machine' since it is either null or empty"
     }
 
+    $hostName = [System.Environment]::GetEnvironmentVariable("HOSTNAME", "process")
```
ganga1980, line 499: how are we getting value of HOSTNAME envvar? is it being set in the yaml?
similar to linux, I think, we should get from /pods api call.

rashmichandrashekar: its from the env, we are doing it this way for windows at all the other places. Keeping it consistent.

rashmichandrashekar: Also, another thing i noticed was we dont send the hostname in tags when we scrape metrics from the replicaset. This is the bug we actually need to fix - to remove it from windows. I think we should remove this section for windows too, from the comment looks like it was added for telemetry purposes which we are not using.

ganga1980: > its from the env, we are doing it this way for windows at all the other places. Keeping it consistent.

Ok, got it. thanks.

ganga1980 (approved):

PR #710, opened 2022-02-02
Gangams/chart version update for jan release

ganga1980:
This PR has following changes
 - chart version update
 - adding missing agentVersion annotation to keep consistent with aks agent 
 -  add the FBIT env vars for windows agent ds

### DIFF
--- a/charts/azuremonitor-containers/Chart.yaml
+++ b/charts/azuremonitor-containers/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 7.0.0-1
 description: Helm chart for deploying Azure Monitor container monitoring agent in Kubernetes
 name: azuremonitor-containers
-version: 2.9.0
+version: 2.9.1
 kubeVersion: "^1.10.0-0"
 keywords:
   - monitoring
--- a/charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
+++ b/charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
@@ -21,7 +21,7 @@ spec:
    labels:
     dsName: "omsagent-ds"
    annotations:
-    agentVersion: {{ .Values.omsagent.image.tagWindows }}
+    agentVersion: {{ .Values.omsagent.winAgentVersion }}
     dockerProviderVersion: {{ .Values.omsagent.image.dockerProviderVersion }}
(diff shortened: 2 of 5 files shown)


vishiy:
can we make sure we have 1 PR for 1 release and hence 1 that gets merged to prod per release? This was one of the problems since the later half of last year, and its almost impossible to track trails of our release PRs for future investigations.

ganga1980:
> can we make sure we have 1 PR for 1 release and hence 1 that gets merged to prod per release? This was one of the problems since the later half of last year, and its almost impossible to track trails of our release PRs for future investigations.
> sure, will ensure that going forward.

charts/azuremonitor-containers/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 7.0.0-1
 description: Helm chart for deploying Azure Monitor container monitoring agent in Kubernetes
 name: azuremonitor-containers
-version: 2.9.0
+version: 2.9.1
```
vishiy, line 5: is this the only place to up the chart version? I saw several other places like deployment scripts ... ?

ganga1980: > is this the only place to up the chart version? I saw several other places like deployment scripts ... ?

> Yes, this is only place.

charts/azuremonitor-containers/templates/omsagent-daemonset-windows.yaml
```diff
@@ -53,6 +53,12 @@ spec:
        resources:
 {{ toYaml .Values.omsagent.resources.daemonsetwindows | indent 9 }}
        env:
+       - name: FBIT_SERVICE_FLUSH_INTERVAL
```
vishiy, line 56: this was missed in last release ? or is it that we never released for long time since this was added on AKS side

ganga1980: > this was missed in last release ? or is it that we never released for long time since this was added on AKS side

> This is for windows agent and which targeted for this release with FBIT upgrade.

vishiy: ok got it.

vishiy (approved):

PR #714, opened 2022-02-11
adx bug + misc

vishiy:
* Fix ADX Bug with database name
* Fix jwt CVE
* Clean up un-used dependencies
* This will also fix ADX data breach issue that was reported when doing multi cluster


source/plugins/go/src/oms.go
```diff
@@ -1713,7 +1713,7 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 		// Try to read the ADX database name from environment variables. Default to DefaultAdsDatabaseName if not set.
 		// This SHOULD be set by tomlparser.rb so it's a highly unexpected event if it isn't.
 		// It should be set by the logic in tomlparser.rb EVEN if ADX logging isn't enabled
-		AdxDatabaseName := strings.TrimSpace(os.Getenv("AZMON_ADX_DATABASE_NAME"))
```
vishiy, line 1716: @asjddd - This should have never worked as the database name wasn't initialized (with & without overrides) that was intriduced by your previous PR. Are you not using the last released bits ? If so, this should be broken for you now in any of your clusters that is ingesting to ADX.

gracewehner (approved):

rashmichandrashekar (approved):

asjddd (approved):

source/plugins/go/src/oms.go
asjddd: We never caught this because we aren't using this feature yet. Dodged a bullet I guess.

PR #717, opened 2022-03-01
Gangams/http proxy support

ganga1980:
This PR has following changes 1) proxy support for AKS 2) MDSD update which has proxy support changes  to address OOM kill related issues.


charts/azuremonitor-containers/templates/omsagent-secret.yaml
```diff
@@ -13,19 +13,14 @@ data:
...
+  {{- else if and (.Values.Azure.proxySettings.isProxyEnabled) (.Values.Azure.proxySettings.httpProxy) }}
   PROXY: {{ .Values.Azure.proxySettings.httpProxy | b64enc | quote }}
-  {{- else if and (and (.Values.Azure.proxySettings.isProxyEnabled) (.Values.Azure.proxySettings.httpProxy)) (empty $httpProxyDict.userinfo) }} 
-  # adding arbitrary creds since omsagent expects arbitrary creds in case of no auth
-  PROXY: {{ urlJoin (dict "scheme" $httpProxyDict.scheme "userinfo" "admin:secret"  "host"  $httpProxyDict.host) | b64enc | quote }}
   {{- else if ne .Values.omsagent.proxy "<your_proxy_config>" }}
   PROXY: {{ .Values.omsagent.proxy | b64enc | quote }}
   {{- end }}
+  {{- if and (.Values.Azure.proxySettings.isProxyEnabled) (.Values.Azure.proxySettings.proxyCert) }}
+  PROXYCERT.crt: {{.Values.Azure.proxySettings.proxyCert | b64enc | quote}}
```
vishiy, line 24: why would we ignore noproxy ?

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
```diff
@@ -179,6 +185,12 @@ spec:
          - mountPath: /etc/omsagent-secret
            name: omsagent-secret
            readOnly: true
+         {{- if and (.Values.Azure.proxySettings.isProxyEnabled) (.Values.Azure.proxySettings.proxyCert) }}
+         - mountPath: /etc/ssl/certs/proxy-cert.crt
```
vishiy, line 189: ca cert can be changed (as part of cert rolling) post cluster creation in AKS, after which all our pods should be manually restarted to pick up the new cert . I see this in Ace's PR in AKS RP as well, which needs to be updated as well. https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/5449520 -- Please include the team in PRs.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
```diff
@@ -113,6 +113,12 @@ spec:
         - mountPath: /etc/omsagent-secret
           name: omsagent-secret
           readOnly: true
+        {{- if and (.Values.Azure.proxySettings.isProxyEnabled) (.Values.Azure.proxySettings.proxyCert) }}
+        - mountPath: /etc/ssl/certs/proxy-cert.crt
```
vishiy, line 117: please add telemetry to track how many clusters are providing ca cert.

kubernetes/linux/main.sh
```diff
@@ -190,8 +190,8 @@ if [ -e "/etc/omsagent-secret/WSID" ]; then
             # extract the port
             port="$(echo $hostport | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
 
-            if [ -z "$user" -o -z "$pwd" -o -z "$host" -o -z "$port" ]; then
-               echo "-e error proxy endpoint should be in this format http(s)://<user>:<pwd>@<hostOrIP>:<port>"
+            if [ -z "$host" -o -z "$port" ]; then
+               echo "-e error proxy endpoint should be in this format http(s)://<hostOrIP>:<port> or http(s)://<user>:<pwd>@<hostOrIP>:<port>"
```
vishiy, line 194: where do we check if the ca cert given is indeed a valid cert object ?

ganga1980: > where do we check if the ca cert given is indeed a valid cert object ?

> the cert validation should be taken care of the AKS RP since passed via aks.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
ganga1980: > please add telemetry to track how many clusters are providing ca cert.

Yep, good idea to have telemetry for the cert. let me add it.

kubernetes/linux/main.sh
vishiy: for non-aks ?

charts/azuremonitor-containers/templates/omsagent-secret.yaml
vishiy: i see that we are ignoring. My question was why are we ignoring .

ganga1980: > i see that we are ignoring. My question was why are we ignoring .
Because we dont use it. We only need HTTPS or HTTP proxy to use for outbound external endpoints.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
ganga1980: > ca cert can be changed (as part of cert rolling) post cluster creation in AKS, after which all our pods should be manually restarted to pick up the new cert . I see this in Ace's PR in AKS RP as well, which needs to be updated as well. https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/5449520 -- Please include the team in PRs.
> This will be taken care of the  AKS cert update workflow  for proxy CA cert (az aks update -> new node pool -> delete old node pool).

kubernetes/linux/main.sh
ganga1980: > for non-aks ?

it will be taken care of the Arc k8s platform since these being based by the RPs.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
vishiy: Also looking thru ubuntu man pages, i see the recommended way to install custom ca certs seems using the tool 'update-ca-certs' which will also update the index list of certs in /etc/ssl/certs/ca-certificates.crt, and we drop the cert in  /usr/share/ca-certificates .

charts/azuremonitor-containers/templates/omsagent-secret.yaml
vishiy: so if we want to scrape an ip address inside cluster, that should not go thru outbound proxy, how do we do it ?

kubernetes/linux/main.sh
vishiy: i hope they are getting validated. Because i see we just mount this into /etc/ssl which has global (container wide) impact. They could be some scripts that can harm us as well.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
ganga1980: > Also looking thru ubuntu man pages, i see the recommended way to install custom ca certs seems using the tool 'update-ca-certs' which will also update the index list of certs in /etc/ssl/certs/ca-certificates.crt, and we drop the cert in /usr/share/ca-certificates .

update-ca-certs after drop /usr/share/ca-certificates is other option instead of mounting the cert to /etc/ssl/certs/proxy-cert.crt.  since we are mounting the cert to /etc/ssl/certs/proxy-cert.crt so we dont need to run the update-ca-certs.

vishiy: why did we choose this option then ? Looks to me update-ca-certs also looks to ensure its a cert before adding to its index of certs...

kubernetes/linux/main.sh
ganga1980: > i hope they are getting validated. Because i see we just mount this into /etc/ssl which has global (container wide) impact. They could be some scripts that can harm us as well.

> yes. these be are being validated at the RP since  these are being used for other cluster components also.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
ganga1980: > why did we choose this option then ? Looks to me update-ca-certs also looks to ensure its a cert before adding to its index of certs...

>  As I I said, either options fine. I picked this option after talking folks who added in the Arc k8s, they opted this option to to work with distroless.

charts/azuremonitor-containers/templates/omsagent-secret.yaml
ganga1980: > so if we want to scrape an ip address inside cluster, that should not go thru outbound proxy, how do we do it ?

> Proxy configured for only external endpoints and everything else doesn't go through outbound proxy for example scrape IP address inside cluster.

vishiy: so with our current proxy approach, anything scraped outside the cluster will be broken, as there is no system wide proxy. so this will be broken anyways, and ssl is no first concern here.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
vishiy: i checked , and the only validation happening is base64 encoding. So you can wrap anything as base64 string and get that landed in /etc/ssl/proxycert.crt

kubernetes/linux/main.sh
vishiy: i checked , and the only validation happening is base64 encoding. So you can wrap anything as base64 string and get that landed in /etc/ssl/proxycert.crt

pfrcks (approved):

PR #734, opened 2022-03-23
PodReadyPercentage metric bug fix

gracewehner:
Fixes the issue where integer division was causing the percentage to either be 0 or 100.

I checked that this is not happening for the other percentage metrics. These are all calculated in the `filter_cadvisor2mdm.rb` file and use the format `value * 100 / total` where either `value` or `total` is a double or float.

I tested this with linux and windows controllers that have 0%, 50%, 66.67%, and 100% of pods ready.
I did not test with an ARM64 image. Please let me know if I should. I would need to whitelist my subscription to do so.


vishiy:
Thanks Grace.

vishiy (approved):

rashmichandrashekar (approved):

PR #740, opened 2022-04-06
Add CI Windows Build to MultiArch Dev pipeline

pfrcks:


.pipelines/azure_pipeline_dev.yaml
```diff
@@ -14,112 +14,194 @@ pr:
...
+  - task: PublishBuildArtifacts@1
+    inputs:
+      pathToPublish: '$(Build.ArtifactStagingDirectory)'
+      artifactName: drop
+
+- job: build_windows
+  dependsOn: common
+  pool:
+    name: Azure-Pipelines-Windows-CI-Test-EO
+  variables:
```
sarahpeiffer, line 148: The string "win-$(datetime)-$(commit)" is used several times (especially now that we are generating the metadata file separately. might be worth it to set it to a tag variable to make it easier to change in the future

.pipelines/azure_pipeline_dev.yaml
```diff
@@ -14,112 +14,194 @@ pr:
...
+    displayName: 'build base'
+
+  - task: AzureCLI@2
+    displayName: "Docker windows build"
+    inputs:
+      azureSubscription: ${{ variables.armServiceConnectionName }}
+      scriptType: ps
+      scriptLocation: inlineScript
+      inlineScript: |
+        # todo add metadata
```
sarahpeiffer, line 173: this is no longer todo

PR #745, opened 2022-04-21
force run trivy stage

pfrcks:


kubernetes/linux/Dockerfile.multiarch
```diff
@@ -43,4 +43,10 @@ RUN trivy rootfs --exit-code 1 --ignore-unfixed --no-progress --severity HIGH,CR
```
vishiy, line 43: This wont catch the OS/system CVEs - this will catch only lib & app CVEs. Will your change apply to this PR as well , so we can see what this PR is catching with your changes ?

pfrcks: i thought using rootfs scans root filesystem thereby scanning OS/system CVEs? (https://aquasecurity.github.io/trivy/v0.22.0/vulnerability/scanning/rootfs/)

Yes, this change will apply to this PR itself. The ContainerInsights-MultiArch-Dev check will fail

vishiy: can we take a previous image and scan both ways to see the results ? I am sure they will be different (both ways).

pfrcks: 1. Results scanning docker image mcr.microsoft.com/azuremonitor/containerinsights/ciprod:ciprod03172022

image

2. Results scanning filesystem using rootfs

image

vishiy: Thanks. That looks better. We should scan for MEDIUM as well. As customers report on Medium as well (though it will be best effort on our side). If you can change in ADO checks, i will change in other places to ionclude medium.

pfrcks: we do --exit-code 1 with trivy to fail the build in case we find any vulnerabilities. If we include medium as well the builds will not pass for medium vulnerabilities. is that what we want?

pfrcks: @vishiy can you take a look now? I have added the new checks and a trivyignore file. Currently it contains CVEs being reported in /opt/telegraf binary (except the newer go crypto module one)

kubernetes/linux/Dockerfile.multiarch
```diff
@@ -39,8 +39,17 @@ RUN chmod 775 $tmpdir/*.sh; sync; $tmpdir/setup.sh ${TARGETARCH}
 # Do vulnerability scan in a seperate stage to avoid adding layer
 FROM base_image AS vulnscan
 COPY --from=aquasec/trivy:latest /usr/local/bin/trivy /usr/local/bin/trivy
-RUN trivy rootfs --exit-code 1 --ignore-unfixed --no-progress --severity HIGH,CRITICAL --skip-files "/usr/sbin/telegraf" --skip-files "/opt/telegraf" --skip-files "/usr/local/bin/trivy" /
+COPY .pipelines/.trivyignore .trivyignore
+RUN trivy rootfs --ignore-unfixed --no-progress --severity HIGH,CRITICAL,MEDIUM --skip-files "/usr/local/bin/trivy" /
+RUN trivy rootfs --ignore-unfixed --no-progress --severity HIGH,CRITICAL,MEDIUM /usr/lib
+RUN trivy rootfs --exit-code 1 --ignore-unfixed --no-progress --severity HIGH,CRITICAL,MEDIUM --skip-files "/usr/local/bin/trivy" / > /dev/null 2>&1 && trivy rootfs --exit-code 1 --ignore-unfixed --no-progress --severity HIGH,CRITICAL,MEDIUM /usr/lib > /dev/null 2>&1
```
vishiy, line 45: why do we need to do 4 times now (instead of 2) ?

vishiy: oh first 2 times just to show output and the next 2 to bail out ?

pfrcks: unfortunately, yes :(

vishiy: ok thanks.

vishiy (approved):

PR #747, opened 2022-04-26
update go-autorest to address vuln

pfrcks:
image
used go mod graph to see that crypto package is a dependency of go-autorest.

updated it to address the vuln.


source/plugins/go/src/go.mod
```diff
@@ -4,14 +4,17 @@ go 1.14
 
 require (
 	github.com/Azure/azure-kusto-go v0.5.2
+	github.com/Azure/go-autorest/autorest v0.11.27 // indirect
```
vishiy, line 7: hi amol, these are brought in by ADX sdk i believe, so we should update that rather than these individually, as we dont know if this will work.

pfrcks: is the ADX sdk you are referring to the one above? (azure-kusto-go) ?

PR #750, opened 2022-04-28
add cifs & fuse file systems to ignore list

vishiy:


andyzhangx (approved):

andyzhangx:
so we don't want to fix it in omsagent? and what if customer hit such stateFS call issue? The only workaround I could think of is disabling the omsagent

gracewehner (approved):

PR #751, opened 2022-04-28
CA Cert Fix for Mariner Hosts in Air Gap

gracewehner:
This is the script changes. The AKS RP changes for mounting /anchors/mariner and /anchors/ubuntu will need to go in with this image change. I will make a PR for this when we release.

Issue is Mariner host has CA certs in a different location that we still need to mount on the container. Other teams saw issues when just mounting directly for both Ubuntu and Mariner, so solution is to mount into a different location and copy over in the main.sh script.

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -296,16 +296,19 @@ echo "export CLOUD_ENVIRONMENT=$CLOUD_ENVIRONMENT" >> ~/.bashrc
 # OS_ID here is the container distro.
 # Adding Mariner now even though the elif will never currently evaluate. 
 if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
-  OS_ID=$(cat /etc/os-release | grep ^ID= | cut -d '=' -f2)
-  if [ $OS_ID == "ubuntu" ]; then
+  OS_ID=$(cat /etc/os-release | grep ^ID= | cut -d '=' -f2 | tr -d '"' | tr -d "'")
+  if [ $OS_ID == "mariner" ]; then
+    cp /anchors/ubuntu/* /etc/pki/ca-trust/source/anchors
+    cp /anchors/mariner/* /etc/pki/ca-trust/source/anchors
+    update-ca-trust
+  else
+    if [ $OS_ID != "ubuntu" ]; then
+      echo "Error: The ID in /etc/os-release is not ubuntu or mariner. Defaulting to ubuntu."
+    fi
     cp /anchors/ubuntu/* /usr/local/share/ca-certificates/
     cp /anchors/mariner/* /usr/local/share/ca-certificates/
     update-ca-certificates
     cp /etc/ssl/certs/ca-certificates.crt /usr/lib/ssl/cert.pem
-  elif [ $OS_ID == "mariner" ]; then
-    cp /anchors/ubuntu/* /etc/pki/ca-trust/source/anchors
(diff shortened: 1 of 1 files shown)


kubernetes/linux/main.sh
```diff
@@ -288,6 +288,27 @@ fi
...
+# so both /anchors/ubuntu and /anchors/mariner are mounted in the yaml.
+# One will have the certs and the other will be empty.
+# These need to be copied to a different location for Mariner vs Ubuntu containers.
+# OS_ID here is the container distro.
+# Adding Mariner now even though the elif will never currently evaluate. 
+if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
+  OS_ID=$(cat /etc/os-release | grep ^ID= | cut -d '=' -f2)
+  if [ $OS_ID == "ubuntu" ]; then
+    cp /anchors/ubuntu/* /usr/local/share/ca-certificates/
+    cp /anchors/mariner/* /usr/local/share/ca-certificates/
```
pfrcks, line 302: if this an ubuntu system then /anchors/mariner/ will be empty? If so, why this extra copy.  vice-versa in the else clause

gracewehner: The OS_ID is for the container distro but the /anchors/ubuntu and anchors/mariner are coming from the node which can have a different distro than the container, such as a Mariner host and our Ubuntu container. We do not know the system of the node so we need to mount both and one directory will always be empty. I could add an if statement to only copy over the directory that isn't empty, but have checked trying to copy both doesn't cause issues

The elif is for when we move over to use mariner as our container distro. When this happens, there can be an Ubuntu host for our Mariner container, and the certs need to be copied over into a different directory than for an ubuntu container

rashmichandrashekar (approved):

kubernetes/linux/main.sh
```diff
@@ -288,6 +288,27 @@ fi
...
 
+# Copying over CA certs for airgapped clouds. This is needed for Mariner vs Ubuntu hosts.
+# We are unable to tell if the host is Mariner or Ubuntu,
+# so both /anchors/ubuntu and /anchors/mariner are mounted in the yaml.
+# One will have the certs and the other will be empty.
+# These need to be copied to a different location for Mariner vs Ubuntu containers.
+# OS_ID here is the container distro.
+# Adding Mariner now even though the elif will never currently evaluate. 
+if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
+  OS_ID=$(cat /etc/os-release | grep ^ID= | cut -d '=' -f2)
```
ganga1980, line 299: As per manpage - https://www.freedesktop.org/software/systemd/man/os-release.html.  Format of the ID kv pair can have quotes and also it seems optional.  Suggestions 1. probable we should have default behavior (ubuntu) in such cases or echo error message if the OS_ID either "ubuntu" or "mariner" which  will help us debug from our telemetry if we get into such situation. 
ID=
A lower-case string (no spaces or other characters outside of 0–9, a–z, ".", "_" and "-") identifying the operating system, excluding any version information and suitable for processing by scripts or usage in generated filenames. **If not set, a default of "ID=linux" may be used**. Note that even though this string may not include characters that require shell quoting, quoting may nevertheless be used.

**Examples: "ID=fedora", "ID=debian".**

gracewehner: Thanks Ganga. Changed this to trim for quotes and to default to ubuntu and echo an error message if it's neither

rashmichandrashekar (approved):

PR #754, opened 2022-05-05
Fix ruby vulnerabilities

pfrcks:
- Updates ruby to 3.1.0
- Uses RVM to install ruby instead of brightbox ppa
- Updates fluentd gem version to 1.14..6
- Updates cool.io gem version to 1.7.1
- Move from source code to gem for tomlrb version 2.0.1 for Linux. Update gem version for Windows to 2.0.1


pfrcks:
Changes are live on the [cluster](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/692aea0b-2d89-4e7e-ae30-fffe40782ee2/resourceGroups/arm64-amol-test/providers/Microsoft.ContainerService/managedClusters/arm64-test-cluster2/overview)

kubernetes/linux/setup.sh
```diff
@@ -11,6 +13,20 @@ fi
...
+curl -sSL https://rvm.io/pkuczynski.asc | gpg --import -
+curl -sSL https://get.rvm.io | bash -s stable
+
+# setup paths for ruby and rvm
+if [ -f /etc/profile.d/rvm.sh ]; then
+    source /etc/profile.d/rvm.sh
+    echo "[ -f /etc/profile.d/rvm.sh ] && source /etc/profile.d/rvm.sh" >> ~/.bashrc
+fi
+
+rvm install 3.1.0
```
ganga1980, line 27: @pfrcks , In ruby 3.1, Yajl JSON gem is root cause for the data gaps and CPU throttling on scale cluster.  This issue easily reproducible with 2K pods on the cluster or running test ruby app (within container) parsing ~2MB JSON payload. Default JSON parser addresses this issue but we would need more testing as this JSON parser in critical path and doesn't cause any perf regressions. Lets spend more time on testing once your are back.  If all the tests pass, then we get this change in next agent release.

PR #762, opened 2022-05-17
Adding v2 schema options

austonli:
Adding commented out section in log collection settings for v2 schema

### DIFF
--- a/kubernetes/container-azm-ms-agentconfig.yaml
+++ b/kubernetes/container-azm-ms-agentconfig.yaml
@@ -43,9 +43,9 @@ data:
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
        #[log_collection_settings.schema]
-          # In the absense of this configmap, default value for containerlog_schema_version is "v1"
+          # In the absence of this configmap, default value for containerlog_schema_version is "v1"
           # Supported values for this setting are "v1","v2"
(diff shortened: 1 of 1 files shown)


kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -42,6 +42,12 @@ data:
           # When the setting is set to false, only the kube events with !normal event type will be collected
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
+       #[log_collection_settings.schema]
+          # In the absense of this configmap, default value for containerlog_schema_version is "v1"
+          # Supported values for this setting are "v1","v2"
+          # See documentation for benefits of v2 schema over v1 schema before opting for "v2" schema
```
ganga1980, line 48: probable, we should we add public doc  link here otherwise "see documentation" will be confusing, right?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -42,6 +42,12 @@ data:
           # When the setting is set to false, only the kube events with !normal event type will be collected
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
+       #[log_collection_settings.schema]
```
ganga1980, line 45: @austonli , we are working on the release. let me know if you want to take this change as part of the current agent release.

ganga1980 (approved):

PR #768, opened 2022-05-23
(AzureCXP) MicrosoftDocs/azure-docs#93132

KrishnaG-MSFT:


scripts/onboarding/aks/mdmonboarding/mdm_onboarding_atscale.sh
```diff
@@ -38,7 +38,7 @@ for cluster in $(echo $CLUSTERS_LIST | jq -c '.[]'); do
     export SP_ID=$(az aks show -g $clusterRG -n $clusterName --query servicePrincipalProfile.clientId -o tsv)
 
     if [ -z $SP_ID ]; then
-        export MSI_ID=$(az aks show -g $clusterRG -n $clusterName --query addonProfiles.omsagent.identity.clientId -o tsv)
+        export MSI_ID=$(az aks show -g $clusterRG -n $clusterName --query addonProfiles.omsAgent.identity.clientId -o tsv)
```
ganga1980, line 41: The name contract we  have is "omsagent"                                                                                       https://docs.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-enable-existing-clusters#create-and-execute-a-template 
but if customer has different case in their templates then we can ask the customer to fix it. With regards to this change, can you please convert entire addonProfiles to lowercase and get the the clientid so that this handles even if customer has different casing.

ganga1980, line 41: The name contract we  have is "omsagent"                                                                                       https://docs.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-enable-existing-clusters#create-and-execute-a-template 
but if customer has different case in their templates then we can ask the customer to fix it. With regards to this change, can you please convert entire addonProfiles to lowercase and get the the clientid so that this handles even if customer has different casing.

PR #770, opened 2022-05-25
trivy image scan in dev

pfrcks:


pfrcks:
@vishiy we will need to disable the pr build and scan github workflow from github itself as well

vishiy (approved):

PR #772, opened 2022-05-31
Make custom metrics endpoint configurable to be adapted to edge environments

MSFTXiangyu:
This changes make custom metrics endpoint configurable, so that extension connected to edge control plane service could emit metrics into monitoring services on edge. This PR containing the following changes:
- Custom metrics endpoint param was added in values.yaml, then be set into environment variables in omsagent yaml files.
- Read metrics endpoint settings from environment variable in out_mdm.rb.
- Make custom metrics check pass on edge env when ArcA (the edge environment product name) flag is set to true.

Ps: The two values will be set by environment admin, rather that end users, so it's not placed in "Azure" section of values.

### DIFF
--- a/source/plugins/ruby/CustomMetricsUtils.rb
+++ b/source/plugins/ruby/CustomMetricsUtils.rb
@@ -13,6 +13,11 @@ class CustomMetricsUtils
             if aks_region.to_s.empty? || aks_resource_id.to_s.empty?
                 return false # This will also take care of AKS-Engine Scenario. AKS_REGION/AKS_RESOURCE_ID is not set for AKS-Engine. Only ACS_RESOURCE_NAME is set
             end
+            # If this is cluster is connected to ArcA control plane and metrics endpoint provided, custom metrics shall be emitted.
+            is_arca_cluster = ENV['IS_ARCA_CLUSTER']
+            if is_arca_cluster.to_s.downcase == "true" && !ENV['CUSTOM_METRICS_ENDPOINT'].to_s.empty?
+                return true
+            end
 
             return aks_cloud_environment.to_s.downcase == 'azurepubliccloud'
         end
--- a/source/plugins/ruby/out_mdm.rb
+++ b/source/plugins/ruby/out_mdm.rb
@@ -26,7 +26,8 @@ module Fluent::Plugin
       @@token_resource_audience = "https://monitor.azure.com/"
       @@grant_type = "client_credentials"
       @@azure_json_path = "/etc/kubernetes/host/azure.json"
-      @@post_request_url_template = "https://%{aks_region}.monitoring.azure.com%{aks_resource_id}/metrics"
+      @@public_metrics_endpoint_template = "https://%{aks_region}.monitoring.azure.com"
+      @@post_request_url_template = "%{metrics_endpoint}%{aks_resource_id}/metrics"
       @@aad_token_url_template = "https://login.microsoftonline.com/%{tenant_id}/oauth2/token"
 
       # msiEndpoint is the well known endpoint for getting MSI authentications tokens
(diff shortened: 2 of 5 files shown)


source/plugins/ruby/CustomMetricsUtils.rb
```diff
@@ -13,6 +13,11 @@ def check_custom_metrics_availability
             if aks_region.to_s.empty? || aks_resource_id.to_s.empty?
                 return false # This will also take care of AKS-Engine Scenario. AKS_REGION/AKS_RESOURCE_ID is not set for AKS-Engine. Only ACS_RESOURCE_NAME is set
             end
+            # If this is cluster is connected to ArcA control plane and metrics endpoint provided, custom metrics shall be emitted.
+            is_ArcA_Cluster = ENV['IS_ARCA_CLUSTER']
+            if is_ArcA_Cluster.to_s.downcase == "true" && !ENV['Custom_Metrics_Endpoint'].to_s.empty?
```
ganga1980, line 18: In case of Azure, custom metrics are only available in public cloud. what regions and clouds (if there is any) supported in case of Arc-A?  is that assumption, as long as custom metrics endpoint provided custom metrics available in case of Arc-A?

source/plugins/ruby/out_mdm.rb
```diff
@@ -98,7 +99,20 @@ def start
           if aks_resource_id.downcase.include?("microsoft.kubernetes/connectedclusters")
             @isArcK8sCluster = true
           end
-          @@post_request_url = @@post_request_url_template % { aks_region: aks_region, aks_resource_id: aks_resource_id }
+
+          # If Custom_Metrics_Endpoint provided, the url format shall be validated before emitting metrics into given endpoint.
+          Custom_Metrics_Endpoint = ENV['Custom_Metrics_Endpoint']
```
ganga1980, line 104: change the variable naming convention to custom_metrics_endpoint to be consistent with other local variables.

source/plugins/ruby/out_mdm.rb
```diff
@@ -98,7 +99,20 @@ def start
...
+          if !Custom_Metrics_Endpoint.to_s.empty?
+            metrics_endpoint = Custom_Metrics_Endpoint.lstrip.rstrip
```
ganga1980, line 106: you can use "strip" to remove leading and trailing whitespaces, right?

source/plugins/ruby/out_mdm.rb
```diff
@@ -408,5 +422,14 @@ def isWindows()
       end
       return isWindows
     end
+
+    def valid_url?(string)
+      uri = URI.parse(string)
+      %w( http https ).include?(uri.scheme)
```
ganga1980, line 428: what this **%w( http https ).include?(uri.scheme)** do?

source/plugins/ruby/out_mdm.rb
```diff
@@ -408,5 +422,14 @@ def isWindows()
       end
       return isWindows
     end
+
+    def valid_url?(string)
+      uri = URI.parse(string)
```
ganga1980, line 427: URI.parse should take care of throwing an exception if the URI is not valid. why do we need this helper method which does same thing what URI.parse does?

source/plugins/ruby/CustomMetricsUtils.rb
MSFTXiangyu: 1. In Arc-A, the regions could be different since the cluster could placed on different areas. Also, the clouds (found from domain) is not usable for us, since the domain of Arc-A is a FQDN from user internal network. So currently we have to add a parameter IS_ARCA_CLUSTER to mark Arc-A cluster rather than clouds.
2. Yep, the assumption is right. Since the metrics endpoint is provided by cluster admin, and custom metrics is supported from Arc-A preview 1, it can be confirmed to provide custom metrics in this case.

source/plugins/ruby/out_mdm.rb
MSFTXiangyu: It's used to verify if this uri starts with http or https. Considering this endpoint is setup by ArcA admin, it not necessary here.

source/plugins/ruby/out_mdm.rb
MSFTXiangyu: Yep. URI.parse would raise [URI::InvalidURIError](https://apidock.com/ruby/URI/InvalidURIError) if it's not valid. Considering there's exception rescue in start function. Have replaced this helper method with URI.parse directly.

ganga1980 (approved):

PR #773, opened 2022-06-02
Gangams/fix telegraf issue

ganga1980:
For the token, IMDS HTTP call made during the out_mdm plugin startup and also in send_to_mdm for the token refresh. If the IMDS HTTP call blocked due to AAD POD identity standard mode (in which case, IMDS call intercepted by the AAD Pod identity hence the call failed with read timeout if its not whitelisted to AzurePodIdentityException) or the IMDS call failed with read timeout (default: 1min) or for any other reason. This IMDS call will be retried with 3 times with delay (1s & 2s).  

The issue here is the if the IMDS http API call blocked during the FLUENTD startup, TCP listener 25228 will not come up until all the all retries tried.  Because TCP listener 25228 is not ready, telegraf cant connect even with 30s.   The fix here is to avoid IMDS HTTP API call during the out_mdm plugin startup and only invoke the call in send_to_mdm which will be invoked when the plugin received the data.

- Verified telegraf connects to FLUENTD TCP port 25228 with AAD POD identity standard mode
-  Verified the metrics works as expected 

- Added telemetry dashboard to track the number of clusters which are failing to connect Fluentd TCP port 25228 by agent version so we can track with agent version which has fix.


gracewehner:
> Added telemetry to track the number of clusters which are failing to connect Fluentd TCP port 25228 by agent version so we can track with agent version which has fix.

Is this added just in our dashboards by querying the traces, so no code changes needed?

source/plugins/ruby/out_mdm.rb
```diff
@@ -118,12 +118,10 @@ def start
               @useMsi = true
               msi_endpoint = @@imds_msi_endpoint_template % { resource: @@token_resource_audience }
               @parsed_token_uri = URI.parse(msi_endpoint)
-              @cached_access_token = get_access_token
```
gracewehner, line 121: So get_access_token will still retry for a total of 6 minutes when the AAD pod identity is enabled without the exception, but this is ok since now it will only block the MDM metrics, which wouldn't be able to authenticate to send anyway?

ganga1980: > So get_access_token will still retry for a total of 6 minutes when the AAD pod identity is enabled without the exception, but this is ok since now it will only block the MDM metrics, which wouldn't be able to authenticate to send anyway?

>  This is not issue in case of the MSI auth mode (meaning with DCRs) for customers with AAD Pod identity in standard mode. But, the agent which is onboarded without MSI/DCR based auth, IMDS call failed if the customer (with AAD Pod identity in standard mode) not added the exception and this **only** impact the MDM metrics.

ganga1980:
> > Added telemetry to track the number of clusters which are failing to connect Fluentd TCP port 25228 by agent version so we can track with agent version which has fix.
> 
> Is this added just in our dashboards by querying the traces, so no code changes needed?

> This is dashboard to track the clusters (which are failed to connect fluentd port:25228) by agent version. updated the PR description to reflect this.

gracewehner (approved):

PR #774, opened 2022-06-09
add integration for azure subnet ip usage

vishiy:
* Add integration for collecting Subnets IP usage metrics for Azure CNI (turned OFF by default)
* You can see the data in this work book from a sample cluster in my sub - https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/8f6da2d9-ff10-4800-9239-c7e0e8b3407f/resourceGroups/vishwasip/providers/Microsoft.ContainerService/managedClusters/vishwasip/workbooks
* Also added a telemetry point for the integration (to get number of clusters that has this turned ON)
* We need to update our docs (i will create a separate PR after we roll this out) along with Azure network team's doc (they will be updating their docs after we roll out) for this

### DIFF
--- a/source/plugins/ruby/CAdvisorMetricsAPIClient.rb
+++ b/source/plugins/ruby/CAdvisorMetricsAPIClient.rb
@@ -36,6 +36,7 @@ class CAdvisorMetricsAPIClient
   @containerLogsRoute = ENV["AZMON_CONTAINER_LOGS_ROUTE"]
   @npmIntegrationBasic = ENV["TELEMETRY_NPM_INTEGRATION_METRICS_BASIC"]
   @npmIntegrationAdvanced = ENV["TELEMETRY_NPM_INTEGRATION_METRICS_ADVANCED"]
+  @subnetIpUsageMetrics = ENV["TELEMETRY_SUBNET_IP_USAGE_INTEGRATION_METRICS"]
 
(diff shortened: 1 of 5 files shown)


source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -282,6 +283,10 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
                     elsif (!@npmIntegrationBasic.nil? && !@npmIntegrationBasic.empty?)
                       telemetryProps["int-npm-b"] = "1"
                     end
+                    # telemetry for subnet ip usage integration
+                    if (!@subnetIpUsageMetrics.nil? && !@subnetIpUsageMetrics.empty?)
+                      telemetryProps["int-ipsubnetusage"] = "1"
```
rashmichandrashekar, line 288: Just curious, why is this "1" instead of "true"?

rashmichandrashekar (dismissed):

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
vishiy: saving telemetry 'bytes' for cost :)

rashmichandrashekar (approved):

PR #794, opened 2022-07-18
Address vulnerabilities through package updates

pfrcks:
- Updates to ruby 3.1.1
- Uses RVM as ruby manager instead of the brightbox ppa 
- Updates fluentd to 1.14.6
- Use default JSON gem instead of yajl-json
- Consume tomlrb as a gem instead of committed source code


.trivyignore
```diff
@@ -4,16 +4,12 @@
...
-CVE-2021-31799
-CVE-2021-28965
-
 #dpkg vulnerability in ubuntu
-CVE-2022-1304
\ No newline at end of file
+CVE-2022-1304
+
+# telegraf vulnerabilities
+CVE-2022-29190
```
ganga1980, line 11: These Vulns will be fixed with other PR just merged, right? so we can remove them from ignore file.

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -2,7 +2,7 @@
 # frozen_string_literal: true
 
 class CAdvisorMetricsAPIClient
-  require "yajl/json_gem"
+  require "json"
```
ganga1980, line 5: since you are removing  require "yajl". please double check we dont have any references of using "yajl"

kubernetes/windows/Dockerfile
```diff
@@ -11,28 +11,28 @@ ARG IMAGE_TAG=win-ciprod06142022
...
 && choco install -y vim
 
 # gangams - optional MSYS2 update via ridk failing in merged docker file so skipping that since we dont need optional update
 RUN refreshenv \
 && ridk install 3 \
 && echo gem: --no-document >> C:\ProgramData\gemrc \
-&& gem install cool.io -v 1.5.4 --platform ruby \
+&& gem install cool.io -v 1.7.1 --platform ruby \
 && gem install oj -v 3.3.10 \
 && gem install json -v 2.2.0 \
```
ganga1980, line 24: isnt that JSON will be default comes with ruby? if so,  then do we need this specific version of JSON gem?

ganga1980, line 24: isnt that JSON will be default comes with ruby? if so, then do we need this specific version of JSON gem?

.trivyignore
pfrcks: updated

kubernetes/windows/Dockerfile
pfrcks: removed

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
pfrcks: done

ganga1980 (approved):

PR #796, opened 2022-07-23
Gangams/fix log loss inode reuse

ganga1980:
This PR fixes the issue about container log loss issue by having Ignore_Older  value same as fluent-bit default . For any reason, customer wants to configure the previous default or any other value, customer can configure through configmap setting.

If there are no changes to the container log files for more than values of the Ignore_Older (i.e. 5m current value) and those container log files will be ignored to monitor.

If the file gets deleted (because of the container deletion), then delete notification of these unmonitored files will not be detected by the fluent bit since these container files marked as un-monitored by the fluent bit. 

Only monitored files will be checked whether those files deleted or not periodic  interval basis.  If the  container log file deleted for the monitored files,  then fluent bit removes the entry in the DB as well from the tail file list as these being checked periodically. 

Since the fluent bit doesn’t detect deletion of the un-monitored files then this causes the stale DB entries for the deleted files.  From Linux kernel standpoint, these files are deleted  from the filesystem which means that inode of these deleted files can be re-used for other files.

If Linux kernel uses the inode value of the  already deleted file, then Fluent bit uses wrong offset of the deleted file (since there is stale entry in the DB with matching inode) instead of offset=0 for newly created container log file. Because of this,  all the log lines until recorded offset in the DB will not be collected.

If the container log file has less than the recorded offset, then all the content in log file will be dropped.


pfrcks:
Since the default value now is now to read all specified files irrespective of the last modified time how does this impact both cpu and memory perf for fluent bit?

ganga1980:
> Since the default value now is now to read all specified files irrespective of the last modified time how does this impact both cpu and memory perf for fluent bit? Do we have any data showing the difference between having the value as 5 minutes and having the default as monitor all?
> There is no impact to CPU but little impact to memory at scale of 1K pods/container/containers logs on the node. This additional memory for tracking the file names. 1K is high scale which is not common to customer to use this scale unless customer has completed pods for the Jobs which we recommend them to have policy to clean up the completed jobs.
Memory
image
CPU
image
Note - the spike because the node has 1K containers.

pfrcks (approved):

PR #798, opened 2022-07-28
fix: Set the proper metric value for node allocatable metrics (HOTFIX)

bragi92:


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -517,7 +517,7 @@ def getContainerResourceRequestsAndLimits(pod, metricCategory, metricNameToColle
               #No container level limit for the given metric, so default to node level limit
             else
               if (metricCategory == "limits" && !nodeAllocatableRecord.nil? && !nodeAllocatableRecord.empty? && nodeAllocatableRecord.has_key?(metricNameToCollect))
-                metricValue = nodeAllocatableRecord[metricNameToCollect]
+                metricValue = getMetricNumericValue(metricNameToCollect, nodeAllocatableRecord[metricNameToCollect])
```
vishiy, line 520: kaveesh, can you please search for places where this function is called (getMetricNumericValue) to see if there are any more occurences ?

ganga1980, line 520: Thanks for investigating and making the fix. lets get the image tag update from "ciprod06272022" to "ciprod07282022" in omsagent.yaml and DockerFiles.

bragi92: Found one other location, added it there.

ganga1980 (approved):
lgtm

PR #805, opened 2022-08-05
User/amagraw/fix milli bytes bug

pfrcks:


pfrcks:
image

ganga1980 (approved):

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -751,8 +751,13 @@ def getMetricNumericValue(metricName, metricVal)
             metricValue.chomp!("k")
             metricValue = Float(metricValue) * 1000.0 ** 1
           elsif (metricValue.end_with?("m"))
+            #original value before downcase ending with M is megabyte and value ending with m is milli-byte
```
vishiy, line 754: Amol, change looks ok. But please log a future work-item to make this function all case-insensitive, so we can add support for future units easily.

vishiy (approved):

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -751,8 +751,13 @@ def getMetricNumericValue(metricName, metricVal)
...
             metricValue.chomp!("m")
-            metricValue = Float(metricValue) * 1000.0 ** 2
```
vishiy, line 755: also, can you check if Ux (live) supportes this unit ? if not we need to fix that as well.

pfrcks: sure

PR #807, opened 2022-08-10
CI Aug 2022 Release

pfrcks:
Refer to ReleaseNotes.md for detailed list of changes.


.trivyignore
```diff
@@ -1,27 +1,10 @@
...
-CVE-2021-42836
-
-# ruby in /usr/lib
-CVE-2020-36327
-CVE-2021-43809
-CVE-2021-41816
-CVE-2021-41819
-CVE-2021-31799
-CVE-2021-28965
+CVE-2022-29190
```
ganga1980, line 3: are all these telegraf vulnerabilities? isnt that we have 0  telegraf vulnerabilities because of the telegraf upgrade and purging the telegraf in mdsd?

pfrcks: these vulnerabilities are present in the packages consumed by telegraf that we use. Since telegraf does not directly use these packages, they have not fixed them and therefore they show up even in the latest version.

ganga1980 (approved):

PR #809, opened 2022-08-16
Rename resources from omsagent to ama-logs

pfrcks:


charts/azuremonitor-containers/values.yaml
```diff
@@ -29,7 +29,7 @@ omsagent:
     agentVersion: "azure-mdsd-1.17.0"
     winAgentVersion: "0.0.0-0" # there is no base agent version for windows agent
 
-  # The priority used by the omsagent priority class for the daemonset pods
+  # The priority used by the ama-logs priority class for the daemonset pods
```
jatakiajanvi12, line 32: Not a 100% sure about what this comment means, but should this be amalogs?

jatakiajanvi12 (commented):
Looks good overall. Do we need to add in description just to show all the insights/charts, etc. are looking as expected after this change?

kubernetes/ama-logs.yaml
```diff
@@ -1,13 +1,13 @@
 apiVersion: v1
```
ganga1980, line 1: Please get the rename updates for the addon-VPA-resizer.   Please marge latest dev and double check if there are no updates which requires rename updates.

pfrcks: done

charts/azuremonitor-containers/values.yaml
pfrcks: i think this is fine since it is referring to the actual pod names rather than the internal value names

pfrcks:
rebased with ci_dev and made rename changes in recent commits to ci_dev

PR #814, opened 2022-09-01
fix: Update DCR creation to Clusters resource group instead of workspace (ARM Template update)

bragi92:
Link to CLI PR: https://github.com/Azure/azure-cli/pull/23630.
Deepanshu is updating this in the UX with the prometheus update

### DIFF
--- a/scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
+++ b/scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
@@ -20,12 +20,6 @@
               "description": "Existing or new tags to use on AKS, ContainerInsights and DataCollectionRule Resources"
             }
         },
-        "workspaceLocation": {
-            "type": "string",
-            "metadata": {
-                "description": "Worksapce Location for data collection rule"
-            }
-        },
         "workspaceResourceId": {
             "type": "string",
             "metadata": {
@@ -38,19 +32,17 @@
         "clusterResourceGroup": "[split(parameters('aksResourceId'),'/')[4]]",
         "clusterName": "[split(parameters('aksResourceId'),'/')[8]]",
         "clusterLocation": "[replace(parameters('aksResourceLocation'),' ', '')]",
-        "workspaceSubscriptionId": "[split(parameters('workspaceResourceId'),'/')[2]]",
-        "workspaceResourceGroup": "[split(parameters('workspaceResourceId'),'/')[4]]",
         "dcrName": "[Concat('MSCI', '-', variables('clusterName'), '-', variables('clusterLocation'))]",
         "associationName":  "ContainerInsightsExtension",
-        "dataCollectionRuleId": "[resourceId(variables('workspaceSubscriptionId'), variables('workspaceResourceGroup'), 'Microsoft.Insights/dataCollectionRules', variables('dcrName'))]"
+        "dataCollectionRuleId": "[resourceId(variables('clusterSubscriptionId'), variables('clusterResourceGroup'), 'Microsoft.Insights/dataCollectionRules', variables('dcrName'))]"
     },
     "resources": [
         {
             "type": "Microsoft.Resources/deployments",
             "name": "[Concat('aks-monitoring-msi-dcr', '-',  uniqueString(variables('dcrName')))]",
             "apiVersion": "2017-05-10",
-            "subscriptionId": "[variables('workspaceSubscriptionId')]",
-            "resourceGroup": "[variables('workspaceResourceGroup')]",
+            "subscriptionId": "[variables('clusterSubscriptionId')]",
+            "resourceGroup": "[variables('clusterResourceGroup')]",
             "properties": {
               "mode": "Incremental",
               "template": {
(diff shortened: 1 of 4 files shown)


scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
```diff
@@ -38,19 +32,17 @@
         "clusterResourceGroup": "[split(parameters('aksResourceId'),'/')[4]]",
         "clusterName": "[split(parameters('aksResourceId'),'/')[8]]",
         "clusterLocation": "[replace(parameters('aksResourceLocation'),' ', '')]",
-        "workspaceSubscriptionId": "[split(parameters('workspaceResourceId'),'/')[2]]",
-        "workspaceResourceGroup": "[split(parameters('workspaceResourceId'),'/')[4]]",
         "dcrName": "[Concat('MSCI', '-', variables('clusterName'), '-', variables('clusterLocation'))]",
         "associationName":  "ContainerInsightsExtension",
-        "dataCollectionRuleId": "[resourceId(variables('workspaceSubscriptionId'), variables('workspaceResourceGroup'), 'Microsoft.Insights/dataCollectionRules', variables('dcrName'))]"
+        "dataCollectionRuleId": "[resourceId(variables('clusterSubscriptionId'), variables('clusterResourceGroup'), 'Microsoft.Insights/dataCollectionRules', variables('dcrName'))]"
```
ganga1980, line 37: what's the reason for moving DCR resource to cluster RG instead of workspace RG? The reason for keeping the DCR in workspace RG to avoid data loss if the cluster RG gets deleted.  Did you check moving DCR to cluster RG doesnt have any impact?

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
```diff
@@ -63,7 +55,7 @@
                         "type": "Microsoft.Insights/dataCollectionRules",
                         "apiVersion": "2021-04-01",
                         "name": "[variables('dcrName')]",
-                        "location": "[parameters('workspaceLocation')]",
+                        "location": "[parameters('clusterRegion')]",
```
ganga1980, line 58: Location of the DCR MUST be the same workspace location. DCR cant created with different region than workspace.

ganga1980, line 58: There is  no "clusterRegion" parameter defined in the template so this will cause deployment failure.

scripts/onboarding/templates/arc-k8s-extension-msi-auth/existingClusterOnboarding.json
```diff
@@ -20,12 +20,6 @@
         "description": "Azure Monitor Log Analytics Resource ID"
       }
     },
-    "workspaceRegion": {
```
ganga1980, line 23: Please note that,  MSI auth support exists in k8s-extension of the azure-cli extensions so which needs to be updated as well.

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
bragi92: Two reasons:

- We're creating the prometheus DCR's in the cluster's resource group and wanted to keep consistency between what both logs and CI are doing.
- The naming convention we're following only contains the cluster name and cluster location. This was causing an issue if we're using another cluster with the same and same location but sending data to the same workspace (same issue exists on the prom metrics side) as there would be conflicting DCRs. Hence the logic was to move the DCR into the clusters resource group so that we can avoid the conflict.

Yes, currently the UX private preview flight for the MAC, prometheus changes is also creating the DCR in the cluster RG and data is flowing as expected.

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
bragi92: yes, updated!

bragi92: removed

scripts/onboarding/templates/arc-k8s-extension-msi-auth/existingClusterOnboarding.json
bragi92: Will sync on this, with you.

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
ganga1980: Did you check with AMCS team whether they will be any issue accessing the historical data if the DCR Resource gets deleted (which is common if the customer deletes the cluster RG)?

ganga1980: Discussed with AMCS team and they confirmed there is no impact accessing historical data when the DCR deleted.

bragi92:
Making this PR against ci_prod as we've switched to using one branch

scripts/onboarding/templates/arc-k8s-extension-msi-auth/existingClusterOnboarding.json
bragi92: created draft PR with the change here -> https://github.com/AzureArcForKubernetes/azure-cli-extensions/pull/175

bragi92:
New PR here -> https://github.com/microsoft/Docker-Provider/pull/819/files

PR #817, opened 2022-09-09
try codeql

pfrcks:


ganga1980 (approved):

pfrcks:
@ganga1980 I removed cpp because for that we need to pass build instructions which require creating the complete setup and prereqs as well

ganga1980 (approved):

PR #827, opened 2022-09-23
workaround by symlinking docker location

pfrcks:
- symlink docker exe so that az acr login can find it


jatakiajanvi12 (approved):
Looks good to me

PR #832, opened 2022-09-28
Add support to specify resource endpoint for metrics

ms-hujia:
For cloud other than public Azure, a different resource endpoint for MSI retrieving might be used. This PR adds a method to override the default one used when sending metrics.

### DIFF
--- a/charts/azuremonitor-containers/values.yaml
+++ b/charts/azuremonitor-containers/values.yaml
@@ -65,7 +65,7 @@ amalogs:
   proxy: <your_proxy_config>
   # This metricsEndpoint used to define the endpoint custom metrics emit to. If not defined, default public Azure monitoring endpoint '{aks_region}.monitoring.azure.com' will be used.
   metricsEndpoint: <your_metrics_endpoint>
-  resourceEndpoint: <your_resource_endpoint>
+  tokenAudience: <your_token_audience>
   env:
     clusterName: <your_cluster_name>
     ## Applicable for only managed clusters hosted in Azure
--- a/charts/azuremonitor-containers/values.yaml
+++ b/charts/azuremonitor-containers/values.yaml
@@ -65,6 +65,7 @@ amalogs:
   proxy: <your_proxy_config>
   # This metricsEndpoint used to define the endpoint custom metrics emit to. If not defined, default public Azure monitoring endpoint '{aks_region}.monitoring.azure.com' will be used.
   metricsEndpoint: <your_metrics_endpoint>
+  resourceEndpoint: <your_resource_endpoint>
   env:
     clusterName: <your_cluster_name>
     ## Applicable for only managed clusters hosted in Azure
--- a/source/plugins/ruby/out_mdm.rb
+++ b/source/plugins/ruby/out_mdm.rb
@@ -119,6 +119,10 @@ module Fluent::Plugin
             if @isAADMSIAuth && !@isWindows
               @log.info "using IMDS sidecar endpoint for MSI token since its Arc k8s and Linux node"
               @useMsi = true
+              custom_resource_endpoint = ENV["customResourceEndpoint"]
+              if !custom_resource_endpoint.nil? && !custom_resource_endpoint.empty?
+                @@token_resource_audience = custom_resource_endpoint.strip
+              end
               msi_endpoint = @@imds_msi_endpoint_template % { resource: @@token_resource_audience }
               @parsed_token_uri = URI.parse(msi_endpoint)
             else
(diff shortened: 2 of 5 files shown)


source/plugins/ruby/out_mdm.rb
```diff
@@ -119,6 +119,10 @@ def start
             if @isAADMSIAuth && !@isWindows
               @log.info "using IMDS sidecar endpoint for MSI token since its Arc k8s and Linux node"
               @useMsi = true
+              custom_resource_endpoint = ENV["customResourceEndpoint"]
+              if !custom_resource_endpoint.nil? && !custom_resource_endpoint.empty?
+                @@token_resource_audience = custom_resource_endpoint.strip
```
ganga1980, line 124: what's the custom token resource audience look like? can you please share the audience names?

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
```diff
@@ -111,6 +111,10 @@ spec:
        - name: CUSTOM_METRICS_ENDPOINT
          value: {{ .Values.omsagent.metricsEndpoint | quote }}
        {{- end }}
+       {{- if ne .Values.omsagent.resourceEndpoint "<your_resource_endpoint>" }}
```
ganga1980, line 114: Please note that, our base branch is ci_prod so you will have to make PRs against ci_prod branch not the ci_dev.

ms-hujia: Got it, rebased.

source/plugins/ruby/out_mdm.rb
ms-hujia: For now, we suppose it to be in the format `https://monitor.{FQDN}/` while no final decision has been done yet.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
ganga1980, line 114: we have changed our agent name from "omsagent" to "ama-logs" so this needs to be updated.

ganga1980, line 114: rename "resourceEndpoint" to "tokenAudience"

source/plugins/ruby/out_mdm.rb
ganga1980: > For now, we suppose it to be in the format `https://monitor.{FQDN}/` while no final decision has been done yet.

> Did you test this change for both metrics and LA ingestion in Managed Identity mode?

charts/azuremonitor-containers/values.yaml
```diff
@@ -65,6 +65,7 @@ omsagent:
   proxy: <your_proxy_config>
   # This metricsEndpoint used to define the endpoint custom metrics emit to. If not defined, default public Azure monitoring endpoint '{aks_region}.monitoring.azure.com' will be used.
   metricsEndpoint: <your_metrics_endpoint>
+  resourceEndpoint: <your_resource_endpoint>
```
ganga1980, line 68: For Arc K8s, the audience specified here - https://github.com/microsoft/Docker-Provider/blob/ci_prod/charts/azuremonitor-containers/templates/ama-logs-arc-k8s-crd.yaml#L22 which will indicate the Arc K8s extension platform to get the token with the specified audience.

charts/azuremonitor-containers/templates/omsagent-daemonset.yaml
ms-hujia: Updated.

ms-hujia: Renamed all occurrences in YAML files.

source/plugins/ruby/out_mdm.rb
ms-hujia: This PR is only intended to support Metrics in MSI mode. Will make changes for Logs in a separate PR later.

charts/azuremonitor-containers/values.yaml
ms-hujia: Thanks for the information! Does this mean the corresponding Ruby code will be changed to get IMDS token from K8s CRD later?

ganga1980: Hi, @ms-hujia ,  The audience which we have is fixed. whatever the audience defined in the CRD matches with ruby code.  In your case, if the audience is custom, in that case, ruby code should use the same audience defined in the CRD.   do you have the steps which you have followed to validate this change?

source/plugins/ruby/out_mdm.rb
ganga1980: > This PR is only intended to support Metrics in MSI mode. Will make changes for Logs in a separate PR later.

> @ms-hujia , are you using the Arc K8s addon-token adapter which we have as side car container? If yes, then audience needs to be updated in the CRD and same audience should be used in ruby code.

PR #836, opened 2022-10-04
add_repo_type_binding

wanlonghenry:


deployment/arc-k8s-extension/ServiceGroupRoot/ScopeBindings/Public.ScopeBindings.json
```diff
@@ -33,7 +37,11 @@
                 {
                     "find": "__ACR_NAME__",
                     "replaceWith": "$(ACRName)"
-                },              
+                },
```
ganga1980, line 40: why do we need to replace **twice** _REPO_TYPE value in this file?

wanlonghenry: This file has all regions scopes including canary, pilot, ... so on, for each region scope, need to add the repo type

ganga1980 (approved):

PR #847, opened 2022-10-31
Telegraf agent config for DS/RS

pfrcks:
- adds config options for telegraf for DS RS containers
- config params exposed are: tcp_listener_chunk_size, tcp_listener_buffer_size, tcp_listener_mem_buf_limit


build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -57,6 +57,28 @@
...
+# Checking to see if container is not prometheus sidecar.
+# CONTAINER_TYPE is populated only for prometheus sidecar container.
+@containerType = ENV["CONTAINER_TYPE"]
+
+@promFbitChunkSize = 0
+@promFbitBufferSize = 0
+@promFbitMemBufLimit = 0
+
+# set defaults
+if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp(@daemonset) == 0
```
ganga1980, line 72: if condition can be merge except the last condition.

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -148,11 +148,28 @@ data:
...
     # buffer size should be greater than or equal to chunk size else we set it to chunk size.
+    # settings scoped to prometheus sidecar container
     [agent_settings.prometheus_fbit_settings]
       tcp_listener_chunk_size = 10
       tcp_listener_buffer_size = 10
       tcp_listener_mem_buf_limit = 200
 
+    # prometheus scrape fluent bit settings for high scale
+    # buffer size should be greater than or equal to chunk size else we set it to chunk size.
+    # settings scoped to daemonset container
```
ganga1980, line 159: ideally, config parameter should have units in it. I understand that, you are existing Nam convention then probably lets have comments and units of these config parameters are  in "MBs"

source/plugins/go/src/telemetry.go
```diff
@@ -194,6 +194,10 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 				if fbitTailMemBufLimitMBs != "" {
 					telemetryDimensions["FbitMemBufLimitSizeMBs"] = fbitTailMemBufLimitMBs
 				}
+				telemetryDimensions["PromFbitChunkSize"] = os.Getenv("AZMON_FBIT_CHUNK_SIZE")
```
ganga1980, line 197: add MB suffix for these telemetry keys.

kubernetes/linux/defaultpromenvvariables-rs
```diff
@@ -9,4 +9,3 @@ export AZMON_TELEGRAF_CUSTOM_PROM_PLUGINS_WITH_NAMESPACE_FILTER=""
 export AZMON_TELEGRAF_OSM_PROM_PLUGINS=""
```
ganga1980, line 9: this file can be reverted since there are no changes.

build/common/installer/scripts/tomlparser-agent-config.rb
ganga1980, line 72: since multiple places this condition has been used so this can be wrapped into the helper function for example isDaemonSet.

build/linux/installer/conf/td-agent-bit.conf
```diff
@@ -73,9 +73,9 @@
     Tag         oms.container.perf.telegraf.*
     Listen      0.0.0.0
     Port        25226
-    Chunk_Size  32
-    Buffer_Size 64
-    Mem_Buf_Limit 5m
+    Chunk_Size  ${AZMON_FBIT_CHUNK_SIZE}
+    Buffer_Size ${AZMON_FBIT_BUFFER_SIZE}
```
ganga1980, line 77: we will have to make the same change for Windows Telegraf conf file as well.

build/common/installer/scripts/tomlparser-agent-config.rb
pfrcks: not sure I understand. which parts?

kubernetes/container-azm-ms-agentconfig.yaml
pfrcks: i'll add in comment

source/plugins/go/src/telemetry.go
pfrcks: ack

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -57,6 +57,28 @@
...
+  @promFbitChunkSize = 32
+  @promFbitBufferSize = 64
+  @promFbitMemBufLimit = 5
+elsif !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp(@daemonset) != 0
+  @promFbitChunkSize = 32
+  @promFbitBufferSize = 64
```
ganga1980, line 78: we don't need to worry about 5m vs. 10m. Lets have 32,64, 10 as default and get rid of this condition.

source/plugins/go/src/telemetry.go
```diff
@@ -194,6 +194,10 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 				if fbitTailMemBufLimitMBs != "" {
 					telemetryDimensions["FbitMemBufLimitSizeMBs"] = fbitTailMemBufLimitMBs
 				}
+				telemetryDimensions["PromFbitChunkSizeKB"] = os.Getenv("AZMON_FBIT_CHUNK_SIZE")
```
ganga1980, line 197: This can be KB or MB. In case of the default, it will be KB else its MB.

source/plugins/go/src/telemetry.go
```diff
@@ -194,6 +194,10 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
...
+				telemetryDimensions["PromFbitBufferSizeKB"] = os.Getenv("AZMON_FBIT_BUFFER_SIZE")
```
ganga1980, line 198: same case here . This can be KB or MB. In case of the default, it will be KB else its MB.

source/plugins/go/src/telemetry.go
pfrcks: since the values now have the units we can remove the mention of KB/MB in key names

source/plugins/go/src/telemetry.go
pfrcks: since the values now have the units we can remove the mention of KB/MB in key names

ganga1980 (approved):

PR #857, opened 2022-11-16
Longw/kube events

wanlonghenry:
Kubernetes deprecated Event API Migration and Node name mismatch fix
Kubernetes event will be deprecated after v1.25: https://kubernetes.io/docs/reference/using-api/deprecation-guide/

### DIFF
--- a/source/plugins/ruby/in_kube_events.rb
+++ b/source/plugins/ruby/in_kube_events.rb
@@ -153,7 +153,10 @@ module Fluent::Plugin
             next
           end
 
-          nodeName = items["source"].key?("host") ? items["source"]["host"] : ""
+          nodeName = ""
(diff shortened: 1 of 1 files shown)


source/plugins/ruby/in_kube_events.rb
```diff
@@ -153,7 +153,7 @@ def parse_and_emit_records(events, eventQueryState, newEventQueryState, batchTim
             next
           end
 
-          nodeName = items["source"].key?("host") ? items["source"]["host"] : (OMS::Common.get_hostname)
+          nodeName = items["source"].key?("host") ? items["source"]["host"] : ""
```
ganga1980, line 156: This can break if the items["source"] doesnt exist. lets add null check around items["source"] for both places.

ganga1980 (approved):

PR #862, opened 2022-11-23
Geneva Config Map

cacarlt:


build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
```diff
@@ -0,0 +1,143 @@
...
+            @geneva_account_namespace = geneva_account_namespace
+            @geneva_logs_config_version = geneva_logs_config_version
+          else
+            puts "config::error:unsupported geneva config environment"
+          end
+        else
+          puts "config::error:invalid geneva logs config"
+        end
+        geneva_gcs_authid = geneva_logs_config[:authid]
+        if geneva_gcs_authid.nil? || geneva_gcs_authid.empty?
```
chris-raitano, line 64: Looks like the checks that the authid starts with "object_id#", etc were removed when the authid is provided in the config map. Wouldn't we still want these? Otherwise, we could end up in a state where we fail to parse the authid but the parser still states that everything was successfully parsed. Then the MA doesn't start but all the log messages show success

chris-raitano (approved):

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -0,0 +1,105 @@
...
+    $hostName = [System.Environment]::GetEnvironmentVariable("HOSTNAME", "process")
+    if (![string]::IsNullOrEmpty($hostName)) {
+        [System.Environment]::SetEnvironmentVariable("HOSTNAME", $hostName, "machine")
+        Write-Host "Successfully set environment variable HOSTNAME - $($hostName) for target 'machine'..."
+    }
+    else {
+        Write-Host "Failed to set environment variable HOSTNAME for target 'machine' since it is either null or empty"
+    }
+
+    $podName = [System.Environment]::GetEnvironmentVariable("PODNAME", "process")
```
ganga1980, line 36: why do we need this?

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -0,0 +1,105 @@
...
+        }
+        $env:AZMON_AGENT_CFG_SCHEMA_VERSION
+    }
+
+    # Set env vars for geneva monitor
+    $envVars = @{
+        MONITORING_DATA_DIRECTORY = (Join-Path $rootDir "opt\genevamonitoringagent\datadirectory")
+        MONITORING_GCS_AUTH_ID_TYPE = "AuthMSIToken"
+        MONITORING_GCS_REGION = $aksregion
+        MONITORING_TENANT = "CloudAgent"
```
ganga1980, line 56: These should be configurable from ConfigMap.

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -0,0 +1,105 @@
...
+Start-FileSystemWatcher
+
+if(Get-GenevaEnabled){
+    Invoke-Expression ".\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe -useenv"
+} else {
+    Write-Host "Geneva not configured. Watching for config map"
+    # Infinite loop keeps container alive while waiting for config map
+    # Otherwise when the process ends, kubernetes sees this as a crash and the container will enter a crash loop
+    while($true){
+        Start-Sleep 5
```
ganga1980, line 102: We should use ping or better way to prevent windows container stopping. Lets ping windows team to find out better guidance here and will share the windows engineering team contact offline.

kubernetes/host-logs-geneva.yaml
```diff
@@ -0,0 +1,96 @@
...
+       volumeMounts:
+        - mountPath: \etc\config\settings
+          name: settings-vol-config
+          readOnly: true
+       livenessProbe:
+          exec:
+            command:
+              - cmd
+              - /c
+              - .\opt\hostlogswindows\scripts\cmd\livenessprobe.exe
```
ganga1980, line 72: Missing the genevagent exe here. Liveness probe should monitor the geneva agent process. If the process terminates/hangs then liveness probe should trigger the restart.

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
```diff
@@ -0,0 +1,145 @@
...
+    if !parsedConfig.nil? && !parsedConfig[:hostlogs_settings].nil?
+      geneva_logs_config = parsedConfig[:hostlogs_settings][:geneva_logs_config]
+      if !geneva_logs_config.nil?
+        puts "config: parsing geneva_logs_config settings"
+        geneva_account_environment = geneva_logs_config[:environment]
+        geneva_account_name = geneva_logs_config[:account]
+        geneva_account_namespace = geneva_logs_config[:namespace]
+        geneva_logs_config_version = geneva_logs_config[:configversion]
+        if !geneva_account_environment.nil? && !geneva_account_name.nil? && !geneva_account_namespace.nil? && !geneva_logs_config_version.nil?
+          if GENEVA_SUPPORTED_ENVIRONMENTS.include?(geneva_account_environment)
```
ganga1980, line 51: Either we need to extend this map for airgap clouds or remove this check. Based on my discussion with geneva team, they recommend to skip the validation on the environment name as customer knows these values based on the internal doc.

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  if !file.nil?
+
+    if !@geneva_account_environment.empty? && !@geneva_account_name.empty? && !@geneva_account_namespace.empty? && !@geneva_logs_config_version.empty? && !@geneva_gcs_authid.empty?
+      file.write(get_command_windows("MONITORING_GCS_ENVIRONMENT", @geneva_account_environment))
+      file.write(get_command_windows("MONITORING_GCS_ACCOUNT", @geneva_account_name))
+      file.write(get_command_windows("MONITORING_GCS_NAMESPACE", @geneva_account_namespace))
+      file.write(get_command_windows("MONITORING_CONFIG_VERSION", @geneva_logs_config_version))
+      
+      authIdParts =  @geneva_gcs_authid.split('#', 2)
```
ganga1980, line 106: if the @geneva_gcs_authid is not valid, what will be behavior? is container gets crashed?

cacarlt: The monitoring agent just crashes and will retry connecting again every 120 seconds. The container however is seen as healthy in this instance.

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
cacarlt: Sounds good.

kubernetes/windows/hostlogs/main.ps1
cacarlt: Do we want to default to the AKS provided environment variables once we reach AKS RP?

kubernetes/windows/hostlogs/main.ps1
cacarlt: Sounds good.

kubernetes/windows/hostlogs/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+ARG IMAGE_TAG=win-ciprod10042022-3c05dd1b
+
+# Do not split this into multiple RUN!
+# Docker creates a layer for every RUN-Statement
+RUN powershell -Command "Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))"
+# Fluentd depends on cool.io whose fat gem is only available for Ruby < 2.5, so need to specify --platform ruby when install Ruby > 2.5 and install msys2 to get dev tools
+RUN choco install -y ruby --version 3.1.1.1 --params "'/InstallDir:C:\ruby31'" \
+&& choco install -y msys2 --version 20211130.0.0 --params "'/NoPath /NoUpdate /InstallDir:C:\ruby31\msys64'" \
+&& choco install -y vim
```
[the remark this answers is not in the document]
cacarlt: Removed

kubernetes/host-logs-geneva.yaml
cacarlt: Should this be put here instead of main.ps1

`if(Get-GenevaEnabled){
    Invoke-Expression ".\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe -useenv"
} else {
    Write-Host "Geneva not configured. Watching for config map"
    # Infinite loop keeps container alive while waiting for config map
    # Otherwise when the process ends, kubernetes sees this as a crash and the container will enter a crash loop
    while($true){
        Start-Sleep 5
    }
}`

kubernetes/windows/hostlogs/main.ps1
chris-raitano: We could move the FileSystemWatcher to the main process in this case. If we're not starting the agent, then there's no need to create a separate process for the FileSystemWatcher. And by running it in the main process, it'll keep the process alive

kubernetes/host-logs-geneva.yaml
chris-raitano: It might make sense to check Get-Geneva enabled in both places. We don't want to start the Geneva agent without the GCS variables, since we know it will crash. But we also don't want the liveness probe to trigger a restart if we intentionally didn't start the agent while waiting for a configmap to be mounted. If we're not consistent, we could get an infinite loop of false negatives from the liveness probe whenever the configmap isn't mounted yet

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
ganga1980: log the error message if the authid is not valid format.

cacarlt: Line 85 - 92 has logging for invalid authids. If there is a condition, we are overlooking let us know.

kubernetes/windows/hostlogs/main.ps1
cacarlt: Region is moved to the config parser.

AuthId Type is set to AuthMSIToken in the Config Parser.

kubernetes/windows/hostlogs/main.ps1
chris-raitano: Increased sleep duration to 3600 (1 hour). I'm seeing 0 cpu utilization during the sleep

kubernetes/host-logs-geneva.yaml
cacarlt: We have deiced to make this specific instance a separate PR and discussion will conduct during our next weekly meeting with @ganga1980. 

As the image stays separate from Container Insights main offering this isn't a must for us but will still be considered in scope until we can confirm with Ganga.

PR #863, opened 2022-11-23
collect troubleshooting data for syslog

pfrcks:
- copy dcr data folder from ds pod as part of troubleshooting log collection
- capture node image version info


scripts/troubleshoot/LogCollection/AgentLogCollection.sh
```diff
@@ -184,6 +185,9 @@ other_logCollection()
     fi
 
     kubectl get nodes > node.txt
+    # contains info regarding node image version, images present on disk, etc
+    kubectl get nodes -o json > node-detailed.json
```
ganga1980, line 189: if customer enabled syslog then we should check the node image version >= what we would need. This way customer can self troubleshoot without creating incident for us.

pfrcks: tough to add that since it depends on the image present on the box. lot of if else cases. wdyt?

pfrcks: maybe we can add a log output like "If you have enabled syslog please ensure that node image meets the following requirements: <insert doc link with minimum versions required>"

ganga1980: Ok, then just log _**aka.ms**_ link  for the public doc which has the details about the pre-requisites for the syslog feature.

pfrcks: sure I'll add a link with TODO for now

ganga1980 (approved):

PR #871, opened 2022-12-06
Changing ifconfig.co to MCR url

jatakiajanvi12:
- Updating the ifconfig.co with the mcr url to avoid curl requests on any external url

Testing steps: 
- Added the domain name to indicate usnat/ussec/azurechinacloud yaml
- Added the MCR_URLs as environment variables in yaml
- Correct URLs are being picked up as given in environment var for non-public clouds
- mcr.microsoft.com is picked up when domain is not provided.

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -353,8 +353,10 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
 
       if [ $? -ne 0 ]; then
             registry="https://mcr.microsoft.com/v2/"
-            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
-                  registry=$MCR_URL
+            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ] || [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
+                  if [ ! -z "$MCR_URL" ]; then
+                        registry=$MCR_URL
+                  fi
             fi
             echo "The registry is: $registry"
             if [ ! -z "$PROXY_ENDPOINT" ]; then
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -353,10 +353,11 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
 
       if [ $? -ne 0 ]; then
             registry="https://mcr.microsoft.com/v2/"
-            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ] || [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
-                  if [ ! -z "$MCR_URL" ]; then
-                        registry=$MCR_URL
-                  fi
+            if [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
+                  registry="https://mcr.azk8s.cn/v2/"
+            fi
+            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
+                  registry=$MCR_URL
             fi
             if [ -z $registry ]; then
                   echo "The environment variable MCR_URL is not set for CLOUD_ENVIRONMENT: $CLOUD_ENVIRONMENT"
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -355,8 +355,7 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
             registry="https://mcr.microsoft.com/v2/"
             if [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
                   registry="https://mcr.azk8s.cn/v2/"
-            fi
-            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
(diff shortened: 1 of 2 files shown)


kubernetes/linux/main.sh
```diff
@@ -334,32 +334,35 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
       fi
 
       if [ $? -ne 0 ]; then
+            registry="mcr.microsoft.com"
```
ganga1980, line 337: prefix https:// missing.  you can have registry=https://mcr.microsoft.com/v2  and use this value in applicable places

kubernetes/linux/main.sh
```diff
@@ -334,32 +334,35 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
...
+            if  [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
+                  registry=$MCR_URL
```
ganga1980, line 339: Please test the changes.

ganga1980, line 339: This requires change in AKS side.

jatakiajanvi12: Not able to create a PR there because in the process of getting permissions. Will merge this PR after that is done

jatakiajanvi12: Created a PR on aks-rp

jatakiajanvi12: Tested the changes. Adding the scenarios in the description

kubernetes/linux/main.sh
```diff
@@ -334,32 +352,39 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
       fi
 
       if [ $? -ne 0 ]; then
+            registry="https://mcr.microsoft.com/v2/"
+            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ] || [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
+                  if [ ! -z "$MCR_URL" ]; then
```
ganga1980, line 357: q1. what will will be the curl response if the MCR_URL is not set?  we should also the log message here if the MCR_URL is empty/not-set.

jatakiajanvi12: Yes makes sense. Added the logic for trying curl only if registry is present. Also, added the log messaged in failure cases

kubernetes/linux/main.sh
```diff
@@ -334,32 +352,43 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
...
-            if [ ! -z "$PROXY_ENDPOINT" ]; then
-               if [ -e "/etc/ama-logs-secret/PROXYCERT.crt" ]; then
-                  echo "Making curl request to ifconfig.co with proxy and proxy CA cert"
-                  RET=`curl --max-time 10 -s -o /dev/null -w "%{http_code}" ifconfig.co --proxy $PROXY_ENDPOINT --proxy-cacert /etc/ama-logs-secret/PROXYCERT.crt`
-               else
-                  echo "Making curl request to ifconfig.co with proxy"
-                  RET=`curl --max-time 10 -s -o /dev/null -w "%{http_code}" ifconfig.co --proxy $PROXY_ENDPOINT`
-               fi
+            registry="https://mcr.microsoft.com/v2/"
+            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ] || [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
```
ganga1980, line 356: we can have the azurechinacloud in main.sh itself and we dont need to pass down from RP.

jatakiajanvi12: Updated this.

kubernetes/linux/main.sh
```diff
@@ -334,32 +352,44 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
...
+            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
+                  registry=$MCR_URL
+            fi
+            if [ -z $registry ]; then
+                  echo "The environment variable MCR_URL is not set for CLOUD_ENVIRONMENT: $CLOUD_ENVIRONMENT"
+                  RET=000
             else
-                  echo "Making curl request to ifconfig.co"
-                  RET=$(curl --max-time 10 -s -o /dev/null -w "%{http_code}" ifconfig.co)
+                  if [ ! -z "$PROXY_ENDPOINT" ]; then
```
ganga1980, line 366: There is indention issue. Block under if [ ! -z "$PROXY_ENDPOINT" ]; then needs to be intended correctly.

kubernetes/linux/main.sh
```diff
@@ -334,32 +352,44 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
...
-                  RET=`curl --max-time 10 -s -o /dev/null -w "%{http_code}" ifconfig.co --proxy $PROXY_ENDPOINT --proxy-cacert /etc/ama-logs-secret/PROXYCERT.crt`
-               else
-                  echo "Making curl request to ifconfig.co with proxy"
-                  RET=`curl --max-time 10 -s -o /dev/null -w "%{http_code}" ifconfig.co --proxy $PROXY_ENDPOINT`
-               fi
+            registry="https://mcr.microsoft.com/v2/"
+            if [ $CLOUD_ENVIRONMENT == "azurechinacloud" ]; then
+                  registry="https://mcr.azk8s.cn/v2/"
+            fi
+            if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ]; then
```
ganga1980, line 359: I think, we can use **elif** instead of two separate if conditions here.

ganga1980 (approved):

PR #872, opened 2022-12-06
fix: add telemetry to find the number of impossibly high cpu clusters

bragi92:


source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -247,6 +247,14 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
               metricItem["json_Collections"] = metricCollections.to_json
               metricItems.push(metricItem)
 
+              if metricValue > 10000000000
+                telemetryProperties = {}
+                telemetryProperties["metricJSON"] = JSON.generate(metricJSON)
```
ganga1980, line 252: This will be huge (in the order of MBs) if the node has 100 or more containers, right? do we need all the payload?

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -247,6 +247,14 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
               metricItem["json_Collections"] = metricCollections.to_json
               metricItems.push(metricItem)
 
+              if metricValue > 10000000000
```
ganga1980, line 250: probable we can just log the exception with container name and value if the cpu usage value more than node's cpu usage, right?

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
bragi92: removed

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
bragi92: I was thinking that we shouldn't rely on the numbers coming from the cadvisor endpoint. Currently the value is 10^10.
Do you think a number like this makes more sense or the node cpu usage value? I will update it accordingly.

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -187,7 +187,7 @@ def getMetrics(winNode: nil, namespaceFilteringMode: "off", namespaces: [], metr
             end
             metricDataItems.push(getNodeMetricItem(metricInfo, hostName, "memory", "workingSetBytes", Constants::MEMORY_WORKING_SET_BYTES, metricTime))
 
-            metricDataItems.push(getNodeLastRebootTimeMetric(metricInfo, hostName, "restartTimeEpoch", metricTime))
+            ############################################# metricDataItems.push(getNodeLastRebootTimeMetric(metricInfo, hostName, "restartTimeEpoch", metricTime))
```
ganga1980, line 190: why this being commented?

PR #874, opened 2022-12-08
Enable fluent-bit multiline logs

pfrcks:
- Enables multiline log collection through configmap
- Supports CRI, Docker, go


build/common/installer/conf/azm-containers-parser-multiline.conf
```diff
@@ -0,0 +1,17 @@
+# [MULTILINE_PARSER]
+#     name          dotnet-regex
+#     type          regex
+#     flush_timeout 1000
```
ganga1980, line 4: what's the units of this value? is it 1000 seconds? how this being determined? should we make this configurable through configmap?

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -1,7 +1,7 @@
 #!/usr/local/bin/ruby
 require_relative "ConfigParseErrorLogger"
 
-@td_agent_bit_conf_path = "/etc/opt/microsoft/docker-cimprov/td-agent-bit.conf"
+@td_agent_bit_conf_path = "/etc/opt/microsoft/docker-cimprov/fluent-bit.conf"
```
ganga1980, line 4: lets rename td_agent_bit -> fluent_bit so we have everything consistent new name.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -150,6 +151,21 @@ def populateSettingValuesFromConfigMap(parsedConfig)
       ConfigParseErrorLogger.logError("Exception while reading config map settings for container log schema version - #{errorStr}, using defaults, please check config map for errors")
     end
 
+    # Get multiline log stitching setting
+    begin
+      if !parsedConfig[:log_collection_settings][:stitch_multiline_logs].nil? && !parsedConfig[:log_collection_settings][:stitch_multiline_logs][:enabled].nil?
+        @logStitchMultiline = parsedConfig[:log_collection_settings][:stitch_multiline_logs][:enabled]
+        puts "config::Using config map setting for multiline logging"
+
+        if @containerLogSchemaVersion.strip.casecmp("v2") !=0
```
ganga1980, line 160: why are we overriding to v2? This will break existing customers who are using v1 when they enable multi-line feature. We should make the v2 is pre-req for the multi-line log feature. if the  customer applied multi-line log feature on v1 schema, then this setting should be ignored.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -150,6 +151,21 @@ def populateSettingValuesFromConfigMap(parsedConfig)
       ConfigParseErrorLogger.logError("Exception while reading config map settings for container log schema version - #{errorStr}, using defaults, please check config map for errors")
     end
 
+    # Get multiline log stitching setting
+    begin
+      if !parsedConfig[:log_collection_settings][:stitch_multiline_logs].nil? && !parsedConfig[:log_collection_settings][:stitch_multiline_logs][:enabled].nil?
```
ganga1980, line 156: instead of stitch_multiline_logs, should we call enable_multiline_logs ?

build/common/installer/conf/azm-containers-parser-multiline.conf
```diff
@@ -0,0 +1,17 @@
...
+    # ---------------------------------
+    #
+    # configuration hints:
+    #
+    #  - first state always has the name: start_state
+    #  - every field in the rule must be inside double quotes
+    #
+    # rules |   state name  | regex pattern                  | next state
+    # ------|---------------|--------------------------------------------
+    # rule      "start_state"    "/^System.*/"                    "cont1"
```
ganga1980, line 16: is this only covers the exception starts with System.*, right? are there any other scenarios (which  has application specific exceptions) needs to be covered?

ganga1980 (changes requested):

kubernetes/linux/setup.sh
```diff
@@ -67,11 +67,11 @@ chmod 544 /opt/telegraf
 # Use wildcard version so that it doesnt require to touch this file
 /$TMPDIR/docker-cimprov-*.*.*-*.*.sh --install
 
-#download and install fluent-bit(td-agent-bit)
+#download and install fluent-bit(fluent-bit)
 wget -qO - https://packages.fluentbit.io/fluentbit.key | sudo apt-key add -
 sudo echo "deb https://packages.fluentbit.io/ubuntu/bionic bionic main" >> /etc/apt/sources.list
 sudo apt-get update
-sudo apt-get install td-agent-bit=1.7.8 -y
+sudo apt-get install fluent-bit=2.0.5 -y
```
ganga1980, line 74: are there any changes to the behavior or functionality in 2.0.5 compared to 1.7.8 for both Linux & windows?

build/common/installer/conf/azm-containers-parser-multiline.conf
pfrcks: yes it's in seconds. I don't think we need to expose this for private preview

build/common/installer/conf/azm-containers-parser-multiline.conf
pfrcks: working on improving this

kubernetes/linux/setup.sh
pfrcks: cpu
memory

pfrcks: same comparison for v1 route

cpu
cpuv1

memory
memoryv1

build/common/installer/conf/azm-containers-parser-multiline.conf
ganga1980: @pfrcks , what's the rationale behind picking 1000 seconds?  since flush timeout is 1000 seconds, so the stitched logs shows after 1000 seconds in LA? and also this might cause data loss in high log volume environment (with back pressure) due to accumulation of the logs for this duration , right?

kubernetes/linux/setup.sh
ganga1980: Thanks for sharing these. do we have the CPU and memory charts with multi-line for container log v2?

pfrcks: I already had shared those results. will share again on teams

build/common/installer/scripts/tomlparser.rb
```diff
@@ -150,6 +151,21 @@ def populateSettingValuesFromConfigMap(parsedConfig)
       ConfigParseErrorLogger.logError("Exception while reading config map settings for container log schema version - #{errorStr}, using defaults, please check config map for errors")
     end
 
+    # Get multiline log enabling setting
+    begin
+      if !parsedConfig[:log_collection_settings][:enable_multiline_logs].nil? && !parsedConfig[:log_collection_settings][:enable_multiline_logs][:enabled].nil?
```
ganga1980, line 156: I didnt see the configmap updates for this, is this intentional?

pfrcks: added

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -328,6 +334,10 @@ def populateSettingValuesFromConfigMap(parsedConfig)
     file.write("export IGNORE_PROXY_SETTINGS=#{@ignoreProxySettings}\n")
   end
 
+  if @multiline_enabled.strip.casecmp("true") == 0
```
ganga1980, line 337: have @multiline_enabled=false default, otherwise this can be undefined value if this not defined in the configmap, right?

build/common/installer/conf/azm-containers-parser-multiline.conf
```diff
@@ -0,0 +1,18 @@
+[MULTILINE_PARSER]
+    name          dotnet-multiline
+    type          regex
+    flush_timeout 1000
```
ganga1980, line 4: as we discussed add comments for the units of the flush_timeout or use the 1s instead of 1000 .

ganga1980 (approved):

PR #877, opened 2022-12-16
Add pipeline for host-logs-windows image

magicMystery:
This PR adds a pipeline, similar to the existing .pipelines/azure_pipeline_mergedbranches.yaml, but targeting the sky windows-host-logs image. It also pushes the image to CI's acr.


cacarlt (approved):

cacarlt:
With this PR we are able to move to Infra using kubectl apply commands until we are unblocked by AKS RP.

PR #883, opened 2023-01-04
Gangams/dec 2022 agent release tasks

ganga1980:
This PR has following changes

1. MSI Onboarding ARM templates for handling DCR naming and handling the scenario where DCR name is more than 64 characters and also removing the spaces in the workspace region if user inputs the workspace region name with spaces
2. config option to ignore proxy settings for the AMPLS + Proxy environment.

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -355,6 +355,8 @@ if [ -e "/etc/ama-logs-secret/WSID" ]; then
                   if [ -e "/etc/ama-logs-secret/PROXYCERT.crt" ]; then
                         export PROXY_CA_CERT=/etc/ama-logs-secret/PROXYCERT.crt
                         echo "export PROXY_CA_CERT=$PROXY_CA_CERT" >> ~/.bashrc
+                        cp /etc/ama-logs-secret/PROXYCERT.crt /usr/local/share/ca-certificates/proxy-cert.crt
+                        update-ca-certificates
                   fi
             fi
       fi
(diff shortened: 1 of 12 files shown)


kubernetes/linux/main.sh
```diff
@@ -263,6 +263,24 @@ if [[ ((! -e "/etc/config/kube.conf") && ("${CONTAINER_TYPE}" == "PrometheusSide
       fi
 fi
 
+#Parse the configmap to set the right environment variables for agent config.
+#Note > tomlparser-agent-config.rb has to be parsed first before td-agent-bit-conf-customizer.rb for fbit agent settings
+if [ "${CONTAINER_TYPE}" != "PrometheusSidecar" ]; then
```
pfrcks, line 268: why did we shift this above?

kubernetes/windows/main.ps1
```diff
@@ -325,7 +336,6 @@ function Set-EnvironmentVariables {
     ruby /opt/amalogswindows/scripts/ruby/tomlparser-mdm-metrics-config.rb
     .\setmdmenv.ps1
 }
-
 function Get-ContainerRuntime {
```
pfrcks, line 339: add newline above

kubernetes/linux/main.sh
ganga1980: IGNORE_PROXY_SETTINGS being set in tomlparser-agent-config which required at line 311.

kubernetes/windows/main.ps1
ganga1980: thanks for catching. Fixed.

pfrcks (approved):
lgtm

PR #886, opened 2023-01-06
add xvda in devices

wanlonghenry:
add xvda in devices so that telegraf can fetch the Disk IO data from xvda disk as well

### DIFF
--- a/build/linux/installer/conf/telegraf.conf
+++ b/build/linux/installer/conf/telegraf.conf
@@ -465,7 +465,7 @@
   ## By default, telegraf will gather stats for all devices including
   ## disk partitions.
   ## Setting devices will restrict the stats to the specified devices.
-  devices = ["sd[a-z][0-9]", "xvda"]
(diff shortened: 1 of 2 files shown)


build/linux/installer/conf/telegraf.conf
```diff
@@ -465,7 +465,7 @@
   ## By default, telegraf will gather stats for all devices including
   ## disk partitions.
   ## Setting devices will restrict the stats to the specified devices.
-  devices = ["sd[a-z][0-9]"]
+  devices = ["sd[a-z][0-9]", "xvda"]
```
ganga1980, line 468: we should add regex to cover all different devices of xvd. Refer to this https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html

wanlonghenry: added and changed to xvd[a-z]

ganga1980 (approved):

ganga1980 (approved):

PR #902, opened 2023-02-15
Fixing MUTE_PROM_SIDECAR variable when no config applied

jatakiajanvi12:
The MUTE_PROM_SIDECAR variable was false even when no config was applied. Fixed it to consider empty variable names as well.

fix-mute_prom

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -531,6 +531,8 @@ if [ -e "telemetry_prom_config_env_var" ]; then
             echo $line >>~/.bashrc
       done
       source telemetry_prom_config_env_var
+else
+      setGlobalEnvVar TELEMETRY_CUSTOM_PROM_MONITOR_PODS false
 fi
 
 #Parse sidecar agent settings for custom configuration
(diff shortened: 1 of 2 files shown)


kubernetes/linux/main.sh
```diff
@@ -581,8 +581,8 @@ fi
 
 # If the prometheus sidecar isn't doing anything then there's no need to run mdsd and telegraf in it.
 if [[ ( "${CONTAINER_TYPE}" == "PrometheusSidecar" ) &&
-      ( "${TELEMETRY_CUSTOM_PROM_MONITOR_PODS}" == "false" ) &&
-      ( "${TELEMETRY_OSM_CONFIGURATION_NAMESPACES_COUNT}" -eq 0 ) ]]; then
+      ( "${TELEMETRY_CUSTOM_PROM_MONITOR_PODS}" == "false"  || "${TELEMETRY_CUSTOM_PROM_MONITOR_PODS}" == "" ) &&
```
pfrcks, line 584: how about "${TELEMETRY_CUSTOM_PROM_MONITOR_PODS}" != "true"

kubernetes/linux/main.sh
```diff
@@ -581,8 +581,8 @@ fi
...
+      ( "${TELEMETRY_OSM_CONFIGURATION_NAMESPACES_COUNT}" -eq 0 || "${TELEMETRY_OSM_CONFIGURATION_NAMESPACES_COUNT}" == "" ) ]]; then
```
pfrcks, line 585: can this be improved as well so that there is just 1 condition? is it possible to set default value for TELEMETRY_OSM_CONFIGURATION_NAMESPACES_COUNT in tomlparser-prom-customconfig.rb?

jatakiajanvi12: Added default value of variable in main.sh

pfrcks (approved):
approved. pls however check that the changes are working as expected, close the hole we had and do not have any regressions before merging

PR #929, opened 2023-03-16
disable telegraf on replicaset by default

pfrcks:
- currentl telegraf does not do anything on replicaset since sidecar scraping is enabled by default. Therefore switch it off


kubernetes/linux/main.sh
```diff
@@ -932,11 +932,15 @@ fi
 
 #start telegraf
 if [ "${MUTE_PROM_SIDECAR}" != "true" ]; then
-      /opt/telegraf --config $telegrafConfFile &
-      echo "telegraf version: $(/opt/telegraf --version)"
-      dpkg -l | grep fluent-bit | awk '{print $2 " " $3}'
+    if [ "${CONTROLLER_TYPE}" == "ReplicaSet" ] && [ "${SIDECAR_SCRAPING_ENABLED}" == "true" ]; then
```
jatakiajanvi12, line 935: Can we use the same condition as livenessprobe? if [ "${CONTROLLER_TYPE}" == "ReplicaSet" ] && [ ! -z "${SIDECAR_SCRAPING_ENABLED}" ] && [ "${SIDECAR_SCRAPING_ENABLED}" == "true" ]; then

jatakiajanvi12: Yeah true. May be we can remove from Livenessprobe as well then? Was just thinking to keep it the same at both places unless you think otherwise.

jatakiajanvi12 (approved):

build/linux/installer/scripts/livenessprobe.sh
```diff
@@ -85,9 +85,14 @@ fi
 (ps -ef | grep telegraf | grep -v "grep")
 if [ $? -ne 0 ]
 then
- # echo "Telegraf is not running" > /dev/termination-log
- echo "Telegraf is not running (controller: ${CONTROLLER_TYPE}, container type: ${CONTAINER_TYPE})" > /dev/write-to-traces  # this file is tailed and sent to traces
- # exit 1
+ if [ "${CONTROLLER_TYPE}" == "ReplicaSet" ] && [ ! -z "${SIDECAR_SCRAPING_ENABLED}" ] && [ "${SIDECAR_SCRAPING_ENABLED}" == "true" ]; then
```
ganga1980, line 88: SIDECAR_SCRAPING_ENABLED enabled by default for AKS & Arc. I dont think this required to check. But telegraf in replicaset used if the customer enabled configmap with cluster level prometheus settings. you would need to activate & check telegraf if the customer enabled cluster level prometheus settings.   Did you check that this is working as expected?

kubernetes/linux/main.sh
ganga1980: This change doesnt consider cluster prometheus config. Telegraf only need to turned off if there is not cluster config enabled.

build/common/installer/scripts/tomlparser-prom-customconfig.rb
```diff
@@ -149,6 +151,10 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           urls = parsedConfig[:prometheus_data_collection_settings][:cluster][:urls]
           kubernetesServices = parsedConfig[:prometheus_data_collection_settings][:cluster][:kubernetes_services]
 
+          if fieldPass.nil? && fieldDrop.nil? && urls.nil? && kubernetesServices.nil?
```
ganga1980, line 154: We should also need to turn off telegraf in RS when there is no configmap  applied or configmap with no cluster config settings, right?

pfrcks:
verified 4 scenarios

no configmap - telegraf disabled in rs
configmap with no cluster level settings - telegraf disabled in rs
config with commented cluster level settings - telegraf disabled in rs
config with actual cluster level settings - telegraf enabled in rs

ganga1980 (approved):

PR #931, opened 2023-03-20
dont set AMA specific env vars if 1p mode not configured

ganga1980:
Dont set AMA specific env vars if 1p mode not configured


jatakiajanvi12:
Tested this with non geneva scenario and it is working now.

jatakiajanvi12 (approved):

PR #932, opened 2023-03-20
Update AgentLogCollection.sh

bdegs:
-Script was failing to recognize daemon set and replica set on new AKS Cluster, modified script so it now recognizes these. 
-Added in additional logging lines (for visibility on files being collected, and for help in debugging). 
-Added in section to collect both container-azm-ms-aks-k8scluster and ama-logs-rs-config. 
Additional Notes: Some of the kubectl exec commands resulted in errors, yet files are still collected. These may need to be investigated.

### DIFF
--- a/scripts/troubleshoot/LogCollection/AgentLogCollection.sh
+++ b/scripts/troubleshoot/LogCollection/AgentLogCollection.sh
@@ -1,28 +1,27 @@
-#!/bin/bash
-#
+# !/bin/bash
 # Copyright (c) Microsoft Corporation.
-#
-# This script will collect all logs from the replicaset agent pod and a random daemonset pod, also collect onboard logs with processes
-#
-# Author Nina Li
+# Description: This script will collect all logs from the Replicaset Agent Pod and Daemonset Agent Pod.
+# Also collects onboarding logs with processes.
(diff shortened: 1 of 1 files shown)


bdegs:
@microsoft-github-policy-service agree company="Microsoft"

scripts/troubleshoot/LogCollection/AgentLogCollection.sh
```diff
@@ -1,36 +1,35 @@
-#!/bin/bash
-#
+# !/bin/bash
 # Copyright (c) Microsoft Corporation.
```
ganga1980, line 2: @bdegs , script should be used in the ci_prod branch not the ci_dev - https://github.com/microsoft/Docker-Provider/blob/ci_prod/scripts/troubleshoot/LogCollection/AgentLogCollection.sh.   
which  link you are using which taking ci_dev branch? 
cc: @pfrcks

bdegs: @ganga1980 , used incorrect link when creating this pull request (must have been link saved by browser cache. I'll cancel this one and make one in the correct link. Thanks for the review!

PR #933, opened 2023-03-21
Gangams/fix pod ready for completed jobs

ganga1980:
Fix the issue where the Job pods which are completed/succeeded marked as not ready.

### DIFF
--- a/source/plugins/ruby/KubernetesApiClient.rb
+++ b/source/plugins/ruby/KubernetesApiClient.rb
@@ -1390,9 +1390,7 @@ class KubernetesApiClient
             if condition["type"] == "Ready"
               if condition["status"].downcase == "true"
                 podReadyCondition = true
-              elsif !controllerKind.nil? & !controllerKind.empty? &&
-                    !podStatus.nil? && !podStatus.empty? &&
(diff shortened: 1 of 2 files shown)


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -1382,14 +1382,16 @@ def getHpaOptimizedItem(resourceItem)
...
-    def getPodReadyCondition(podStatusConditions)
+    def getPodReadyCondition(podStatusConditions, controllerKind, podStatus)
       podReadyCondition = false
       begin
         if !podStatusConditions.nil? && !podStatusConditions.empty?
           podStatusConditions.each do |condition|
             if condition["type"] == "Ready"
               if condition["status"].downcase == "true"
                 podReadyCondition = true
+              elsif isCompletedJobPod(controllerKind, podStatus)
```
wanlonghenry, line 1393: can we combine this and make this an or condition?

ganga1980: > can we combine this and make this an or condition?

fixed.

wanlonghenry (approved):

PR #936, opened 2023-03-29
add nil check in kubelet

wanlonghenry:
add nil check in kubelet to address this issue: https://msazure.visualstudio.com/InfrastructureInsights/_workitems/edit/16519364

### DIFF
--- a/source/plugins/ruby/kubelet_utils.rb
+++ b/source/plugins/ruby/kubelet_utils.rb
@@ -56,9 +56,9 @@ class KubeletUtils
         parsed_response = JSON.parse(allocatable_response.body)
 
         begin
-          kubereserved_cpu = parsed_response["kubeletconfig"]["kubeReserved"]["cpu"]
-          if kubereserved_cpu.nil? || kubereserved_cpu == ""
-            kubereserved_cpu = "0.0"
+          kubereserved_cpu = "0.0"
+          if !(parsed_response["kubeletconfig"].nil || parsed_response["kubeletconfig"]["kubeReserved"].nil && parsed_response["kubeletconfig"]["kubeReserved"]["cpu"].nil)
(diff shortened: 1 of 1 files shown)


source/plugins/ruby/kubelet_utils.rb
```diff
@@ -56,9 +56,9 @@ def get_node_allocatable(cpu_capacity, memory_capacity)
         parsed_response = JSON.parse(allocatable_response.body)
 
         begin
-          kubereserved_cpu = parsed_response["kubeletconfig"]["kubeReserved"]["cpu"]
-          if kubereserved_cpu.nil? || kubereserved_cpu == ""
-            kubereserved_cpu = "0.0"
+          kubereserved_cpu = "0.0"
+          if parsed_response.key?("kubeletconfig") && parsed_response["kubeletconfig"].key?("kubeReserved") && parsed_response["kubeletconfig"]["kubeReserved"].key?("cpu")
```
jatakiajanvi12, line 60: You haven't added the empty string check: kubereserved_cpu == "". The keys can be present but the value can still be empty right?

Same everywhere else

jatakiajanvi12 (changes requested):

source/plugins/ruby/kubelet_utils.rb
wanlonghenry: from the previous logic, if the kubereserved_cpu == "" then the value will be overriding to 0.

wanlonghenry: addressed

wanlonghenry (commented):
changed

ganga1980 (approved):

jatakiajanvi12 (approved):

PR #949, opened 2023-04-11
Allow tag override

pfrcks:
- Allows tags to be overwritten when OverrideTag is set to true in pipeline run
- Fixes bug which checks existing tag for linux

### DIFF
--- a/deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
+++ b/deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
@@ -22,15 +22,17 @@ fi
 
 if [[ "$AGENT_IMAGE_FULL_PATH" == *"win-"* ]]; then
   echo "checking windows tags"
-  TAG_EXISTS=$(echo $MCR_TAG_RESULT | jq '.tags | contains(["'"win-$AGENT_IMAGE_TAG_SUFFIX"'"])')
+  echo $MCR_TAG_RESULT | jq '.tags' | grep -q \"win-"$AGENT_IMAGE_TAG_SUFFIX"\"
 else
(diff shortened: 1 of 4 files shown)


deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
```diff
@@ -22,13 +22,17 @@ fi
 
 if [[ "$AGENT_IMAGE_FULL_PATH" == *"win-"* ]]; then
   echo "checking windows tags"
-  TAG_EXISTS=$(echo $MCR_TAG_RESULT | jq '.tags | contains(["'"win-$AGENT_IMAGE_TAG_SUFFIX"'"])')
+  echo $MCR_TAG_RESULT | jq '.tags' | grep -q \"win-"$AGENT_IMAGE_TAG_SUFFIX"\"
 else
   echo "checking linux tags"
-  TAG_EXISTS=$(echo $MCR_TAG_RESULT | jq '.tags | contains(["'":$AGENT_IMAGE_TAG_SUFFIX"'"])')
+  echo $MCR_TAG_RESULT | jq '.tags' | grep -q \""$AGENT_IMAGE_TAG_SUFFIX"\"
```
ganga1980, line 28: This can give false results when the windows agent image pushed already available, right? For example, this condition will be true when MCR has the image has  tag (win-3.1.6) for just windows , right?

pfrcks: no because we are grepping for "3.1.6" (linux) and "win-3.1.6" (windows) including the quotes. grep -q \"3.1.6\" will never match "win-3.1.6"

ganga1980 (approved):

PR #953, opened 2023-04-14
Update README.md

bdegs:
Minor updates to readme.md file. Tested script on ARO clusters with success and updated instructions for ARO clusters.

### DIFF
--- a/scripts/troubleshoot/LogCollection/README.md
+++ b/scripts/troubleshoot/LogCollection/README.md
@@ -9,6 +9,7 @@ This tool will collect:
 ## Prerequisites
 * kubectl: az aks install-cli
 * tar (installed by default)
+* All nodes should be on AKS or running ARO
 * OpenShift CLI (For ARO Clusters Only) https://learn.microsoft.com/en-us/azure/openshift/tutorial-connect-cluster#install-the-openshift-cli
 * Container Insights is enabled: https://docs.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-onboard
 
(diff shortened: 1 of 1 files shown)


scripts/troubleshoot/LogCollection/README.md
```diff
@@ -1,19 +1,20 @@
...
-* cluster/node info, pod deployment, configMap, process logs etc..
+* Agent logs from Linux ds (ama-logs-) and rs (ama-logs-rs-) pods
+* Agent logs from Windows pod if enabled
+* Cluster/node info, pod deployment, configMap, process logs etc.
+* Note: Script can collect logs from both AKS Clusters as well as ARO Clusters
 
 ## Prerequisites
 * kubectl: az aks install-cli
 * tar (installed by default)
-* all nodes should be running on AKS
```
jatakiajanvi12, line 11: This pre-requisite shouldn't deleted. May be we can mention, nodes should be on AKS or running ARO?

jatakiajanvi12 (changes requested):

scripts/troubleshoot/LogCollection/README.md
bdegs: Agreed, I went ahead and added that to the prereqs.

jatakiajanvi12 (approved):

pfrcks:
@bdegs this looks great. I was wondering if we should add a note that customers onboarded via ARO are suggested to migrate to ARC based installations since ARO path is deprecated.?

PR #954, opened 2023-04-18
enable multiline in fluent-bit-common

pfrcks:


pfrcks:
@ganga1980 should the similar be done for geneva-logs_tenant and geneva-logs_infra fluent-bit conf files?

jatakiajanvi12 (approved):

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -2,6 +2,7 @@
 require_relative "ConfigParseErrorLogger"
 
 @fluent_bit_conf_path = "/etc/opt/microsoft/docker-cimprov/fluent-bit.conf"
+@fluent_bit_common_conf_path = "/etc/opt/microsoft/docker-cimprov/fluent-bit-common.conf"
```
ganga1980, line 6: we need to do same the for fluent-bit-common.conf file in windows also.

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -2,6 +2,7 @@
...
```
ganga1980, line 5: we need to do same the for fluent-bit-common.conf file in windows also.

pfrcks: @ganga1980 thanks for pointing it out. added it below.

ganga1980 (approved):

PR #957, opened 2023-04-25
add logic to check and wait for token adapter

wanlonghenry:


kubernetes/linux/main.sh
```diff
@@ -58,6 +58,34 @@ waitforlisteneronTCPport() {
       fi
 }
 
+wait_for_token_adapter() {
```
pfrcks, line 61: overall looks ok but do run this through the shellcheck command to see if it identifies any errors/pitfalls

you can add it to your vscode using https://github.com/vscode-shellcheck/vscode-shellcheck

PR #960, opened 2023-04-27
Adding network listener waittime to configmap

jatakiajanvi12:


kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -189,6 +189,12 @@ data:
     # [agent_settings.proxy_config]
     #    ignore_proxy_settings = "true"  # if this is not applied, default value is false
 
+    #Configuration settings for the waittime for the network listeners to be available
+    [agent_settings.network_listener_waittime]
```
ganga1980, line 193: 1. Make sure to comment out these with default values in the final config map.
2.  Add the comment on each port so that its easy for customer and CSS to understand what this port used for

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -309,6 +313,28 @@ def populateSettingValuesFromConfigMap(parsedConfig)
         @multiline_enabled = multiline_config[:enabled]
         puts "Using config map value: AZMON_MULTILINE_ENABLED = #{@multiline_enabled}"
       end
+
+      network_listener_waittime_config = parsedConfig[:agent_settings][:network_listener_waittime]
+      if !network_listener_waittime_config.nil?
+        waittime = network_listener_waittime_config[:tcp_port_25226]
+        if is_valid_number?(waittime)
```
ganga1980, line 320: We should have check on lower and upper bounds . For example, lower bound is half of the default and upper bound is ~ 3 to 4x . If the value outside lower and upper bounds, we can use the default.

kubernetes/linux/main.sh
```diff
@@ -1075,6 +1078,11 @@ else
       echo "not checking onboarding status (no metrics to scrape since MUTE_PROM_SIDECAR is true)"
 fi
 
+# Get the end time of the setup in seconds
+endTime=$(date +%s)
+elapsed=$((endTime-startTime))
+echo "setup script took: $elapsed seconds"
```
ganga1980, line 1084: probable we should rename from "setup" to main.sh or startup

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -189,6 +189,13 @@ data:
     # [agent_settings.proxy_config]
     #    ignore_proxy_settings = "true"  # if this is not applied, default value is false
 
+    # The following settings are "undocumented", we don't recommend uncommenting them unless directed by Microsoft.
+    # Configuration settings for the waittime for the network listeners to be available
+    # [agent_settings.network_listener_waittime] 
+    #   tcp_port_25226 = 45                           # Port 25226 is used by the fluent-bit in ReplicaSet
```
ganga1980, line 195: probable, we can improve the comment saying for example from "Port 25226 is used by the fluent-bit in ReplicaSet"  to "Port 25226 is used for telegraf to fluent-bit in ReplicaSet" or something like this being used between telegraf to fluent-bit/fluentd etc.

ganga1980 (approved):

PR #961, opened 2023-04-28
Updating the configmap changes for prometheus sidecar

jatakiajanvi12:


build/common/installer/scripts/tomlparser-prom-agent-config.rb
```diff
@@ -63,6 +72,28 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           puts "Using config map value: AZMON_SIDECAR_FBIT_MEM_BUF_LIMIT = #{@promFbitMemBufLimit.to_s + "m"}"
         end
       end
+
+      network_listener_waittime_config = parsedConfig[:agent_settings][:network_listener_waittime]
```
ganga1980, line 76: this change will not work as this configmap only invoked when prometheus side car. this needs to be some separate config file.

jatakiajanvi12: Prometheus side car is the only place where we missed. I think this works.

ganga1980: its my bad. I thought you removed prev code. I see this code there in both configs and should be good.

ganga1980 (approved):

PR #969, opened 2023-05-11
Gangams/agent release 3.1.8 tasks

ganga1980:
Fix environment variable collection for sandboxing pods (aka Kata runtime pods)


source/plugins/ruby/kubernetes_container_inventory.rb
pfrcks, line 1: looks ok to me but have you tested both kata and non kata scenarios?

ganga1980: kata 
image
non-kata
image

pfrcks (approved):

PR #980, opened 2023-05-31
remove vim package and add Mariner version

jatakiajanvi12:
- Removed vim package
- Also, added the libs for each package so if we want to remove any package, there are no dependencies. 
- Added Mariner version in the logs with other packages


kubernetes/linux/Dockerfile.multiarch
```diff
@@ -88,26 +85,25 @@ COPY --from=builder /lib/libreadline.so.8 /lib/
 COPY --from=builder /usr/lib/libncursesw.so.6 /usr/lib/libtinfo.so.6 /usr/lib/
 # inotifywait dependencies
 COPY --from=builder /lib/libinotifytools.so.0 /lib/
+COPY --from=builder /lib/libc.so.6 /lib/
```
ganga1980, line 88: is this new dependency for inotifywait?  and how did you detect this new dependency?

jatakiajanvi12: It is not a new dependency. While deleting the vim package dependency, I just wanted to make sure that we add all the dependencies for all packages. 

For example, libs.so.6 is a dependency for many packages. I missed copying it multiple times before since it was present for some package. But, I realised now that we might want to remove complete packages so that shouldn't affect other package dependencies. Just ran this: https://github.com/microsoft/Docker-Provider/blob/ci_prod/MARINER.md#get-dependencies-for-the-distroless-image

kubernetes/linux/Dockerfile.multiarch
```diff
@@ -88,26 +85,25 @@ COPY --from=builder /lib/libreadline.so.8 /lib/
...
 # inotifywait dependencies
 COPY --from=builder /lib/libinotifytools.so.0 /lib/
+COPY --from=builder /lib/libc.so.6 /lib/
 # crond dependencies
 COPY --from=builder /lib/libselinux.so.1 /lib/libpam.so.0 /lib/libc.so.6 /lib/libpcre.so.1 /lib/libaudit.so.1 /lib/libcap-ng.so.0/ /lib/
-# vim dependencies
-COPY --from=builder /lib/libm.so.6 /lib/libtinfo.so.6 /lib/
 # ruby dependencies
-COPY --from=builder /usr/lib/libruby.so.3.1 /usr/lib/libz.so.1 /usr/lib/libgmp.so.10 /usr/lib/libcrypt.so.1 /usr/lib/libm.so.6 /usr/lib/
+COPY --from=builder /usr/lib/libruby.so.3.1 /usr/lib/libz.so.1 /usr/lib/libgmp.so.10 /usr/lib/libcrypt.so.1 /usr/lib/libm.so.6 /usr/lib/libc.so.6 /usr/lib/
```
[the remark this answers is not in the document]
jatakiajanvi12: Same response as above

ganga1980 (approved):

PR #983, opened 2023-06-01
ICM reported: Arc ama-logs pods with docker container runtime are constantly restarting

jatakiajanvi12:
From the logs: 
```
configured container runtime on kubelet is : docker
set caps for ruby process to read container env from proc
k8s-b8p2162003301cl-m0
*** setting up oneagent in legacy auth mode ***
setting mdsd workspaceid & key for workspace:0498aefa-7893-4e0f-930a-b4647f18f5a5
starting mdsd in main container...
setting up cronjob for ci agent log rotation
*** starting fluentd v1 in daemonset
starting fluent-bit and setting telegraf conf file for daemonset
using fluentbitconf file: fluent-bit.conf for fluent-bit
Fluent Bit v2.0.9
* Copyright (C) 2015-2022 The Fluent Bit Authors
* Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
* https://fluentbit.io

**[2023/05/30 15:40:52] [error] could not open configuration file, aborting.**
```

Fix: 
image


ganga1980 (approved):

kubernetes/linux/main.sh
```diff
@@ -972,7 +972,7 @@ if [ ! -e "/etc/config/kube.conf" ]; then
             fi
             echo "using fluentbitconf file: ${fluentBitConfFile} for fluent-bit"
             if [ "$CONTAINER_RUNTIME" == "docker" ]; then
-                  fluent-bit -c /etc/opt/microsoft/docker-cimprov/${fluentBitConfFile}-e /opt/fluent-bit/bin/out_oms.so &
+                  fluent-bit -c /etc/opt/microsoft/docker-cimprov/${fluentBitConfFile} -e /opt/fluent-bit/bin/out_oms.so &
```
pfrcks, line 975: we should be using linters/static check tools to avoid these. shellcheck is a good one for bash and is available in most of the IDEs

ganga1980: > we should be using linters/static check tools to avoid these. shellcheck is a good one for bash and is available in most of the IDEs

@pfrcks , shellecheck seems good tool we should integrate. But, I dont see this being caught by tool. Are you able to catch this by any linter tool?

jatakiajanvi12: Will create a task for the same

pfrcks: you are right this won't be caught by shellcheck since this is a valid string in bash :(

PR #984, opened 2023-06-02
Gangams/release 3.1.9 tasks

ganga1980:
In case of AKS clusters with dual stacks, nodes can have both IPv4 and IPv6 address. This change picks only IPv4 address of the windows nodes which is used for the cAdvisor API calls.


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -340,7 +341,15 @@ def getWindowsNodes
                 nodeStatusAddresses = nodeStatus["addresses"]
                 if !nodeStatusAddresses.nil?
                   nodeStatusAddresses.each do |address|
-                    winNode[address["type"]] = address["address"]
+                    if !address["type"].nil? && address["type"].eql?("InternalIP")
+                      ipAddress = address["address"]
+                      # Pick Only IPv4 address in case of  dual stack
+                      if IPAddress.valid_ipv4?(ipAddress)
+                        winNode[address["type"]] = ipAddress
```
wanlonghenry, line 348: should a break added here once the ipv4 address found? otherwise the winNode[address["type"]] might be overwritten if the loop continues

ganga1980: > should a break added here once the ipv4 address found? otherwise the winNode[address["type"]] might be overwritten if the loop continues

Its not allowed to have 2 Internal IP address with IPv4. In case of the dual stack, there will be one InternalIP address with Ipv4 and one InternalIP address for IPv6.  Having  2 InternalIP address with IPv4 is not valid scenario.

wanlonghenry (approved):

PR #985, opened 2023-06-05
Muting telegraf for Windows if monitor_kubernetes_pods = false or N/A

jatakiajanvi12:
Scenarios tested: 
- No configmap  : Disable telegraf
- monitor_kubernetes_pods = false : Disable telegraf
- monitor_kubernetes_pods = true : Enable telegraf

MemoryWorkingSet changes for windows pods: 
image

CpuUsage changes for windows pods:
image


build/common/installer/scripts/tomlparser-prom-customconfig.rb
```diff
@@ -342,6 +346,13 @@ def populateSettingValuesFromConfigMap(parsedConfig)
                 file.close
                 puts "config::Successfully created telemetry file for prometheus sidecar"
               end
```
ganga1980, line 348: did you test all these scenarios ? 1. No configmap applied 2. all these settings - https://github.com/microsoft/Docker-Provider/blob/ci_prod/build/windows/installer/conf/telegraf.conf#L133

jatakiajanvi12: Tested scenario 1. 
For 2. I think is monitor_kubernetes_pods = false, then telegraf shouldn't be started right? Or is that not the case with Windows?  @ganga1980

build/common/installer/scripts/tomlparser-prom-customconfig.rb
```diff
@@ -47,6 +47,10 @@
 
 @disableRSTelegraf = false
 
+def get_command_windows(env_variable_name, env_variable_value)
```
pfrcks, line 50: function name is a little confusing? get_command_windows and it's doing set operations inside therefore not very readable. Maybe rename to something like construct_command_windows?

build/common/installer/scripts/tomlparser-prom-customconfig.rb
```diff
@@ -342,6 +346,13 @@ def populateSettingValuesFromConfigMap(parsedConfig)
                 file.close
                 puts "config::Successfully created telemetry file for prometheus sidecar"
               end
+            elsif !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp("windows") == 0
+              # Setting the environment variable for starting telegraf only when monitorKubernetesPods is true 
+              file = File.open("setpromenv.ps1", "w")
+              if !file.nil?
+                commands = get_command_windows("TELEMETRY_CUSTOM_PROM_MONITOR_PODS", monitorKubernetesPods)
```
pfrcks, line 353: can there be a scenario where monitorKubernetesPods is neither true or false (instead it is a malformed input). if it is possible then what will be outcome?

build/common/installer/scripts/tomlparser-prom-customconfig.rb
jatakiajanvi12: We have the same name in other toml parser files as well. Wanted to keep it consistent

pfrcks (approved):

PR #987, opened 2023-06-07
allow selective platform builds for linux

pfrcks:
Moves docker buildx platform values for linux builds to ADO pipeline variable.


jatakiajanvi12:
Should we update this in the readme as well?

jatakiajanvi12 (approved):

PR #994, opened 2023-06-08
Longw/containerlogv2 dcr 

wanlonghenry:
Longw/containerlogv2 fetch from DCR


wanlonghenry:
image

wanlonghenry:
image

wanlonghenry:
image

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterParam.json
```diff
@@ -14,6 +14,9 @@
     "workspaceRegion": {
       "value": "<workspaceRegion>"
     },
+    "enableContainerLogV2": {
+      "value": true
```
ganga1980, line 18: Since we are changing the default to ContainerLogV2 and we would need to update the Public docs to reflect this.

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
```diff
@@ -38,6 +38,12 @@
                 "description": "Array of default container insights streams"
             }
         },
+        "enableContainerLogV2": {
```
ganga1980, line 41: we would need to add this in Bicep, Terraform and Azure Policy templates also.

source/plugins/go/src/oms.go
```diff
@@ -1124,6 +1126,20 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 	}
 	DataUpdateMutex.Unlock()
 
+	if IsAADMSIAuthMode == true && !IsGenevaLogsIntegrationEnabled {
```
ganga1980, line 1129: 1. you would need to add check for !IsWindows as well since this specific to linux. updated condition will be  if IsAADMSIAuthMode == true && !IsGenevaLogsIntegrationEnabled && !IsWindows 
2. Move this logic before this line if ContainerLogSchemaV2 == true || ContainerLogsRouteADX == true {

source/plugins/go/src/extension/extension.go
```diff
@@ -91,3 +118,35 @@ func getDataTypeToStreamIdMapping() (map[string]string, error) {
 	}
 	return datatypeOutputStreamMap, nil
 }
+
+func (e *Extension) IsContainerLogV2() bool {
+	extensionconfiglock.Lock()
+	defer extensionconfiglock.Unlock()
+	if len(e.dataCollectionSettings) > 0 && e.dataCollectionSettings["enableContainerLogV2"] != "" {
+		message := fmt.Sprintf("isContainerLogV2: %s", e.dataCollectionSettings["enableContainerLogV2"])
+		logger.Printf(message)
```
ganga1980, line 127: This will be too much log messages since this log message for every record.  I think, you can remove the info log messages .

source/plugins/go/src/oms.go
```diff
@@ -1124,6 +1126,20 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+			Log("Error::mdsd::mdsd connection does not exist. re-connecting ...")
+			CreateMDSDClient(ContainerLogV2, ContainerType)
+			if MdsdMsgpUnixSocketClient == nil {
+				Log("Error::mdsd::Unable to create mdsd client. Please check error log.")
+				ContainerLogTelemetryMutex.Lock()
+				defer ContainerLogTelemetryMutex.Unlock()
+				ContainerLogsMDSDClientCreateErrors += 1
+				return output.FLB_RETRY
+			}
+			ContainerLogSchemaV2 = extension.GetInstance(FLBLogger, ContainerType).IsContainerLogV2()
```
ganga1980, line 1140: You would need to implement for IsContainerLogV2 for Windows also. you would need to update ingestion_token_utils.go for windows

source/plugins/go/src/extension/extension.go
```diff
@@ -23,66 +24,92 @@ var containerType string
...
+}
+
+func getExtensionSettings() (map[string]map[string]interface{}, error) {
+	extensionSettings := make(map[string]map[string]interface{})
+
+	extensionConfigs, err := getExtensionConfigs()
+	if err != nil {
+		return extensionSettings, err
+	}
+	for _, extensionConfig := range extensionConfigs {
```
ganga1980, line 80: we should log warn/error message if  length of extensionConfigs is greater than 1 since we are overriding the extensionSettings, right? do you know when this can have length greater than 1?

source/plugins/go/src/extension/extension.go
```diff
@@ -23,66 +24,92 @@ var containerType string
...
+	dataCollectionSettings := make(map[string]string)
+
+	extensionSettings, err := getExtensionSettings()
+	if err != nil {
+		return dataCollectionSettings, err
+	}
+	dataCollectionSettingsItr := extensionSettings["dataCollectionSettings"]
+	if dataCollectionSettingsItr != nil && len(dataCollectionSettingsItr) > 0 {
+		for k, v := range dataCollectionSettingsItr {
+			dataCollectionSettings[k] = fmt.Sprintf("%v", v)
```
ganga1980, line 100: we should convert the keys to lowercase to handle scenario where customer has different casing for example enablecontainerlogv2 or EnableContainerLogV2  instead of enableContainerLogV2.

source/plugins/go/src/extension/extension.go
wanlonghenry: this only has length greater than 1 if the other extensions are set together with us

source/plugins/go/src/oms.go
wanlonghenry: do we want to add it here or cover it when dealing with windows?

source/plugins/go/src/oms.go
wanlonghenry: changed

source/plugins/go/src/extension/extension.go
wanlonghenry: removed

source/plugins/go/src/extension/extension.go
wanlonghenry: converted both key and value and changed the corresponding place

source/plugins/go/src/oms.go
```diff
@@ -1153,6 +1155,22 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+				Log("Error::mdsd::mdsd connection does not exist. re-connecting ...")
+				CreateMDSDClient(ContainerLogV2, ContainerType)
+				if MdsdMsgpUnixSocketClient == nil {
+					Log("Error::mdsd::Unable to create mdsd client. Please check error log.")
+					ContainerLogTelemetryMutex.Lock()
+					defer ContainerLogTelemetryMutex.Unlock()
+					ContainerLogsMDSDClientCreateErrors += 1
+					return output.FLB_RETRY
+				}
+				ContainerLogSchemaV2 = extension.GetInstance(FLBLogger, ContainerType).IsContainerLogV2()
```
ganga1980, line 1170: This line (ContainerLogSchemaV2 = extension.GetInstance(FLBLogger, ContainerType).IsContainerLogV2())  needs to be outside of **if MdsdMsgpUnixSocketClient == nil {** condition

wanlonghenry: changed

wanlonghenry:
image

wanlonghenry:
image

ganga1980 (approved):

PR #999, opened 2023-06-19
Longw/containerlogv2  windows

wanlonghenry:
Longw/containerlogv2 windows fetch

PR #1004, opened 2023-06-23
Longw/containerlogv2 dcr merge

wanlonghenry:
cover scenario for containerlogv2 from config map


source/plugins/go/src/ingestion_token_utils.go
```diff
@@ -450,7 +450,7 @@ func getAgentConfiguration(imdsAccessToken string) (configurationId string, chan
 
 	configurationId = agentConfiguration.Configurations[0].Configurationid
 	channelId = agentConfiguration.Configurations[0].Content.Channels[0].ID
-	if len(agentConfiguration.Configurations[0].Content.Extensionconfigurations.Containerinsights) > 0 {
+	if !ContainerLogV2ConfigMap && len(agentConfiguration.Configurations[0].Content.Extensionconfigurations.Containerinsights) > 0 {
 		ContainerLogSchemaV2 = agentConfiguration.Configurations[0].Content.Extensionconfigurations.Containerinsights[0].Extensionsettings.DataCollectionSettings.EnableContainerLogV2
```
ganga1980, line 454: you should add nil & empty check

wanlonghenry: responded in this PR: https://github.com/microsoft/Docker-Provider/pull/999
Extensionsettings and DataCollectionSettings are embedded structs, not pointers to structs or slices/arrays. struct fields will be automatically initialized to their zero values
so if not set, the output of agentConfiguration.Configurations[0].Content.Extensionconfigurations.Containerinsights[0].Extensionsettings will be { { "" "" nil false } }

if we need access the other field in dcr settings then we need to do the check on those feilds

wanlonghenry:
s1: MSI cluster enable v2 with only containerlogv2: 
image
image

source/plugins/go/src/ingestion_token_utils.go
ganga1980: are you saying, if there is no extensionSettings or DataCollectionSettings in the DCR , this will  be return to false "agentConfiguration.Configurations[0].Content.Extensionconfigurations.Containerinsights[0].Extensionsettings.DataCollectionSettings.EnableContainerLogV2", is that right? Did you valid just enabling to MSI cluster without data collection settings?

wanlonghenry:
s2: msi with dcr only: 
image

wanlonghenry:
s3: dcr v2 with config map v2
image

wanlonghenry:
s4: without dcr setting in backdoor arm but with config map: 
image

source/plugins/go/src/ingestion_token_utils.go
wanlonghenry: tested with s4 and s5 attached the backdoor yaml screenshot without dcr settings

ganga1980 (approved):

PR #1007, opened 2023-06-29
adjust CI dev image tags

wanlonghenry:
adjust CI dev image tags

### DIFF
--- a/.pipelines/azure_pipeline_mergedbranches.yaml
+++ b/.pipelines/azure_pipeline_mergedbranches.yaml
@@ -33,11 +33,12 @@ jobs:
   ## prod release are in format Major.Minor.Patch
   ## preview release are in format preview-Major.Minor.Patch
   - bash: |
-      commit=$(git describe) 
-      linuxImagetag="$commit"
-      windowsImageTag=win-"$commit"
+      commit=$(git describe | cut -d'-' -f1,2)
+      datetime=$(date +%Y%m%d%H%M%S)
+      linuxImagetag="$commit"-"$datetime"
+      windowsImageTag=win-"$commit"-"$datetime"
 
-      telemetryTag=$commit
+      telemetryTag=$linuxImagetag
       if [ -z "$TELEMETRY_TAG" ]
         then
           echo "\$TELEMETRY_TAG variable is not set"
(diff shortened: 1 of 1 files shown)


wanlonghenry:
Triggered build with the same changes  both from automated and manually:
image
image
image

.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -33,11 +33,12 @@ jobs:
...
-      commit=$(git describe) 
-      linuxImagetag="$commit"
-      windowsImageTag=win-"$commit"
+      commit=$(git describe | cut -d'-' -f1,2)
+      datetime=$(date +%Y%m%d%H%M%S)
+      linuxImagetag="$commit"-"$datetime"
+      windowsImageTag=win-"$commit"-"$datetime"
 
-      telemetryTag=$commit
+      telemetryTag=$linuxImagetag
```
jatakiajanvi12, line 41: I think we should keep both the commit id and datetime.

.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -33,11 +33,12 @@ jobs:
   ## prod release are in format Major.Minor.Patch
   ## preview release are in format preview-Major.Minor.Patch
   - bash: |
-      commit=$(git describe) 
-      linuxImagetag="$commit"
-      windowsImageTag=win-"$commit"
+      commit=$(git describe | cut -d'-' -f1,2)
```
ganga1980: why are we removing the commit id? we can keep commit id and add timestamp onto that, right?

wanlonghenry: updated, keep both id and date

.pipelines/azure_pipeline_mergedbranches.yaml
wanlonghenry: updated, keep both id and date

wanlonghenry:
triggered the build with same change:
image
image

ganga1980 (approved):

jatakiajanvi12 (approved):

PR #1024, opened 2023-07-18
Improved livenessprobe for MonAgentManager.exe and MonAgentLauncher.exe is now executed in the background 

cacarlt:


kubernetes/host-logs-geneva.yaml
```diff
@@ -56,10 +56,13 @@ spec:
        livenessProbe:
           exec:
             command:
-              - cmd
-              - /c
-              - .\opt\hostlogswindows\scripts\cmd\livenessprobe.exe
-              - ".\\etc\\hostlogswindows\\filesystemwatcher.txt"
+              - C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
```
chris-raitano: I'm a little hesitant to use a hardcoded path for anything we didn't install ourselves (especially when the path includes a version number). Do we know for certain that powershell will always be at this location on every node?

kubernetes/host-logs-geneva.yaml
```diff
@@ -56,10 +56,13 @@ spec:
...
+              - -command
+              - |
+                .\opt\hostlogswindows\scripts\powershell\livenessprobe.ps1 -exeRelativePath ".\\opt\\genevamonitoringagent\\genevamonitoringagent\\Monitoring\Agent\\MonAgentManager.exe" -fileName ".\\etc\\hostlogswindows\\filesystemwatcher.txt"
+            initialDelaySeconds: 60
```
chris-raitano: Extra whitespace here causing an error. InitialDelaySeconds, periodSeconds, and timeoutSeconds should be in line with exec

kubernetes/host-logs-geneva.yaml
```diff
@@ -56,10 +56,13 @@ spec:
...
           exec:
             command:
-              - cmd
-              - /c
-              - .\opt\hostlogswindows\scripts\cmd\livenessprobe.exe
-              - ".\\etc\\hostlogswindows\\filesystemwatcher.txt"
+              - C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
+              - -command
+              - |
+                .\opt\hostlogswindows\scripts\powershell\livenessprobe.ps1 -exeRelativePath ".\\opt\\genevamonitoringagent\\genevamonitoringagent\\Monitoring\Agent\\MonAgentManager.exe" -fileName ".\\etc\\hostlogswindows\\filesystemwatcher.txt"
```
chris-raitano: Need a 2nd slash between 'Monitoring\\Agent'

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,27 @@
...
+$FILESYSTEM_WATCHER_FILE_EXISTS=1
+$NO_MONAGENT_MANAGER_PROCESS=2
+
+if (Test-Path -Path $fileName -PathType Leaf) 
+{
+    Write-Error "INFO: File: $filename exists indicates Config Map Updated since agent started."
+    exit $FILESYSTEM_WATCHER_FILE_EXISTS
+}
+
+if(-not (Get-WmiObject Win32_Process -Filter "ExecutablePath LIKE '%$exeRelativePath'"))
```
chris-raitano: If there is no config map yet on the host then we don't start the monitoring agent. In this case the liveness probe is still checking for the MA even though it was never started – which causes an infinite crash loop until the config map is applied

kubernetes/host-logs-geneva.yaml
chris-raitano: the exeRelativePath isn't working for me unless I remove ".\\\\" at the start

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,27 @@
...
+{
+    Write-Error "ERROR: Process not running: $exeFullPath"
```
chris-raitano: $exeFullPath isn't defined

chris-raitano (changes requested):

kubernetes/host-logs-geneva.yaml
cacarlt: Got this from the hpc example. I can take it out and just do powershell.exe.

kubernetes/host-logs-geneva.yaml
cacarlt: Done

kubernetes/host-logs-geneva.yaml
cacarlt: Fixed

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Added a flag for when we start the MonAgentLanucher so that the check doesn't fail while we are waiting for the configmap to get deployed.

kubernetes/host-logs-geneva.yaml
cacarlt: Fixed

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Fixed

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,30 @@
+param(
+    [Parameter( Position=0,
+                mandatory=$true,
+                HelpMessage="Path of the exeutable you want to check if it is running. Be sure to provide '\\' when entering the path.")]
```
alrodrig1: Check for typos (ie exeutable).

alrodrig1: For consistency, can we have both paths follow the same format? It's confusing having one parameter require double backslashes while the other parameter doesn't have this requirement. The script could take care of escaping `\` when needed.

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,30 @@
...
+
+    [Parameter( Position=1,
+                mandatory=$true,
+                HelpMessage="Relative path of the file you want to check if it exists.")] 
+    [string]$fileName
+)
+
+$SUCCESS=0
+$FILESYSTEM_WATCHER_FILE_EXISTS=1
+$NO_MONAGENT_MANAGER_PROCESS=2
```
alrodrig1: There seems to be a discrepancy between the specificity of the parameters vs the actual script. The parameters refer to generic "executable" and "file" while the script refers to these as the more concrete "Mon Agent" and "file watcher file".

For consistency, either make it all generic or all specific. Given that the probe is not intended to be reused, maybe keeping it all specific is the most useful option.

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,30 @@
...
+
+if (Test-Path -Path $fileName -PathType Leaf) 
+{
+    Write-Error "INFO: File: $filename exists indicates Config Map Updated since agent started."
```
alrodrig1: Could you provide a comment to clarify why the existence of this file would cause the liveness probe to fail?

Is the intent to restart the pod so that it can load the updated ConfigMap? If that's the case, this unconventional use of a liveness probe should be clearly documented both here and in the design spec.

If this design choice aligns with the pattern previously set by CI in the current liveness probes, it makes sense to maintain consistency. Otherwise, this design decision should be reconsidered.

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +86,19 @@ Set-EnvironmentVariables
...
     Write-Host "Geneva not configured. Watching for config map"
-    # Infinite loop keeps container alive while waiting for config map
-    # Otherwise when the process ends, kubernetes sees this as a crash and the container will enter a crash loop
-    while($true){
-        Start-Sleep 3600
-    }
 }
 
+
+Notepad.exe | Out-Null
```
alrodrig1, line 65: Looks like you were testing something with notepad.exe and forgot to delete this line?

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +86,19 @@ Set-EnvironmentVariables
...
 
 if(Get-GenevaEnabled){
-    Invoke-Expression ".\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe -useenv"
+    Write-Host "Starting Windows AMA in 1P Mode"
+
+    Start-Job -Name "WindowsHostLogsJob" -ScriptBlock { 
+        Start-Process -NoNewWindow -FilePath ".\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe" -ArgumentList @("-useenv")
+    }
+
+    [System.Environment]::SetEnvironmentVariable("HOSTLOGS_MA_STARTED", "true", "process")
```
alrodrig1: This env var is being limited to the process scope. Will it be visible from the liveness probe as it runs as a separate process?

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,30 @@
...
+    exit $FILESYSTEM_WATCHER_FILE_EXISTS
+}
+
+if($Env:HOSTLOGS_MA_STARTED -eq "true")
```
alrodrig1: Could you explain the purpose of this environment variable? If it's not set due to an exception in main.ps1 or some other reason, the probe will always succeed, which I'm assuming is not our intended outcome.

kubernetes/host-logs-geneva.yaml
```diff
@@ -34,12 +34,12 @@ spec:
     hostNetwork: true
     containers:
      - name: host-logs-windows
-       image: "VALUE_IMAGE_NAME"
+       image: "VALUE_CONTAINER_IMAGE"
```
alrodrig1: where is this used?

kubernetes/host-logs-geneva.yaml
```diff
@@ -34,12 +34,12 @@ spec:
...
        imagePullPolicy: IfNotPresent
        resources:
         limits:
-         cpu: "1"
-         memory: 2Gi
+         cpu: "0.5"
```
alrodrig1: are these resource limits for testing only or are they used in the final deployment?

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,30 @@
...
+                HelpMessage="Path of the exeutable you want to check if it is running. Be sure to provide '\\' when entering the path.")] 
+    [string]$exeRelativePath,
+
+    [Parameter( Position=1,
+                mandatory=$true,
+                HelpMessage="Relative path of the file you want to check if it exists.")] 
+    [string]$fileName
+)
+
+$SUCCESS=0
```
alrodrig1: Can't see this variable being used anywhere, is it necessary?

build/hostlogswindows/installer/scripts/livenessprobe.ps1
alrodrig1: check naming consistency, e.g., $file**n**ame vs $file**N**ame

build/hostlogswindows/installer/scripts/livenessprobe.ps1
```diff
@@ -0,0 +1,30 @@
...
+
+if (Test-Path -Path $fileName -PathType Leaf) 
+{
+    Write-Error "INFO: File: $filename exists indicates Config Map Updated since agent started."
+    exit $FILESYSTEM_WATCHER_FILE_EXISTS
+}
+
+if($Env:HOSTLOGS_MA_STARTED -eq "true")
+{
+    if(-not (Get-WmiObject Win32_Process -Filter "ExecutablePath LIKE '%$exeRelativePath'"))
```
alrodrig1: Did you compare performance between `Get-WmiObject` and the newer `Get-CimInstance`? The newer cmdlet is supposed to be more efficient.

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +86,19 @@ Set-EnvironmentVariables
 Start-FileSystemWatcher
 
 if(Get-GenevaEnabled){
-    Invoke-Expression ".\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe -useenv"
+    Write-Host "Starting Windows AMA in 1P Mode"
+
+    Start-Job -Name "WindowsHostLogsJob" -ScriptBlock {
```
alrodrig1, line 55: Doesn't `Start-Process -NoNewWindow` starts this process in the background without needing Start-Job? Also, would Start-Job require eventual clean-up of completed jobs?

alrodrig1 (changes requested):

kubernetes/windows/hostlogs/main.ps1
cacarlt: Nope this keeps the container alive similar to how CI does it now: 

https://github.com/microsoft/Docker-Provider/blob/ca2b6783558a9b86d4dd28b4bfb895d95a3776ad/kubernetes/windows/main.ps1#L588

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: This follows the current design pattern of how CI resets the container when the configmap is updated: https://github.com/microsoft/Docker-Provider/blob/ca2b6783558a9b86d4dd28b4bfb895d95a3776ad/build/windows/installer/livenessprobe/livenessprobe.cpp#L53

The script creating this file is here:
https://github.com/microsoft/Docker-Provider/blob/ci_prod/build/windows/installer/scripts/filesystemwatcher.ps1

kubernetes/host-logs-geneva.yaml
cacarlt: This file is for only development and testing, but the values used here should be in line with the values we use inside AKS RP.

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Ah good catch we should exit success if we reach to end of the script.

kubernetes/windows/hostlogs/main.ps1
cacarlt: This is how CI starts the monitoring agent:

https://github.com/microsoft/Docker-Provider/blob/ca2b6783558a9b86d4dd28b4bfb895d95a3776ad/kubernetes/windows/main.ps1#L570

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Without it the livenessprobe thinks the container crashed because the exe doesn't execute unless the environment variables already exist. So instead, we first check to see if we should even expect the exe to up yet. 

If main.ps1 fails in anyway the container will crash since it is the entrypoint for the container image.

kubernetes/windows/hostlogs/main.ps1
cacarlt: I thought I tested that, but I think I need to just exec into the container and the host to check.

kubernetes/host-logs-geneva.yaml
cacarlt: It is just a placeholder for the container image. When we merge into ci_prod we expect to have the actual image added like so: https://github.com/microsoft/Docker-Provider/blob/ca2b6783558a9b86d4dd28b4bfb895d95a3776ad/kubernetes/ama-logs.yaml#L926

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Nope worth looking into though

kubernetes/windows/hostlogs/main.ps1
alrodrig1: fair enough!

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Done

cacarlt: Done

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Done

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Done

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Done

kubernetes/windows/hostlogs/main.ps1
cacarlt: When I run a Get-Job on the host after deleting the DaemonSet the Job doesn't exist.

build/hostlogswindows/installer/scripts/livenessprobe.ps1
alrodrig1: How about tweaking `initialDelaySeconds` in the liveness probe setup to give the MA enough start time? Relying on a global env var at the host level might get tricky, particularly if it sticks around after the container's life ends.

chris-raitano: > How about tweaking `initialDelaySeconds` in the liveness probe setup to give the MA enough start time? Relying on a global env var at the host level might get tricky, particularly if it sticks around after the container's life ends.

How long is long enough for initialDelaySeconds though? On a node which is extremely low on resources the MA could take a while to start up. If the delay is not long enough we could end up repeatedly crashing a healthy pod. But we also don't want to wait too long to restart an unhealthy pod

kubernetes/host-logs-geneva.yaml
```diff
@@ -56,10 +56,13 @@ spec:
...
           exec:
             command:
-              - cmd
-              - /c
-              - .\opt\hostlogswindows\scripts\cmd\livenessprobe.exe
-              - ".\\etc\\hostlogswindows\\filesystemwatcher.txt"
+              - C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
+              - -command
+              - |
+                .\opt\hostlogswindows\scripts\powershell\livenessprobe.ps1 -exeRelativePath "\\opt\\genevamonitoringagent\\genevamonitoringagent\\Monitoring\\Agent\\MonAgentCore.exe" -fileName "etc\hostlogswindows\filesystemwatcher.txt"
```
chris-raitano: Should we watch MonAgentLauncher instead? MonAgentLauncher will spin up multiple instances of MonAgentCore and has built-in health checks for these. If all instances of MonAgentCore crash, MonAgentLauncher should automatically spin up new instances

build/hostlogswindows/installer/scripts/livenessprobe.ps1
chris-raitano: Once we have custom users in containerd 1.7, a user level variable would work well here. The plan is to create a new admin user for each pod, so a user level variable would be available to all processes in the pod, but should be cleaned up when the pod dies

kubernetes/windows/hostlogs/main.ps1
chris-raitano: As mentioned in my [other comment](https://github.com/microsoft/Docker-Provider/pull/1024/files#r1270059764), I think "user" is the best level to put this at. Until we have containerd 1.7 we can set it to false in setup.ps1 to make sure it's cleared before restarting the container

chris-raitano: nitpick

$true instead of "true". Then we can use boolean operators instead of string comparison to check it

build/hostlogswindows/installer/scripts/livenessprobe.ps1
chris-raitano: > we should exit success if we reach to end of the script

That's the default behavior, so we're already doing it 😃

build/hostlogswindows/installer/scripts/livenessprobe.ps1
vtelleria: Are slow starts something we have observed or is this speculation? I would expect the agent to become present in the process list fairly quickly even if it is taking a while to startup completely, no? 

In any case, we should have an understanding of a worst-case startup time and have a timeout somewhere. The current implementation doesnt protect against long startup times anyway since the monitored process is started in the background and the env variable is set immediately (even potentially before the process actually starts). 

If we want to guard against an issue with slow starts and the process not becoming viewable in the process list before the livenessProbe kicks in, consider using a startupProbe with the same check: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

kubernetes/windows/hostlogs/main.ps1
vtelleria: can we add a comment describing intent? This is head scratching at first glance.

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +86,19 @@ Set-EnvironmentVariables
...
+    Write-Host "Starting Windows AMA in 1P Mode"
+
+    Start-Job -Name "WindowsHostLogsJob" -ScriptBlock { 
+        Start-Process -NoNewWindow -FilePath ".\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe" -ArgumentList @("-useenv")
+    }
+
+    [System.Environment]::SetEnvironmentVariable("HOSTLOGS_MA_STARTED", "true", "process")
+
 } else {
     Write-Host "Geneva not configured. Watching for config map"
```
vtelleria: How is this watching for the config map? Or is this referencing something happening elsewhere?

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +86,19 @@ Set-EnvironmentVariables
 Start-FileSystemWatcher
 
 if(Get-GenevaEnabled){
```
vtelleria: Get-GenevaEnabled isn't so much checking for enablement, but for a valid and complete configuration. We should likely include an output which describes why this check failed so users can easily see "oh I need to set X"

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: To follow the same practices CI has set in livenessprobe.cpp we should retrun a SUCCESS in the code. At the very least to indicate reaching the end is what we want for a healthy container probe.

kubernetes/windows/hostlogs/main.ps1
alrodrig1, line 1: Is there a reason why in `Set-EnvironmentVariables` the _gets_ use the process scope while the _sets_ use the machine scope?

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: I haven't encountered slow starts with this setup and the initalDelaySeconds at 60 should help with that. If MonAgentLauncher.exe doesnt start up by that point then we should reset the container. 

The startupprobe could check to see if the executable is up and running but we run into the same issue if the configmap isn't added to the cluster yet then it isn't going to be return healthy. Unlike the livenessprobe the startupprobe only runs once so the customer only get once chance to get it on the cluster before it restart the container.

kubernetes/windows/hostlogs/main.ps1
cacarlt: filesystemwatcher.ps1 is doing that: https://github.com/microsoft/Docker-Provider/blob/sky-dev/build/hostlogswindows/installer/scripts/filesystemwatcher.ps1

kubernetes/windows/hostlogs/main.ps1
cacarlt: Same process used by CI: https://github.com/microsoft/Docker-Provider/blob/cae75f9854df1842b07886a7bbb3a5ee065945eb/kubernetes/windows/hostlogs/main.ps1#L67

cacarlt: The ruby parser file already does this and follows how CI currently operates with configmap changes.

https://github.com/microsoft/Docker-Provider/blob/sky-dev/build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb

kubernetes/windows/hostlogs/main.ps1
cacarlt: Done

cacarlt: From what I am seeing this this environment variable isn't seen by the livenessprobe 

https://github.com/microsoft/Docker-Provider/blob/cae75f9854df1842b07886a7bbb3a5ee065945eb/kubernetes/windows/hostlogs/main.ps1#L67

cacarlt: As a process env var it will get cleaned up off the container when it is no longer there. As a machine env var then it would stay on the host even if the container crashes/restartes/vacates.

kubernetes/host-logs-geneva.yaml
cacarlt: Yes good point, if MonAgentLauncher isnt there nothing is running.

MonAgentLauncher triggers MonAgentHost

MonAgentHost keeps MonAgentManger and MonAgentCore running.

MonAgentManager just check resoruce usage of MonAgentCore.

https://eng.ms/docs/products/geneva/collect/references/agent/monagenthost

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Changed to CmiInstance running tests with it to make sure it still works

build/hostlogswindows/installer/scripts/livenessprobe.ps1
cacarlt: Instead of using a flag I was able to use placeholder values for the needed env vars and execute the monagent with those. I tested to see

1) Does the container restart when I stop the process - yes

2) Does the container use the new env vars when the config map is changed - yes

3) Does the env vars go away on reset - yes

kubernetes/windows/hostlogs/main.ps1
cacarlt: Done

kubernetes/host-logs-geneva.yaml
```diff
@@ -56,10 +56,13 @@ spec:
...
-              - cmd
-              - /c
-              - .\opt\hostlogswindows\scripts\cmd\livenessprobe.exe
-              - ".\\etc\\hostlogswindows\\filesystemwatcher.txt"
+              - C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
+              - -command
+              - |
+                .\opt\hostlogswindows\scripts\powershell\livenessprobe.ps1 -monAgentManagerExePath "\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentManager.exe" -fileSystemWatcherTextFilePath "etc\hostlogswindows\filesystemwatcher.txt"
+          initialDelaySeconds: 60
+          periodSeconds: 180
```
vtelleria: Per offline conversation these two values should be swapped so the probe period isn't 3 minutes (resulting in a total of 9 minutes before failure would be detected): https://github.com/microsoft/Docker-Provider/blob/ca2b6783558a9b86d4dd28b4bfb895d95a3776ad/kubernetes/ama-logs.yaml#L1007

kubernetes/windows/hostlogs/main.ps1
cacarlt: For which line?

alrodrig1: In the function that starts at line 7.

I know this is a pattern that was copied from the existing CI solution. But was curious why it gets from 'process' and then sets to 'machine', e.g.:

```powershell
 $aksResourceId = [System.Environment]::GetEnvironmentVariable("AKS_RESOURCE_ID", "process") # <-- Here.
    if (![string]::IsNullOrEmpty($aksResourceId)) {
        [System.Environment]::SetEnvironmentVariable("AKS_RESOURCE_ID", $aksResourceId, "machine") # <-- and here.
```

cacarlt: I believe this is carried over from how CI currently does this specific setup.

cacarlt: If the machine doesn't already have this value set on the machine, then we will set it on the process.

cacarlt: https://github.com/microsoft/Docker-Provider/blob/8e78865361a6c18f08018dc9f2131dd2006574e9/kubernetes/windows/main.ps1#L268

build/hostlogswindows/installer/scripts/livenessprobe.ps1
chris-raitano: I think there's 2 different problems we're trying to solve here.

1. If AMA takes a while to start up it could (theoretically) trigger a liveness probe crash – I think the startup probe is probably the right solution to this
2. Previously we were not starting the monitoring agent if it had not been configured via ConfigMap. The env var was being used to determine this, otherwise we'd always get an infinite crash loop until the ConfigMap was provided – Adding placeholder values should address this

So maybe we should do both placeholder values + startup probe

kubernetes/windows/hostlogs/main.ps1
cacarlt: Removed all sets to the machine

kubernetes/windows/hostlogs/main.ps1
cacarlt: Starting up MonAgent with placeholder values until config map is applied

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
...
 {
-    if (argc < 2)
+    try 
     {
-        wprintf_s(L"ERROR:unexpected number arguments and expected is 5");
-        return UNEXPECTED_ERROR;
-    }
+        if (argc < 3) 
+        {  
+            wprintf_s(L"ERROR:unexpected number arguments and expected is 5");
```
chris-raitano: Update error message. We expect 2 args

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
...
+
+        if (!IsMonAgentRunning(argv[1])) 
+        {
+            wprintf_s(L"ERROR:Process:%s is not running\n", argv[1]);
+            return NO_MONAGENT_LAUNCHER_PROCES;
+        }
+
+        if (IsFileExists(argv[2]))
+        {
+            wprintf_s(L"INFO:File:%s exists indicates Config Map Updated since agent started.\n", argv[2]);
```
chris-raitano: nitpick - fix capitalization
Updated -> updated
Config Map -> either config map or ConfigMap

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
...
+bool IsFileExists(const wchar_t* const fileName)
 {
     DWORD dwAttrib = GetFileAttributes(fileName);
     return dwAttrib != INVALID_FILE_SIZE;
 }
 
+/*
+* Function to check if a process is running based on its full path
+*/ 
+bool IsMonAgentRunning(const wchar_t* executableFullPath) {
```
chris-raitano: This function could be used to check any process. Maybe a more generic name like IsProcessRunning

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Spelling carried over from https://github.com/microsoft/Docker-Provider/blob/290f4acbecdf18742781f539c3fb400eafd8c183/build/windows/installer/livenessprobe/livenessprobe.cpp#L126

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: C++ always puts livenessprobe.exe at argv[0]

https://github.com/microsoft/Docker-Provider/blob/290f4acbecdf18742781f539c3fb400eafd8c183/build/windows/installer/livenessprobe/livenessprobe.cpp#L106

https://github.com/microsoft/Docker-Provider/blob/290f4acbecdf18742781f539c3fb400eafd8c183/build/windows/installer/livenessprobe/livenessprobe.cpp#L110

https://github.com/microsoft/Docker-Provider/blob/290f4acbecdf18742781f539c3fb400eafd8c183/kubernetes/ama-logs.yaml#L997-L1006


Good catch for the ERROR message. Fixed.

kubernetes/windows/hostlogs/main.ps1
chris-raitano: > Same process used by CI:
> 
> https://github.com/microsoft/Docker-Provider/blob/cae75f9854df1842b07886a7bbb3a5ee065945eb/kubernetes/windows/hostlogs/main.ps1#L67

That's our file. 
CI sets the GENEVA_LOGS_INTEGRATION env var in the parser https://github.com/microsoft/Docker-Provider/blob/901bbb0895c81fccbcfa68bafad2ee366dafdd4a/build/common/installer/scripts/tomlparser-geneva-config.rb#L275
Then they just check that var in main https://github.com/microsoft/Docker-Provider/blob/901bbb0895c81fccbcfa68bafad2ee366dafdd4a/kubernetes/windows/main.ps1#L739

I think Get-GenevaConfigured would be a better name for this. Or we could switch to doing it in the parser to match CI

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Renamed to IsProcessRunningWithSpecifcPath

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +58,30 @@ Set-EnvironmentVariables
...
+
+    [System.Environment]::SetEnvironmentVariable("MONITORING_GCS_ENVIRONMENT", "Test" ,"Process")
+    [System.Environment]::SetEnvironmentVariable("MONITORING_GCS_ACCOUNT", "PLACEHOLDER", "Process")
+    [System.Environment]::SetEnvironmentVariable("MONITORING_GCS_NAMESPACE", "PLACEHOLDER", "Process")
+    [System.Environment]::SetEnvironmentVariable("MONITORING_CONFIG_VERSION", "PLACEHOLDER", "Process")
+    [System.Environment]::SetEnvironmentVariable("MONITORING_MANAGED_ID_IDENTIFIER", "PLACEHOLDER", "Process")
+    [System.Environment]::SetEnvironmentVariable("MONITORING_MANAGED_ID_VALUE", "PLACEHOLDER", "Process")
+
+
+    Start-Job -Name "WindowsHostLogsJob" -ScriptBlock {
```
alrodrig1, line 55: It seems the agent is initiated even if Geneva isn't configured. Is this by design? Also, since it starts in both the `if true` and `else` blocks, it can be moved outside the conditional.

kubernetes/windows/hostlogs/main.ps1
```diff
@@ -86,13 +58,30 @@ Set-EnvironmentVariables
...
+
 } else {
-    Write-Host "Geneva not configured. Watching for config map"
-    # Infinite loop keeps container alive while waiting for config map
-    # Otherwise when the process ends, kubernetes sees this as a crash and the container will enter a crash loop
-    while($true){
-        Start-Sleep 3600
+    Write-Host "Geneva not configured. Watching for config map."
+
+    [System.Environment]::SetEnvironmentVariable("MONITORING_GCS_ENVIRONMENT", "Test" ,"Process")
```
alrodrig1: Why do these variables need placeholder values? maybe be worth clarifying with a comment.

Also, if I'm not mistaken, since the process scope is the default, `$Env:MYVAR = "something"` is functionally the same to this code while being more idiomatic in PowerShell.

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
...
-        wprintf_s(L"ERROR:unexpected number arguments and expected is 5");
-        return UNEXPECTED_ERROR;
-    }
+        if (argc < 3) 
+        {  
+            wprintf_s(L"ERROR:unexpected number arguments and expected is 3");
+            return UNEXPECTED_ERROR;
+        }
+
+        if (!IsProcessRunningWithSpecifcPath(argv[1]))
```
alrodrig1: typo: `Specific`

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
 #endif
 
 #include <Windows.h>
-#include <tchar.h>
+#include <tlhelp32.h>
+#include <iostream>
 
 #define SUCCESS 0x00000000
-#define FILESYSTEM_WATCHER_FILE_EXISTS 0x00000001
+#define NO_MONAGENT_LAUNCHER_PROCES 0x00000001
```
alrodrig1: typo: `PROCESS`

build/hostlogswindows/installer/startupprobe/startupprobe.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <iostream>
+
+#define SUCCESS 0x00000000
+#define NO_MONAGENT_LAUNCHER_PROCES 0x00000001
+#define UNEXPECTED_ERROR 0xFFFFFFFF
+
+/*
+* Function to check if a process is running based on its full path
+*/ 
+bool IsProcessRunningWithSpecifcPath(const wchar_t* executableFullPath) {
```
alrodrig1: One way to avoid repeating code could be to merge both probes into a single executable, and use a parameter to decide the desired behavior. If the only difference between the startup and liveness probes is that the startup one doesn't validate the filesystem watcher, then exe could be invoked as:

```
# Startup
hostlogsprobe.exe -validateAgent "\path\to\agent\"

# Liveness
hostlogsprobe.exe -validateAgent "\path\to\agent\" -validateFilesystemWatcher "\path\to\filesystemwatcher"
```

kubernetes/host-logs-geneva.yaml
```diff
@@ -1,85 +1,108 @@
...
+            periodSeconds: 60
+            timeoutSeconds: 30
+            successThreshold: 1
+            failureThreshold: 3
+            exec:
+              command:
+                - cmd
+                - /c
+                - .\opt\hostlogswindows\scripts\cmd\startupprobe.exe
+                - >-
```
alrodrig1: It might be worth trying to enclose the path in double quotes to avoid any YAML parsing difficulties caused by `%`. Double quotes tend to be more straightforward and recognizable than the folded block scalar style `- >-`, which could be confusing for people less experienced with YAML syntax.

```
- "%ENV_VAR%\agent.exe"
```

vs

```
- >-
 %ENV_VAR%\agent.exe
```

alrodrig1 (changes requested):

kubernetes/host-logs-geneva.yaml
```diff
@@ -1,85 +1,108 @@
...
+            - name: HOSTNAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+          volumeMounts:
+            - mountPath: \etc\config\settings
+              name: settings-vol-config
+              readOnly: true
+          startupProbe:
+            initialDelaySeconds: 180
```
chris-raitano: 180 seconds might be a bit long for the initial delay. I wouldn't expect it to take that long on a normal startup. The sooner we detect that it has started, the sooner we can switch to the liveness probe and start watching for config map changes.

And maybe increase the failure threshold to account for this change, so we don't fail after 3 minutes if it does take a while to start up

kubernetes/host-logs-geneva.yaml
```diff
@@ -1,85 +1,108 @@
...
+          livenessProbe:
+            exec:
+              command:
+                - cmd
+                - /c
+                - .\opt\hostlogswindows\scripts\cmd\livenessprobe.exe
+                - >-
+                  %CONTAINER_SANDBOX_MOUNT_POINT%\opt\genevamonitoringagent\genevamonitoringagent\Monitoring\Agent\MonAgentLauncher.exe
+                - .\etc\hostlogswindows\filesystemwatcher.txt
+            initialDelaySeconds: 180
```
chris-raitano: We probably don't need an initial delay on the liveness probe anymore since the startup probe fulfills the same purpose

kubernetes/windows/hostlogs/main.ps1
cacarlt: By design since we want to use placeholder values only when the configmap isnt set.

kubernetes/windows/hostlogs/main.ps1
cacarlt: "Also, if I'm not mistaken, since the process scope is the default, $Env:MYVAR = "something" is functionally the same to this code while being more idiomatic in PowerShell. " 

Not sure what you are to trying say here.

build/hostlogswindows/installer/startupprobe/startupprobe.cpp
chris-raitano: Is there any way we could share this function with the liveness probe rather than maintaining 2 copies of the same thing?

Or if we could use the same exe for both, that would be even better

kubernetes/host-logs-geneva.yaml
cacarlt: The default failure threshold is 3 so I saw that was a good starting point as we haven't hit the issue after x amount of time at startup. If we encounter this being a problem, we can always bump this up.

The initalDelaySeconds is carried over from the liveness probe and since ama-logs-windows waits 180 seconds to check anything.

kubernetes/host-logs-geneva.yaml
cacarlt: I wanted to give it 60 seconds so that it can give a minute after the startup probe finished.

build/hostlogswindows/installer/startupprobe/startupprobe.cpp
cacarlt: +1 to a thread Alberto made before this. Closing this as a duplication.

cacarlt: @chris-raitano

kubernetes/windows/hostlogs/main.ps1
alrodrig1: I'm suggesting that we could use `$Env:MONITORING_MANAGED_ID_IDENTIFIER = "PLACEHOLDER"` instead of `[System.Environment]::SetEnvironmentVariable("MONITORING_MANAGED_ID_IDENTIFIER", "PLACEHOLDER", "Process")` since they both achieve the same result, but the former is more typical in PowerShell and less convoluted.

build/hostlogswindows/installer/startupprobe/startupprobe.cpp
alrodrig1: By the way, the point isn't strictly about merging the probes. You could also avoid this repetition by pulling out the common code into a standalone file and including it in both liveness and startup probes. It's important to avoid code duplication, as it causes unnecessary technical debt. For instance, a bug now has to be fixed in two spots, and it's highly likely that someone tweaking the code later might overlook one of them.

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
...
 
+/*
+* Function to check if a process is running based on its full path
+*/ 
+bool IsProcessRunningWithSpecifcPath(const wchar_t* executableFullPath) {
+    PROCESSENTRY32 processEntry{};  // struct to hold process details
+    processEntry.dwSize = sizeof(PROCESSENTRY32);  // set size of struct
+
+    // Handle to snapshot of all processes currently running on the system
+    HANDLE processSnapshot = CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0);
```
alrodrig1, line 39: I see in the documentation (https://learn.microsoft.com/en-us/windows/win32/api/tlhelp32/nf-tlhelp32-createtoolhelp32snapshot#return-value) that this function will return `INVALID_HANDLE_VALUE` if something goes wrong. So, you can do something like this after calling it:

```
    if (processSnapshot == INVALID_HANDLE_VALUE) {
        throw std::runtime_error("Failed to create module snapshot"); // or just exit the program with UNEXPECTED_ERROR ?
    }
```

Same idea applies to the module snapshot below.

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,87 @@
...
+
+    // Traverse through the list of processes in the snapshot
+    if (Process32First(processSnapshot, &processEntry)) {
+        do {
+            // For each process, capture a snapshot of all modules
```
alrodrig1: Here you are getting the modules for every single process, including ones that don't match the executable name.

You could optimize this by only getting the modules of the processes whose exe name matches MonAgentLauncher.exe.

For example:

```
wchar_t* executableName = PathFindFileName(executableFullPath); // Extract the file name from the full path

...

if (!_wcsicmp(processEntry.szExeFile, executableName)) { // If exe matches, then get the modules so we can inspect the full path.
```

kubernetes/windows/hostlogs/main.ps1
alrodrig1: I can gather from the code that we're starting the agent with placeholders if the configmap isn't set.  My question was about the _reasoning_ behind it. I assume there's a good reason why we don't bypass starting the agent instead. So, a comment in the code could help clarify this behavior.

On a related note, you'll find that most comments in code tend to describe 'what' the code is doing, but it's the 'why' (the reason a particular design choice was made) that tends to be more illuminating. Generally, the 'what' can be deduced by reading the code, while the 'why' is much harder to figure out without explicit mention.

kubernetes/windows/hostlogs/main.ps1
cacarlt: Using ```[System.Envrionment]::SetEnvironment...``` is the standard set by CI for setting process level environment variables.

https://github.com/microsoft/Docker-Provider/blob/386c6bf0e4eb9041c5f3751e7025fbaca670d50d/kubernetes/windows/main.ps1#L43

alrodrig1: That makes sense for them, considering they're dealing with both Process and Machine scopes. This doesn't seem to be the case for host logs, correct me if I'm wrong. That said, it's a minor point and more of me nitpicking. Feel free to disregard. 

The main point of my original comment was about getting a clear understanding of the purpose behind these placeholders.

kubernetes/host-logs-geneva.yaml
chris-raitano: A long initial delay makes sense when we have just a livenessprobe. If we start the livenessprobe too early it would think there's a problem and restart the pod

The purpose of the startup probe is a replacement for the initialdelay to determine when we're ready to start the livenessprobe. If it runs before we're ready that's ok – it will just run again. Ideally, we can detect that our pod started up correctly and switch over to the liveness probe as soon as possible. Basically, there's no benefit to a long initial delay, but a short initial delay means we can detect problems and/or configmap changes sooner

kubernetes/host-logs-geneva.yaml
chris-raitano: Only reason I disagree is for config map changes. If someone applied a config map just after starting the pod, no reason to make them wait an extra minute before those changes are detected

cacarlt: Took it out

kubernetes/host-logs-geneva.yaml
cacarlt: Took it out

kubernetes/windows/hostlogs/main.ps1
cacarlt: Moved env vars and renamed this check

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: I wasnt able to find a equvalient invalid_moudle_value error. If the snapshot doesnt exist the the Module32First will return false. https://learn.microsoft.com/en-us/windows/win32/api/tlhelp32/nf-tlhelp32-module32first

cacarlt: I do see ERROR_NO_MORE_FILES but it says we would need to call GetLastError which seems pointless no?

alrodrig1: I mean this line `HANDLE moduleSnapshot = CreateToolhelp32Snapshot(TH32CS_SNAPMODULE, processEntry.th32ProcessID);`, not `Module32First`.

This is done

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
```diff
@@ -105,50 +125,21 @@ def writeEnvScript(filepath)
...
-  
   if (File.file?(@configMapMountPath))
     ConfigParseErrorLogger.logError("config::unsupported/missing config schema version - '#{@configSchemaVersion}' , using defaults, please use supported schema version")
   end
-
 end
 
 def get_command_windows(env_variable_name, env_variable_value)
-  return "[System.Environment]::SetEnvironmentVariable(\"#{env_variable_name}\", \"#{env_variable_value}\", \"Process\")" + "\n"
+  return "[System.Environment]::SetEnvironmentVariable(\"#{env_variable_name}\", \"#{env_variable_value}\", \"Process\")" + "\n" + "[System.Environment]::SetEnvironmentVariable(\"#{env_variable_name}\", \"#{env_variable_value}\", \"Machine\")" + "\n"
```
chris-raitano: Machine level env got added back in

chris-raitano (approved):

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
```diff
@@ -92,25 +95,28 @@ def writeEnvScript(filepath)
...
-      puts "Using config map value: MONITORING_CONFIG_VERSION = #{@geneva_logs_config_version}"
-      puts "Using config map value: MONITORING_MANAGED_ID_IDENTIFIER = #{authIdParts[0]}"
-      puts "Using config map value: MONITORING_MANAGED_ID_VALUE= #{authIdParts[1]}"
+    authIdParts =  @geneva_gcs_authid.split('#', 2)
+    file.write(get_command_windows("MONITORING_MANAGED_ID_IDENTIFIER", authIdParts[0]))
+    file.write(get_command_windows("MONITORING_MANAGED_ID_VALUE", authIdParts[1]))
 
-      puts "config::info:successfully parsed geneva_logs_config settings"
-    end
+    puts "Using config map value: MONITORING_DATA_DIRECTORY = #{@geneva_account_environment}"
```
alrodrig1: Looks like you have the data directory and the environment variables inverted here:

puts "Using config map value: **MONITORING_DATA_DIRECTORY** = #{@_geneva_account_environment_}"

puts "Using config map value: _MONITORING_GCS_ENVIRONMENT_ = #{@**geneva_data_directory**}"

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,105 @@
...
 }
 
-/**
- <exe> <filesystemwatcherfilepath>
-**/
-int _tmain(int argc, wchar_t *argv[])
+/*
+* Function to check if a process is running based on its full path
+*/
+bool IsProcessRunningWithSpecifcPath(const wchar_t* executableFullPath)
```
alrodrig1: typo: Specifc -> Specific

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,105 @@
...
+/**
+ <exe> <monAgentLauncherExe> <filesystemwatcherfilepath>
+**/
+int wmain(int argc, wchar_t* argv[])
+{
+    try
+    {
+        if (argc < 2 || argc > 3)
+        {
+            wprintf_s(L"ERROR:unexpected number of arguments give. The expectation is at least 2 and no more then 3 agruments");
```
alrodrig1: Instead of indicating the incorrect number of arguments, let's specify which arguments are expected. Also, do a quick proofread for typos.

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,105 @@
...
+        } while (Process32Next(processSnapshot, &processEntry));
     }
 
-    return SUCCESS;
+    CloseHandle(processSnapshot);
+    return false;
 }
+
+/**
+ <exe> <monAgentLauncherExe> <filesystemwatcherfilepath>
```
alrodrig1: To avoid confusion down the line, let's add a note clarifying that this tool is used for both liveness and readiness.

Also, a breakdown of the expected parameters would be helpful.

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,105 @@
...
+int wmain(int argc, wchar_t* argv[])
+{
+    try
+    {
+        if (argc < 2 || argc > 3)
+        {
+            wprintf_s(L"ERROR:unexpected number of arguments give. The expectation is at least 2 and no more then 3 agruments");
+            return UNEXPECTED_ERROR;
+        }
+        else if (argc == 3)
```
alrodrig1: nit: `else` is not needed here since you are returning when true above. For clarity you can simplify to `if (argc == 3 && IsFileExists(argv[2]))`.

alrodrig1 (changes requested):

kubernetes/windows/hostlogs/main.ps1
cacarlt: Need to validate there is a comment here

kubernetes/windows/hostlogs/main.ps1
cacarlt: Moved setting the vars to the parser script

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Done

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
cacarlt: Done

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: There it is with the changes I lost where the typo was

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: In the if statement or the error message?

kubernetes/windows/hostlogs/main.ps1
cacarlt: This no longer exists since the ruby script takes care of the env vars always being present even if the config map has not been added..

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Fixed

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Updated error message

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Done

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: done

build/common/installer/scripts/tomlparser-hostlogs-geneva-config.rb
cacarlt: Done

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Fixed

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: fixed

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -7,37 +7,127 @@
...
+                    "Description: " 
+                    "- monitoringAgentLauncherPath: The mandatory path to the Monitoring Agent Launcher exe. The program will return exit code 1 if a process with this full path is not running. "
+                    "- watcherPath: The optional path to the watcher. If provided, the program will fail if this file exists.\n");
+            return UNEXPECTED_ERROR;
+        }
+        else if(argc == 2)
+        {
+            wprintf_s(L"INFO: Only the agentPath provided. The program won't check the watcher file.\n");
+        }
+        else if (argc == 3 && IsFileExists(argv[2]))
```
chris-raitano, line 107: Check the logic here.  If 3 args and configmap changes -> fail, else -> also fail. 
So the liveness probe always fails and only the startup probe checks if the MA is running

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -75,29 +80,46 @@ bool IsProcessRunningWithSpecifcPath(const wchar_t* executableFullPath)
...
 **/
 int wmain(int argc, wchar_t* argv[])
 {
     try
     {
-        if (argc < 2 || argc > 3)
+        if (argc < 2)
         {
-            wprintf_s(L"ERROR:unexpected number of arguments give. The expectation is at least 2 and no more then 3 agruments");
+            wprintf_s(L"ERROR: No arguments provided. "
```
alrodrig1: Consider moving the usage info to a variable to avoid code duplication:

```
const wchar_t* USAGE_INFO = L"Usage: livenessprobe.exe <monitoringAgentLauncherPath> [fileSystemWatcherTextFilePath]\n"
                             "Description: \n"
                             "- monitoringAgentLauncherPath: The mandatory path to the Monitoring Agent Launcher exe. The program will return exit code 1 if a process with this full path is not running.\n"
                             "- watcherPath: The optional path to the watcher. If provided, the program will fail if this file exists.\n";

if (argc < 2)
{
	wprintf_s(L"ERROR: No arguments provided.\n");
	wprintf_s(USAGE_INFO);
	return UNEXPECTED_ERROR;
}

if (argc > 3)
{
	wprintf_s(L"ERROR: Too many arguments provided.\n");
	wprintf_s(USAGE_INFO);
	return UNEXPECTED_ERROR;
}
```

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: Fixed

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
```diff
@@ -75,29 +80,46 @@ bool IsProcessRunningWithSpecifcPath(const wchar_t* executableFullPath)
...
+                    "Usage: livesnessprobe.exe <monitoringAgentLauncherPath> [fileSystemWatcherTextFilePath] "
+                    "Description: " 
+                    "- monitoringAgentLauncherPath: The mandatory path to the Monitoring Agent Launcher exe. The program will return exit code 1 if a process with this full path is not running. "
+                    "- watcherPath: The optional path to the watcher. If provided, the program will fail if this file exists.\n");
             return UNEXPECTED_ERROR;
         }
-        else if (argc == 3)
+        else if(argc == 2)
```
alrodrig1, line 103: Usually, when the `if` section of a conditional has a `return` statement, you don't need to include the code that follows in an `else` block. For example:

```
if (A)
{
    return "A is true";
}
else
{
    return "A is false";
}
```

is equivalent to:

```
if (A)
{
    return "A is true";
}

return "A is false";
```

This reduces nesting and makes the logic easier to follow, especially when the conditionals are more complicated and the code blocks longer.

build/hostlogswindows/installer/livenessprobe/livenessprobe.cpp
cacarlt: done

alrodrig1 (approved):

cacarlt:
Completed

PR #1027, opened 2023-07-18
set mdsd limits

pfrcks:
- Adds new configmap variable backpressure_memory_threshold under agent_config.mdsd_config to set MDSD backpressure limit
- Sets MDSD backpressure limit to 50% of container memory limit if no custom fluent-bit tail memory buffer limit or no custom backpressure limit is provided from configmap


pfrcks:
For container memory limit 750MB running under high load (high scale log generation), total memory consumption hovers around 700MB when we set fbit tail limits as

image


and mdsd backpressure limit to 400MB via configmap.

(field to note in the images below is RSS).

image

If we dont apply any configmap then fbit has low limits causing its buffer to fill up very easily and therefore there is little load on mdsd. For that scenario 50% of limit is more than enough. Eg here is the memory usage without any configmap values. As we can see mdsd memory usage is very low as compared to the above.

image


The default case will help us tackle cases where network outage cause mdsd buffer to grow beyond container limit due to high backpressure limit

kubernetes/ama-logs.yaml
```diff
@@ -441,6 +441,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: CONTAINER_MEMORY_LIMIT
+              valueFrom:
```
ganga1980, line 445: we should add these changes for helm chart as well.

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -263,17 +265,24 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+        # ama-logs daemonset only settings
+        if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp(@daemonset) == 0 && @containerType.nil?
           mdsdMonitoringMaxEventRate = mdsd_config[:monitoring_max_event_rate]
           if is_valid_number?(mdsdMonitoringMaxEventRate)
             @mdsdMonitoringMaxEventRate = mdsdMonitoringMaxEventRate.to_i
             puts "Using config map value: monitoring_max_event_rate  = #{@mdsdMonitoringMaxEventRate}"
           end
         end
+
+        mdsdBackPressureThreshold = mdsd_config[:backpressure_memory_threshold]
```
ganga1980: 1. include the units in the config field name. For example, backpressure_memory_threshold_in_mb 
2.  Add checks to prevent customer preventing higher than limits or very small value. If in such cases, just add the message saying backpressure threshold ignored and used the default.
3.  add telemetry to track back pressure threshold value.

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -404,6 +413,10 @@ def populateSettingValuesFromConfigMap(parsedConfig)
     file.write("export MONITORING_MAX_EVENT_RATE=#{@mdsdMonitoringMaxEventRate}\n")
   end
 
+  if @mdsdBackPressureThreshold > 0
+    file.write("export BACKPRESSURE_THRESHOLD=#{@mdsdBackPressureThreshold}\n")
```
ganga1980: include units in the name of this variable BACKPRESSURE_THRESHOLD as well.

kubernetes/linux/main.sh
```diff
@@ -855,6 +855,23 @@ else
 fi
 source ~/.bashrc
 
+# manually set backpressure value using container limit only when neither backpressure or fbit tail buffer is provided through configmap
+if [ -n "${BACKPRESSURE_THRESHOLD}" ]; then
+      echo "Setting MDSD backpressure threshold from configmap"
+      export MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB=${BACKPRESSURE_THRESHOLD}
```
ganga1980: potential mismatch in the units (left and right), left is MB and we dont know whats the units on the right hand side. ensure we have consistent units on both sides to avoid any manual errors.

kubernetes/linux/main.sh
```diff
@@ -855,6 +855,23 @@ else
...
+      echo "export MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB=$MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB" >> ~/.bashrc
+      source ~/.bashrc
+elif [ -z "${FBIT_TAIL_MEM_BUF_LIMIT}" ]; then
+      if [ -n "${CONTAINER_MEMORY_LIMIT}" ]; then
+            echo "Setting MDSD backpressure threshold from container limit"
+            limit_in_mebibytes=$((CONTAINER_MEMORY_LIMIT / 1048576))
```
ganga1980: is this always bytes regardless of the units on the resource limit (for example MB, Mi, GB, Gi )?

kubernetes/linux/main.sh
```diff
@@ -855,6 +855,23 @@ else
...
+
+            export MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB=$((limit_in_mebibytes * 50 / 100))
```
ganga1980, line 869: does the MDSD has relation between backpressure threshold and max events rate? if we have low threshold and high max events rate and vice versa, does it cause any issues?  please double check.

kubernetes/linux/main.sh
```diff
@@ -855,6 +855,23 @@ else
...
+            echo "export MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB=$MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB" >> ~/.bashrc
```
ganga1980, line 870: add log message with MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB value . this will be helpful for any incident and also with VPA integration.

kubernetes/ama-logs.yaml
```diff
@@ -526,6 +531,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.hostIP
+            - name: CONTAINER_MEMORY_LIMIT
```
ganga1980: Include the units in the environment variable name. and also, check the units are constant regardless of the units specified on the resource limit.

kubernetes/linux/main.sh
pfrcks: yes its always in bytes

build/common/installer/scripts/tomlparser-agent-config.rb
pfrcks: 1. done
2. backpressure now bounded between 100MB and container limit
3. done

build/common/installer/scripts/tomlparser-agent-config.rb
pfrcks: done

kubernetes/linux/main.sh
pfrcks: addressed

kubernetes/linux/main.sh
pfrcks: done

kubernetes/ama-logs.yaml
pfrcks: done

charts/azuremonitor-containers/templates/ama-logs-daemonset.yaml
```diff
@@ -83,6 +83,11 @@ spec:
          valueFrom:
             fieldRef:
               fieldPath: status.hostIP
+       - name: CONTAINER_MEMORY_LIMIT_IN_BYTES
```
ganga1980, line 86: lets get this manifest update as part of this week's AKS RP release, so that we can get the agent changes through toggle first to see if this change working as expected and without any side effects.

pfrcks: acknowledged

kubernetes/linux/main.sh
pfrcks: confirmed with Sid that these two are seperate variables and work independently

ganga1980 (dismissed):

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -263,17 +266,26 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+        if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp(@daemonset) == 0 && @containerType.nil?
           mdsdMonitoringMaxEventRate = mdsd_config[:monitoring_max_event_rate]
           if is_valid_number?(mdsdMonitoringMaxEventRate)
             @mdsdMonitoringMaxEventRate = mdsdMonitoringMaxEventRate.to_i
             puts "Using config map value: monitoring_max_event_rate  = #{@mdsdMonitoringMaxEventRate}"
           end
         end
+
+        mdsdBackPressureThresholdInMB = mdsd_config[:backpressure_memory_threshold_in_mb]
+        if is_valid_number?(mdsdBackPressureThresholdInMB) && is_valid_number?(@containerMemoryLimitInBytes) && mdsdBackPressureThresholdInMB.to_i < (@containerMemoryLimitInBytes.to_i / 1048576) && mdsdBackPressureThresholdInMB.to_i > 100
```
jatakiajanvi12, line 303: Should this also be a function : is_valid_backpressure_limit?  or something
Since it is used in multiple files

kubernetes/linux/main.sh
```diff
@@ -855,6 +855,28 @@ else
...
+# manually set backpressure value using container limit only when neither backpressure or fbit tail buffer is provided through configmap
+if [ -n "${BACKPRESSURE_THRESHOLD_IN_MB}" ]; then
+      export MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB=${BACKPRESSURE_THRESHOLD_IN_MB}
+      echo "export MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB=$MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB" >> ~/.bashrc
+      echo "Setting MDSD backpressure threshold from configmap: ${MDSD_BACKPRESSURE_MONITOR_MEMORY_THRESHOLD_IN_MB} MB"
+      source ~/.bashrc
+elif [ -z "${FBIT_TAIL_MEM_BUF_LIMIT}" ]; then
+      if [ -n "${CONTAINER_MEMORY_LIMIT_IN_BYTES}" ]; then
+            echo "Container limit in bytes: ${CONTAINER_MEMORY_LIMIT_IN_BYTES}"
+            limit_in_mebibytes=$((CONTAINER_MEMORY_LIMIT_IN_BYTES / 1048576))
```
jatakiajanvi12, line 867: limit_in_mebibytes, is this a typo?

build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -263,17 +266,26 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+        # ama-logs daemonset only settings
+        if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp(@daemonset) == 0 && @containerType.nil?
           mdsdMonitoringMaxEventRate = mdsd_config[:monitoring_max_event_rate]
           if is_valid_number?(mdsdMonitoringMaxEventRate)
             @mdsdMonitoringMaxEventRate = mdsdMonitoringMaxEventRate.to_i
             puts "Using config map value: monitoring_max_event_rate  = #{@mdsdMonitoringMaxEventRate}"
           end
         end
+
+        mdsdBackPressureThresholdInMB = mdsd_config[:backpressure_memory_threshold_in_mb]
```
jatakiajanvi12, line 302: Should we add this change in the configmap as well?

ganga1980 (approved):

PR #1028, opened 2023-07-20
Gangams/disable threaded by default and make optional

ganga1980:
Disable fluent-bit threading by default and make it optional since fluent-bit has bug (https://github.com/fluent/fluent-bit/issues/7071) with threaded. Once the fluent-bit fixes threading bug, then this can be configurable.

### DIFF
--- a/build/common/installer/scripts/tomlparser-geneva-config.rb
+++ b/build/common/installer/scripts/tomlparser-geneva-config.rb
@@ -23,6 +23,7 @@ GENEVA_SUPPORTED_ENVIRONMENTS = ["Test", "Stage", "DiagnosticsProd", "Firstparty
 @tenant_namespaces = ""
 @geneva_gcs_authid = ""
 @azure_json_path = "/etc/kubernetes/host/azure.json"
+@enable_fbit_threading = false
 
 # Use parser to parse the configmap toml file to a ruby structure
 def parseConfigMap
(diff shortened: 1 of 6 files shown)


build/common/installer/scripts/tomlparser-geneva-config.rb
```diff
@@ -76,6 +77,11 @@ def populateSettingValuesFromConfigMap(parsedConfig)
                 end
               end
             end
+            enable_fbit_threading = parsedConfig[:integrations][:geneva_logs][:enable_threading].to_s
```
jatakiajanvi12, line 80: Should we add this change in configmap as well?

pfrcks:
why are we exposing this via configmap? once the bug is fixed shouldn't we just change to code for everyone to threaded on?

build/common/installer/scripts/tomlparser-geneva-config.rb
ganga1980: > Should we add this change in configmap as well?

> since this internal only feature hence not added to configmap.

ganga1980:
> why are we exposing this via configmap? once the bug is fixed shouldn't we just change to code for everyone to threaded on?

> this bug is active since quite sometime and when they fixed, we can ask internal customers to configure the configmap and validate . Potentially, we can make default if its reliable after the validation.

pfrcks:
> > why are we exposing this via configmap? once the bug is fixed shouldn't we just change to code for everyone to threaded on?
> 
> > this bug is active since quite sometime and when they fixed, we can ask internal customers to configure the configmap and validate . Potentially, we can make default if its reliable after the validation.

ok. pls add the bug link here to correlate in the future.

pfrcks (approved):
pls check that this works as expected (setting and removing configmap values). if already done then lgtm.

also pls add the fbit bug link to the pr description
 **added bug link and validated this change.**

PR #1046, opened 2023-08-10
retry only retriable http status codes

ganga1980:
Existing implementation of Windows agent retries non 200 status codes. This can impact the ingestion endpoint for client side errors.  This PR fixes to only retry for the retriable http status codes.

### DIFF
--- a/source/plugins/go/src/oms.go
+++ b/source/plugins/go/src/oms.go
@@ -1084,7 +1084,7 @@ func PostTelegrafMetricsToLA(telegrafRecords []map[interface{}]interface{}) int 
 				UpdateNumTelegrafMetricsSentTelemetry(0, 1, 1, 0)
 			}
 			return output.FLB_RETRY
-		} else if resp.StatusCode >= 200 && resp.StatusCode < 300 {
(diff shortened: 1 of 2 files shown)


source/plugins/go/src/oms.go
```diff
@@ -1072,21 +1072,25 @@ func PostTelegrafMetricsToLA(telegrafRecords []map[interface{}]interface{}) int
...
 		}
 
-		if resp == nil || resp.StatusCode != 200 {
+		if resp != nil && resp.Body != nil {
+			defer resp.Body.Close()
+		}
+
+		if resp == nil || IsRetriableError(resp.StatusCode) {
 			if resp != nil {
 				Log("PostTelegrafMetricsToLA::Error:(retriable) RequestID %s Response Status %v Status Code %v", reqID, resp.Status, resp.StatusCode)
```
pfrcks, line 1081: should this Log be updated to denote this is a retriable error?

ganga1980: > should this Log be updated to denote this is a retriable error?

> Log message already has **(retriable)**. are you referring something else?

pfrcks (approved):

PR #1048, opened 2023-08-11
Updating telegraf to version 1.27.2

jatakiajanvi12:
MemoryRSS
image

CPU
image


kubernetes/linux/setup.sh
```diff
@@ -67,7 +67,7 @@ sudo tdnf install jq-1.6-1.cm2 -y
 #used to setcaps for ruby process to read /proc/env
 sudo tdnf install libcap -y
 
-sudo tdnf install telegraf-1.26.0 -y
+sudo tdnf install telegraf-1.27.2 -y
```
ganga1980, line 70: @jatakiajanvi12 , 1. did you verify functionality (disk, diskio, network and prom metrics) works and there are no regressions? 2. Does this upgrade addresses all the critical and high telegraf vulns which we were added trivyignore? if yes, please remove the vulns which addressed from trivyignore so that we know the remaining ones to address.

.trivyignore
```diff
@@ -1,17 +1,4 @@
...
-
-# telegraf HIGH
-CVE-2023-27561
-CVE-2023-28840
-CVE-2023-28642
-CVE-2023-2253
-CVE-2023-34231
-
+#telegraf MEDIUM
 GHSA-2w8w-qhg4-f78j
```
ganga1980, line 2: is this not addressed with telegraf upgrade?

ganga1980:
> MemoryRSS image
> 
> CPU image

are these perf charts for ds, rs and prom side car? are these with load or no-load?

.trivyignore
jatakiajanvi12: No this will be in the next version 1.27.3

jatakiajanvi12:
> > MemoryRSS image
> > CPU image
> 
> are these perf charts for ds, rs and prom side car? are these with load or no-load?

yes these charts are for all the containers. The cluster has one extra logs container, nothing else

kubernetes/linux/setup.sh
jatakiajanvi12: 1. Yes all of them are present. 
2. Updated the PR

ganga1980 (approved):

PR #1049, opened 2023-08-11
Move from omuxsock to omfwd for syslog

pfrcks:
- uses omfwd for syslog data on port 28330 (same as in VMs)
- updates to MDSD to 1.27.4 which has omfwd support


wanlonghenry:
could you attach some testing screenshots since we are upgrading mdsd

pfrcks:
> could you attach some testing screenshots since we are upgrading mdsd

compared with 3.1.11. Do not see any perf impact

PR #1056, opened 2023-08-23
Telemetry optimization - Remove unnecessary customDimensions field and shortening some fields

jatakiajanvi12:
image

### DIFF
--- a/source/plugins/go/src/telemetry.go
+++ b/source/plugins/go/src/telemetry.go
@@ -68,6 +68,11 @@ var (
 	PromMonitorPodsLabelSelectorLength int
 	//Tracks the number of monitor kubernetes pods field selectors and sent only from prometheus sidecar (uses ContainerLogTelemetryTicker)
 	PromMonitorPodsFieldSelectorLength int
+	//Map to get the short name for the controller type
+	ControllerType = map[string]string{
+		"daemonset":  "DS",
+		"replicaset": "RS",
+	}
 )
 
 const (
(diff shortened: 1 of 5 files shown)


source/plugins/go/src/telemetry.go
```diff
@@ -68,6 +68,11 @@ var (
 	PromMonitorPodsLabelSelectorLength int
 	//Tracks the number of monitor kubernetes pods field selectors and sent only from prometheus sidecar (uses ContainerLogTelemetryTicker)
 	PromMonitorPodsFieldSelectorLength int
```
ganga1980, line 70: as part of this change, can we make following changes 

1. from DockerVersion to "cri" 
2.  workspaceID to WSID
3. workspaceCloud to "WSCloud"
4. PodName to Pod

source/plugins/go/src/telemetry.go
```diff
@@ -360,40 +365,27 @@ func InitializeTelemetryClient(agentVersion string) (int, error) {
 	CommonProperties = make(map[string]string)
 	CommonProperties["Computer"] = Computer
 	CommonProperties["WorkspaceID"] = WorkspaceID
-	CommonProperties["ControllerType"] = os.Getenv("CONTROLLER_TYPE")
-	CommonProperties["AgentVersion"] = agentVersion
+	CommonProperties["ControllerType"] = ControllerType[strings.ToLower(os.Getenv("CONTROLLER_TYPE"))]
```
ganga1980: instead of ControllerType, how about going with just "Controller" ?

source/plugins/go/src/telemetry.go
```diff
@@ -360,40 +365,27 @@ func InitializeTelemetryClient(agentVersion string) (int, error) {
...
 	// if the aks resource id is not defined, it is most likely an ACS Cluster
 	if aksResourceID == "" {
-		CommonProperties["ACSResourceName"] = os.Getenv(envACSResourceName)
-		CommonProperties["ClusterType"] = clusterTypeACS
+		CommonProperties["ID"] = os.Getenv(envACSResourceName)
 
-		CommonProperties["SubscriptionID"] = ""
-		CommonProperties["ResourceGroupName"] = ""
-		CommonProperties["ClusterName"] = ""
 		CommonProperties["Region"] = ""
```
ganga1980: We can remove this since this is empty string which will not be useful, right?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -47,59 +49,43 @@ def initializeUtility()
...
           @@CustomProperties["Region"] = ENV[@@EnvAksRegion]
         end
 
         #Commenting it for now from initilize method, we need to pivot all telemetry off of kubenode docker version
         #getDockerInfo()
         @@CustomProperties["WorkspaceID"] = getWorkspaceId
-        @@CustomProperties["AgentVersion"] = ENV[@@EnvAgentVersion]
-        @@CustomProperties["ControllerType"] = ENV[@@EnvControllerType]
+        @@CustomProperties["Version"] = ENV[@@EnvAgentVersion]
+        @@CustomProperties["ControllerType"] = @@controllerType[ENV[@@EnvControllerType].downcase]
```
ganga1980: Please double check these naming changes are updated every where.

source/plugins/go/src/telemetry.go
ganga1980: validate these changes. and also, we would need to update our dashboard once after the change rollout. please add a task for this.

source/plugins/go/src/telemetry.go
jatakiajanvi12: Not changing the DockerVersion field. We use the same field name in the actual KubeNodeInventory table. I think better to have this consistency. Even for PodName, it is used in ContainerInventory. What do you think @ganga1980 ?

source/plugins/ruby/ApplicationInsightsUtility.rb
jatakiajanvi12: There are mentions in the telegraf.conf files in comments...Do we want to make changes there as well?

source/plugins/go/src/telemetry.go
ganga1980: > Not changing the DockerVersion field. We use the same field name in the actual KubeNodeInventory table. I think better to have this consistency. Even for PodName, it is used in ContainerInventory. What do you think @ganga1980 ?

> @jatakiajanvi12 , since this is telemetry and we dont need to keep consistency. as long as we undestand these fields  we are good.  We kept the DockerVersion in KubeNodeInventory  table to avoid breaking changes for customers.

source/plugins/ruby/ApplicationInsightsUtility.rb
ganga1980: > There are mentions in the telegraf.conf files in comments...Do we want to make changes there as well?

> we only need to change specific to telemetry. which fields are you referring in telegraf.conf? I see the ControllerType but its not commented out and also not related to telemetry.

source/plugins/go/src/telemetry.go
jatakiajanvi12: Updated the fields

source/plugins/ruby/ApplicationInsightsUtility.rb
jatakiajanvi12: Talking about these: https://github.com/microsoft/Docker-Provider/blob/750b6b3a941c4b16e30c89c84c4372fa5581f5dd/build/linux/installer/conf/telegraf.conf#L19C1-L27C39

ganga1980: > Talking about these: https://github.com/microsoft/Docker-Provider/blob/750b6b3a941c4b16e30c89c84c4372fa5581f5dd/build/linux/installer/conf/telegraf.conf#L19C1-L27C39

all these fields are commented out so we dont need to touch this file.

source/plugins/go/src/telemetry.go
jatakiajanvi12: image
image
image

jatakiajanvi12: Tested on both windows and all tables

ganga1980 (approved):

wanlonghenry:
Do we know whether we have any dashboard or alerts that need to be adjusted because of this change?

jatakiajanvi12:
> Do we know whether we have any dashboard or alerts that need to be adjusted because of this change?

Yes have added a task for this

PR #1059, opened 2023-09-05
Update the pipeline to take WINDOWS_AMA_URL from the pipeline

jatakiajanvi12:
- Added the variable in the pipeline
- The default url is the one we are using currently

### DIFF
--- a/.pipelines/azure_pipeline_mergedbranches.yaml
+++ b/.pipelines/azure_pipeline_mergedbranches.yaml
@@ -38,6 +38,9 @@ jobs:
       linuxImagetag="$commit"-"$datetime"
       windowsImageTag=win-"$commit"-"$datetime"
 
+      echo "Just checking prints"
+      echo "The url is $WINDOWS_AMA_URL"
+
       telemetryTag=$linuxImagetag
       if [ -z "$TELEMETRY_TAG" ]
         then
--- a/.pipelines/azure_pipeline_mergedbranches.yaml
+++ b/.pipelines/azure_pipeline_mergedbranches.yaml
@@ -236,7 +236,7 @@ jobs:
         az account set -s ${{ variables.subscription }}
         az acr login -n ${{ variables.containerRegistry }}
 
-        docker build --isolation=hyperv --tag ${{ variables.repoImageName }}:$(windowsImageTag)-$(windows2019BaseImageVersion) --build-arg WINDOWS_VERSION=$(windows2019BaseImageVersion) --build-arg IMAGE_TAG=$(windowsTelemetryTag) .
+        docker build --isolation=hyperv --tag ${{ variables.repoImageName }}:$(windowsImageTag)-$(windows2019BaseImageVersion) --build-arg WINDOWS_VERSION=$(windows2019BaseImageVersion) --build-arg IMAGE_TAG=$(windowsTelemetryTag) --build-arg WINDOWS_AMA_URL=$(WINDOWS_AMA_URL) .
         if ("$(Build.Reason)" -ne "PullRequest") {
            docker push ${{ variables.repoImageName }}:$(windowsImageTag)-$(windows2019BaseImageVersion)
         }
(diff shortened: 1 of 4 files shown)


.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -236,7 +236,7 @@ jobs:
         az account set -s ${{ variables.subscription }}
         az acr login -n ${{ variables.containerRegistry }}
 
-        docker build --isolation=hyperv --tag ${{ variables.repoImageName }}:$(windowsImageTag)-$(windows2019BaseImageVersion) --build-arg WINDOWS_VERSION=$(windows2019BaseImageVersion) --build-arg IMAGE_TAG=$(windowsTelemetryTag) .
+        docker build --isolation=hyperv --tag ${{ variables.repoImageName }}:$(windowsImageTag)-$(windows2019BaseImageVersion) --build-arg WINDOWS_VERSION=$(windows2019BaseImageVersion) --build-arg IMAGE_TAG=$(windowsTelemetryTag) --build-arg WINDOWS_AMA_URL=$(WINDOWS_AMA_URL) .
```
ganga1980, line 239: does build fail when WINDOWS_AMA_URL is not specified? how can we ensure the build url specified and its valid url?

kubernetes/windows/setup.ps1
```diff
@@ -67,9 +67,9 @@ Write-Host ('Extracting Certificate Generator Package')
     Expand-Archive -Path /opt/amalogswindows/certificategenerator.zip -Destination /opt/amalogswindows/certgenerator/ -Force
 Write-Host ('Finished Extracting Certificate Generator Package')
 
-Write-Host ('Installing Windows Azure Monitor Agent');
+$windowsazuremonitoragent = [System.Environment]::GetEnvironmentVariable('WINDOWS_AMA_URL')
```
ganga1980, line 70: I think, we should validate the url here and fail the build if this is not valid url.

jatakiajanvi12: The build will fail if the url is not valid and the download doesn't work. What other kind of validation you want to add?

.pipelines/azure_pipeline_mergedbranches.yaml
jatakiajanvi12: When not specified, the default is used

kubernetes/windows/setup.ps1
ganga1980: > The build will fail if the url is not valid and the download doesn't work. What other kind of validation you want to add?

> if build able to catch then we are good.

kubernetes/windows/setup.ps1
```diff
@@ -67,9 +67,12 @@ Write-Host ('Extracting Certificate Generator Package')
...
+if ([string]::IsNullOrEmpty($windowsazuremonitoragent)) {
+    $windowsazuremonitoragent = "https://github.com/microsoft/Docker-Provider/releases/download/windows-ama-bits/genevamonitoringagent.46.3.2.zip"
```
ganga1980, line 73: We should add log message about this default so that we can troubleshoot issues if any.

kubernetes/windows/setup.ps1
```diff
@@ -67,9 +67,12 @@ Write-Host ('Extracting Certificate Generator Package')
...
+}
+Write-Host ('Installing Windows Azure Monitor Agent: ' + $windowsazuremonitoragent);
```
ganga1980: nit: remove redundant ; here

.pipelines/azure_pipeline_mergedbranches.yaml
ganga1980, line 239: is WINDOWS_AMA_BUILD_URL or WINDOWS_AMA_BUILD_PATH better naming for this?

kubernetes/windows/setup.ps1
wanlonghenry: If we specify the windowsazuremonitoragent  in the pipeline variable, then we don't need to hardcode here right? And we should also log it in build pipeline

.pipelines/azure_pipeline_mergedbranches.yaml
jatakiajanvi12: I think this is okay...There is no other meaning for WINDOWS_AMA_URL anyway

kubernetes/windows/setup.ps1
jatakiajanvi12: Not defaulting in the pipeline so that we have a log history of when we made the changes for default version. We do print the url link log in the next line. But have added another log

ganga1980 (approved):

PR #1062, opened 2023-09-11
Windows Host Logs Scale Infrastructure

cacarlt:


test/whl-scale-tests/common.ps1
```diff
@@ -0,0 +1,66 @@
...
+
+    # Ensure the input file exists
+    if (-not (Test-Path -Path $InputFilePath)) {
+        Write-Host "  Input File: '$InputFilePath' does not exist" -ForegroundColor Red
+        exit
+    }
+
+    foreach($subItem in $Substitutions.GetEnumerator())
+    {
+        (Get-Content $InputFilePath).Replace($subItem.Name, $subItem.Value) | Set-Content $OutputFilePath
```
chris-raitano: This only works when input file and output file are the same since each replace is being applied to the original input file. 

Maybe something like this instead, which also reduces number of file reads/writes
```
$content = Get-Content $InputFilePath;
foreach($subItem in $Substitutions.GetEnumerator())
{
    $content = $content.Replace($subItem.Name, $subItem.Value); 
}
$content | Set-Content $OutputFilePath;
```

camcast3: Good catch

camcast3: Fixed

cacarlt: Fixed

PR #1072, opened 2023-09-22
Update java agent version and add namespace attribute

ramanujbhattacharjee93:


appmonitoring/ts/src/AddedTypes.ts
```diff
@@ -137,6 +137,7 @@ export class AddedTypes {
 cloud.region=${armRegion},\
 k8s.cluster.name=${clusterName},\
 k8s.pod.namespace=$(POD_NAMESPACE),\
```
tokaplan: I've asked on a thread, if nobody objects against it - let's just remove _k8s.pod.namespace_.

tokaplan: Ok we have a green light, let's proceed with the removal.

tokaplan (changes requested):

tokaplan (approved):
Approved

PR #1075, opened 2023-09-26
Gangams/implement timebound token and token refresh

ganga1980:
Implementation of timebound service account and token refresh, and token expiry is 1hour and refreshed from file before 10 mins of the expiry. Verified on both Linux and windows. From k8s min version 1.20.13 and latest version 1.28.0. We dont have any AKS test clusters to valid older than 1.20.  and also validated HELM chart.

As per [KEP] , this feature introduced in 1.13 as beta - https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/1205-bound-service-account-tokens#graduation-criteria

To avoid any breaking change on unsupported versions, we will add k8s version check in AKS yaml for manifest. Agent code agnostic to the yaml changes and it will support both legacy and JWT tokens.

### DIFF
--- a/charts/azuremonitor-containers/templates/ama-logs-deployment.yaml
+++ b/charts/azuremonitor-containers/templates/ama-logs-deployment.yaml
@@ -128,6 +128,9 @@ spec:
        - containerPort: 25224
          protocol: UDP
        volumeMounts:
+        - name: token
+          mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+          readOnly: true
         - mountPath: /var/run/host
           name: docker-sock
         - mountPath: /var/log
@@ -182,6 +185,12 @@ spec:
    {{- end }}
    {{- end }}
    volumes:
+    - name: token
+      projected:
+        sources:
+        - serviceAccountToken:
+            path: token
+            expirationSeconds: 3600
     - name: docker-sock
       hostPath:
        path: /var/run
--- a/charts/azuremonitor-containers/templates/ama-logs-deployment.yaml
+++ b/charts/azuremonitor-containers/templates/ama-logs-deployment.yaml
@@ -128,7 +128,7 @@ spec:
        - containerPort: 25224
          protocol: UDP
        volumeMounts:
-        - name: token
+        - name: kube-api-access
           mountPath: /var/run/secrets/kubernetes.io/serviceaccount
           readOnly: true
         - mountPath: /var/run/host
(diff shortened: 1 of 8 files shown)


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -114,16 +116,27 @@ def getKubeResourceInfoV2(resource, api_group: nil)
...
-          return @@TokenStr
-        else
-          @Log.warn("Unable to read token string from #{@@TokenFileName}: #{error}")
-          return nil
+        if (@@TokenStr.nil? || (@@TokenExpiry - DateTime.now.to_time.to_i).abs <= Constants::SERVICE_ACCOUNT_TOKEN_REFRESH_INTERVAL_SECONDS) # refresh token from token file if its near expiry
+          if File.exist?(@@TokenFileName) && File.readable?(@@TokenFileName)
+            @@TokenStr = File.read(@@TokenFileName).strip
+            begin
+              token_info = JWT.decode(@@TokenStr, nil, false)
+              @@TokenExpiry = token_info[0]["exp"]
```
pfrcks, line 125: is it possible that token_info[0] does not have "exp" key?

charts/azuremonitor-containers/templates/ama-logs-deployment.yaml
```diff
@@ -182,6 +185,23 @@ spec:
...
+            path: token
+            expirationSeconds: 3600
+        - configMap:
+            items:
+            - key: ca.crt
+              path: ca.crt
+            name: kube-root-ca.crt
+        - downwardAPI:
+            items:
+            - fieldRef:
```
jatakiajanvi12, line 201: For my understanding, why do we need the namespace ?

source/plugins/ruby/KubernetesApiClient.rb
ganga1980: > is it possible that token_info[0] does not have "exp" key?

> The **exp** field is mandatory field in JWT token.  Refer to "Standard fields" in https://en.wikipedia.org/wiki/JSON_Web_Token.  If the token is JWT then "exp" field must be there.

charts/azuremonitor-containers/templates/ama-logs-deployment.yaml
ganga1980: > For my understanding, why do we need the namespace ?
> We dont use namespace. I kept the same behavior as today i.e.  under /var/run/secrets/kubernetes.io/serviceaccount, there will be token, crt files and namespace. The namespace here indicates what the namespace of the current running pod.

pfrcks:
although it should not be different but is it worth testing in ARC as well?

ganga1980:
> although it should not be different but is it worth testing in ARC as well?

validated the chart as well.

pfrcks (approved):

PR #1093, opened 2023-10-05
use omfwd for syslog

pfrcks:
- updates mdsd 1.27.4 from 1.26.1
- uses omfwd instead of omuxsock for syslog data transfer between host node and container
- uses default hostPort 28330 which can be overridden by providing SYSLOG_HOST_PORT (comes from toggle)

### DIFF
--- a/build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
+++ b/build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
@@ -5,6 +5,6 @@ template(name="MDSD_RSYSLOG_TraditionalForwardFormat" type="string" string="<%PR
 template="MDSD_RSYSLOG_TraditionalForwardFormat"
 queue.type="LinkedList"
 action.resumeRetryCount="-1"
-queue.size="50000"
+queue.size="10000"
 queue.saveonshutdown="on"
 target="127.0.0.1" Port="28330" Protocol="tcp")
--- a/build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
+++ b/build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
@@ -1,5 +1,10 @@
-$ModLoad omuxsock 
-$OMUxSockSocket /var/run/mdsd-ci/default_syslog.socket 
 template(name="MDSD_RSYSLOG_TraditionalForwardFormat" type="string" string="<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag%%msg:::sp-if-no-1st-sp%%msg%")
-$OMUxSockDefaultTemplate MDSD_RSYSLOG_TraditionalForwardFormat
-*.* :omuxsock:
+
+# Forwarding all events through TCP port
+*.* action(type="omfwd"
+template="MDSD_RSYSLOG_TraditionalForwardFormat"
+queue.type="LinkedList"
+action.resumeRetryCount="-1"
+queue.size="50000"
+queue.saveonshutdown="on"
+target="127.0.0.1" Port="28330" Protocol="tcp")
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -842,6 +842,8 @@ else
             echo "export MONITORING_USE_GENEVA_CONFIG_SERVICE=$MONITORING_USE_GENEVA_CONFIG_SERVICE" >> ~/.bashrc
             export MDSD_USE_LOCAL_PERSISTENCY="false"
             echo "export MDSD_USE_LOCAL_PERSISTENCY=$MDSD_USE_LOCAL_PERSISTENCY" >> ~/.bashrc
+            export MDSD_DEFAULT_TCP_SYSLOG_PORT=28330
+            echo "export MDSD_DEFAULT_TCP_SYSLOG_PORT=$MDSD_DEFAULT_TCP_SYSLOG_PORT" >> ~/.bashrc
       else
             echo "*** setting up oneagent in legacy auth mode ***"
             CIWORKSPACE_id="$(cat /etc/ama-logs-secret/WSID)"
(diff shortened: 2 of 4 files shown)


kubernetes/linux/setup.sh
```diff
@@ -40,9 +40,9 @@ gem uninstall time --version 0.2.0
 gem uninstall uri --version 0.11.0
 
 if [ "${ARCH}" != "arm64" ]; then
-    wget "https://github.com/microsoft/Docker-Provider/releases/download/official%2Fmdsd%2F1.26.1/azure-mdsd-1.26.1-build.master.97.x86_64.rpm" -O azure-mdsd.rpm
+    wget "https://github.com/microsoft/Docker-Provider/releases/download/official%2Fmdsd%2F1.27.4/azure-mdsd-1.27.4-build.master.140.x86_64.rpm" -O azure-mdsd.rpm
```
jatakiajanvi12: Can we use tdnf install instead?

wanlonghenry, line 43: this is the same change with PR: update mdsd to 1.27.4  right?  do you want to combine the PR

kubernetes/linux/main.sh
```diff
@@ -842,6 +842,18 @@ else
...
+                  echo "Updating rsyslog config file with non default SYSLOG_HOST_PORT value ${SYSLOG_HOST_PORT}"
+                  if sed -i "s/Port=\"[0-9]*\"/Port=\"$SYSLOG_HOST_PORT\"/g" /etc/opt/microsoft/docker-cimprov/70-rsyslog-forward-mdsd-ci.conf; then
+                        echo "Successfully updated the rsylog config file."
+                  else
+                        echo "Failed to update the rsyslog config file."
+                  fi
+            else
+                  echo "SYSLOG_HOST_PORT is ${SYSLOG_HOST_PORT}. No changes made."
+            fi
+            export MDSD_DEFAULT_TCP_SYSLOG_PORT=28330
```
jatakiajanvi12, line 855: Is this different from the user given input(if any)?

kubernetes/linux/setup.sh
wanlonghenry: could you also add this in the PR description?

build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
```diff
@@ -1,5 +1,10 @@
...
 template(name="MDSD_RSYSLOG_TraditionalForwardFormat" type="string" string="<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag%%msg:::sp-if-no-1st-sp%%msg%")
-$OMUxSockDefaultTemplate MDSD_RSYSLOG_TraditionalForwardFormat
-*.* :omuxsock:
\ No newline at end of file
+
+# Forwarding all events through TCP port
+*.* action(type="omfwd"
+template="MDSD_RSYSLOG_TraditionalForwardFormat"
+queue.type="LinkedList"
+action.resumeRetryCount="-1"
```
ganga1980, line 7: Does this mean infinite retry? will this put pressure on memory if MDSD failed to ingest the syslog data?

build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
```diff
@@ -1,5 +1,10 @@
...
+queue.size="10000"
```
ganga1980, line 8: is the queue size for rsyslog service buffer ? is 10k records/events? or do we need have configurability for high syslog customers?

kubernetes/linux/setup.sh
ganga1980: yes, agree, we should use the mdsd from mariner repo through tdnf which will be safer option.

build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
pfrcks: this is only when connection breaks, so I think 10K is ok for now.

build/linux/installer/conf/70-rsyslog-forward-mdsd-ci.conf
pfrcks: no, since there is also an interval for retries which grows with the number of retries. from rsyslog docs

If multiple retries fail, the interval is automatically extended. This is to prevent excessive resource use for retries. After each 10 retries, the interval is extended by itself. To be precise, the actual interval is (numRetries / 10 + 1) * action.resumeInterval. Using the default value of 30, this means that on the 10th try the suspension interval will be 60 (seconds) and after the 100th try it will be 330 (seconds).

kubernetes/linux/main.sh
pfrcks: there is no user given input. either we use the default value or if we have toggled a specific cluster then that will be received from rp. In future once we have syslog in unified profile we will use that value to populate the variable in manifest

pfrcks:
Memory without load
memory no load

Cpu without load
cpu no load

Memory with load (~400 eps with event size 1K)
memory load

Cpu with load (~400 eps with event size 1K)
cpu load


Number of syslog records in LAW (~400 eps with event size 1K)
data reaching law

kubernetes/linux/main.sh
ganga1980, line 855: this should be **export MDSD_DEFAULT_TCP_SYSLOG_PORT=${SYSLOG_HOST_PORT}** instead of hard code port :28330

pfrcks: no, this is correct. since MDSD_DEFAULT_TCP_SYSLOG_PORT is the containerport and SYSLOG_HOST_PORT  is the hostPort. hostPort can change via toggle but containerPort will always be the same

kubernetes/linux/main.sh
```diff
@@ -842,6 +842,18 @@ else
...
+            echo "export MDSD_DEFAULT_TCP_SYSLOG_PORT=$MDSD_DEFAULT_TCP_SYSLOG_PORT" >> ~/.bashrc
```
ganga1980, line 856: does this cause any issue in non AKS scenario due to lack of container port in the manifest?

pfrcks: none since this will just be mdsd running an extra port. we are not doing any hostPort mount in non-AKS

ganga1980 (approved):

PR #1105, opened 2023-10-31
fix telegraf rs false errors

pfrcks:
When configmap is applied to the cluster TELEMETRY_RS_TELEGRAF_DISABLED true was not being written to the /opt/env_vars file which is used by liveness probe to get the env variable settings


.trivyignore
```diff
@@ -4,6 +4,10 @@ GHSA-fr2g-9hjm-wr23
...
 CVE-2023-3978
 CVE-2023-44487
+GHSA-jq35-85cj-fj4p
+
+#telegraf HIGH
+GHSA-m425-mq94-257g
 
 # ruby HIGH
-CVE-2017-10784
+CVE-2017-10784
```
ganga1980: Does this warrant for us to upgrade ruby?

pfrcks: dont know this is the same vuln we have had for some time now

pfrcks: https://github.com/microsoft/Docker-Provider/commit/d5f9bb7c5bec6a9b34d3b123726235f242026ec4#diff-a2b39041fe7520e8037b4e596d015ea4825043e383f3fa5f85d5ac7f06f4c177

jatakiajanvi12: This doesn't need ruby update. This will need fluentd upgrade because it is a dependency of fluentd.

ganga1980: Lets discuss about this in todays standup.   I think, this change can be merged.

ganga1980 (approved):

PR #1106, opened 2023-10-31
update mdsd to 1.28.11

pfrcks:
- update mdsd from 1.27.4 to 1.28.11

msi mode
memory
msi mem
cpu
msi cpu

legacy mode
memory
legacy mem
cpu
legacy cpu


kubernetes/linux/setup.sh
```diff
@@ -39,7 +39,7 @@ mv /usr/lib/ruby/gems/3.1.0/specifications/default/uri-0.11.0.gemspec /usr/lib/r
 gem uninstall time --version 0.2.0
 gem uninstall uri --version 0.11.0
 
-sudo tdnf install -y azure-mdsd-1.27.4
+sudo tdnf install -y azure-mdsd-1.28.11
```
ganga1980, line 42: Can you please check the mdsd logs and see if you are not seeing any errors in the new version? and also, can you please check no regressions with regards to syslog?  Once we get this change to CI/CD clusters, we would need to watch resource usage and data flow, and errors. Please have a task/plan to validate MDSD update related changes.

ganga1980, line 42: We should also update annotation agentVersion with MDSD version in the yaml. This annotation for the customer to check if AMA agent version. This was introduced when the critical security vulns found in base omsagent  and customer couldnt figure out which version of the agent has vulns.

pfrcks: checked the logs after running overnight. do not see any errors

pfrcks: done

charts/azuremonitor-containers/values.yaml
```diff
@@ -28,7 +28,7 @@ amalogs:
     tagWindows: "win-3.1.15"
     pullPolicy: IfNotPresent
     dockerProviderVersion: "18.0.1-0"
-    agentVersion: "azure-mdsd-1.17.0"
```
jatakiajanvi12, line 31: Will we need this in the aks manifest as well?

pfrcks: yes we need to update the aks manifest as well. can do it with the release aks-rp pr

ganga1980 (dismissed):

charts/azuremonitor-containers-geneva/values.yaml
```diff
@@ -17,7 +17,7 @@ image:
   repository: mcr.microsoft.com/azuremonitor/containerinsights/ciprod
   tag: "3.1.15"
   pullPolicy: IfNotPresent
-  agentVersion: "azure-mdsd-1.17.1"
+  agentVersion: "azure-mdsd-1.28.11"
```
wanlonghenry, line 20: @pfrcks  looking at the attached screenshots, does the first screenshot mean that there is memory increase with MSI mode with latest mdsd version?

pfrcks: its only a slight increase so can be ignored. will observe more closely in ci/cd clusters

ganga1980 (dismissed):

ganga1980 (approved):

PR #1108, opened 2023-11-01
add user agent to KubeAPI and CAdvisor clients

pfrcks:
- Sets User-Agent as "ama-logs/<version> (<os_type>; <platform>)". Eg for 3.1.15 release User-Agent is "ama-logs/3.1.15 (linux; Ruby x86_64-linux)" for Linux nodes
- Adds User-Agent to Kube API and CAdvisor API calls


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -97,6 +99,7 @@ def getKubeResourceInfoV2(resource, api_group: nil)
             Net::HTTP.start(uri.host, uri.port, :use_ssl => true, :ca_file => @@CaFile, :verify_mode => OpenSSL::SSL::VERIFY_PEER, :open_timeout => 20, :read_timeout => 40) do |http|
               kubeApiRequest = Net::HTTP::Get.new(uri.request_uri)
               kubeApiRequest["Authorization"] = "Bearer " + getTokenStr
+              kubeApiRequest["User-Agent"] = @@userAgent
               @Log.info "KubernetesAPIClient::getKubeResourceInfoV2 : Making request to #{uri.request_uri} @ #{Time.now.utc.iso8601}"
               response = http.request(kubeApiRequest)
```
ganga1980, line 104: we should add useragent for watch api calls also - https://github.com/microsoft/Docker-Provider/blob/ci_prod/source/plugins/ruby/KubernetesApiClient.rb#L944

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -62,6 +62,8 @@ class CAdvisorMetricsAPIClient
   @@telemetryCpuMetricTimeTracker = DateTime.now.to_time.to_i
   @@telemetryMemoryMetricTimeTracker = DateTime.now.to_time.to_i
   @@telemetryPVKubeSystemMetricsTimeTracker = DateTime.now.to_time.to_i
+  @@userAgent = "ama-logs-" + (ENV["AGENT_VERSION"].nil? ? "" : ENV["AGENT_VERSION"])
```
ganga1980: (1) I think,  we should follow standard naming convention of the user agent  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent.  In this case, ama-logs/AgentVersion instead of ama-logs-AgentVersion
(2) should we include the OS information also, for example ama-logs/AgentVersion(Linux)  or ama-logs/AgentVersion(Windows) ?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -943,6 +946,7 @@ def watch(resource_name, options = {})
         }
         http_headers = {
           Authorization: "Bearer " + getTokenStr,
+          "User-Agent": @@userAgent,
```
ganga1980: can you please check in AKS API telemetry for your clusters and see User-Agent value being emitted for kube api calls (watch and non-watch) APIs?

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -62,6 +62,8 @@ class CAdvisorMetricsAPIClient
   @@telemetryCpuMetricTimeTracker = DateTime.now.to_time.to_i
   @@telemetryMemoryMetricTimeTracker = DateTime.now.to_time.to_i
   @@telemetryPVKubeSystemMetricsTimeTracker = DateTime.now.to_time.to_i
+  @@userAgent = "ama-logs/#{ENV['AGENT_VERSION'].nil? ? '0.0.0' : ENV['AGENT_VERSION']} (#{ENV['OS_TYPE'].nil? ? 'linux' : ENV['OS_TYPE']}; Ruby #{RUBY_PLATFORM})"
```
ganga1980: wrap this code to function and use that function. define this function in KubernetesAPIClient so that we only have single function to use every where.

pfrcks: i wrote this here to avoid multiple function calls

ganga1980 (approved):

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -221,6 +224,16 @@ def getClusterId(env = ENV)
...
     end
 
+    def getUserAgent()
+      return @@userAgent if !@@userAgent.nil?
+      begin
+          @@userAgent = "ama-logs/#{ENV['AGENT_VERSION'].nil? ? '0.0.0' : ENV['AGENT_VERSION']} (#{ENV['OS_TYPE'].nil? ? 'linux' : ENV['OS_TYPE']}; Ruby #{RUBY_PLATFORM})"
+      rescue => error
+        @Log.warn("KubernetesAPIClient::getUserAgent : getUserAgent failed: #{error}")
+      end
+      return "ama-logs/0.0.0 ()"
```
ganga1980: actually, this is bug. The return value should be @@userAgent otherwise very first case, this will be "ama-logs/0.0.0 ()"

ganga1980 (changes requested):
see latest comment

ganga1980 (approved):

PR #1110, opened 2023-11-02
Gangams/dotnet core version update

ganga1980:
update - upgrading dotnet version from dotnetcoreapp3.1 to dotnet 7.0 since dotnetcoreapp3.1 end of life. 
validated - on clusters with both windows server 2019 and windows server 2022 nodes.

### DIFF
--- a/build/windows/Makefile.ps1
+++ b/build/windows/Makefile.ps1
@@ -5,7 +5,7 @@
 #  3. copy the files under installer directory to ..\..\kubernetes\windows\amalogswindows
 #  4. Builds the livenessprobe cpp and copy the executable to the under directory ..\..\kubernetes\windows\amalogswindows
 
-$dotnetcoreframework = "net7.0"
+$dotnetcoreframework = "net6.0"
 
 Write-Host("Building Certificate generator code...")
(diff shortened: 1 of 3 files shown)


build/windows/Makefile.ps1
```diff
@@ -5,7 +5,7 @@
 #  3. copy the files under installer directory to ..\..\kubernetes\windows\amalogswindows
 #  4. Builds the livenessprobe cpp and copy the executable to the under directory ..\..\kubernetes\windows\amalogswindows
 
-$dotnetcoreframework = "netcoreapp3.1"
+$dotnetcoreframework = "net7.0"
```
jatakiajanvi12: Should we move to net6.0? That has long term support as compared to 7.0 https://dotnet.microsoft.com/en-us/platform/support/policy

ganga1980: > Should we move to net6.0? That has long term support as compared to 7.0 https://dotnet.microsoft.com/en-us/platform/support/policy

Thanks for catching this. Make sense to go with net6.0 considering its LTS. updated.

jatakiajanvi12 (approved):

PR #1113, opened 2023-11-09
Add perf and container inventory plugins for Windows and integrating them to named pipe

jatakiajanvi12:


.gitignore
```diff
@@ -31,4 +31,8 @@ source/plugins/go/src/*.h
 *.log
 *.byebug_history
 go.work.sum
-go.work
\ No newline at end of file
+go.work
+#npm modules and build output for the typescript project
```
pfrcks: the appmonitoring changes are no longer needed

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -91,6 +96,7 @@ def substituteFluentBitPlaceHolders
 
     new_contents = substituteMultiline(multilineLogging, new_contents)
     new_contents = substituteResourceOptimization(resoureceOptimizationEnabled, new_contents)
+    puts new_contents
```
pfrcks: leftover debug statement?

build/windows/installer/conf/fluent-bit-common.conf
```diff
@@ -56,3 +56,11 @@
     Retry_Limit                     10
     TelemetryPushIntervalSeconds    300
     Match                           oms.container.*
+
```
pfrcks, line 59: I am wondering whether we should change Tags to keep just 1 output section? Will it be more performant to have a single output oms rather than two?

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -253,6 +253,7 @@ func GetMetrics(winNode map[string]string, namespaceFilteringMode string, namesp
 
 func GetMetricsHelper(metricInfo map[string]interface{}, winNode map[string]string, namespaceFilteringMode string, namespaces []string, metricTime string) []metricDataItem {
 	var metricDataItems []metricDataItem
+	osType := strings.TrimSpace(os.Getenv("OS_TYPE"))
```
pfrcks: we already have a osType in init. Maybe we can make it global and move this logic there?

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -267,12 +268,15 @@ func GetMetricsHelper(metricInfo map[string]interface{}, winNode map[string]stri
...
+			operatingSystem = "Windows"
+		} else {
+			operatingSystem = "Linux"
+		}
 	}
 
-	osType := strings.TrimSpace(os.Getenv("OS_TYPE"))
 	// Checking if we are in windows daemonset and sending only a few metrics that are needed for MDM
-	if osType != "" && strings.EqualFold(osType, "windows") {
+	if osType != "" && strings.EqualFold(osType, "windows") && !IsAADMSIAuthMode() {
```
pfrcks, line 280: not sure I understand this. Windows fluent-bit input plugins are only for MSI mode. In that case what are we collecting here?

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -649,7 +653,7 @@ func getNodeMetricItemRate(metricInfo map[string]interface{}, hostName, metricGr
 				linuxNodePrevMetricRate = metricRateValue
 				metricValue = metricRateValue
 			}
-		} else if operatingSystem == "Windows" {
+		} else if operatingSystem == "Windows" { // For Windows with non-MSI mode
```
pfrcks, line 657: again, is this even a valid scenario ?

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -625,7 +629,7 @@ func getNodeMetricItemRate(metricInfo map[string]interface{}, hostName, metricGr
 		return nil
 	} else {
 		metricRateValue := 0.0
-		if operatingSystem == "Linux" {
+		if operatingSystem == "Linux" || IsAADMSIAuthMode() { // For Linux and Windows with MSI mode
```
pfrcks: can we make this explicit? if (os is linux) || (os is windows && isMsi)

source/plugins/go/input/cmetrics_windows.go
```diff
@@ -0,0 +1,315 @@
+//go:build windows
```
pfrcks, line 1: any progress on the PR on cmetrics to include this?

source/plugins/go/src/oms.go
```diff
@@ -1177,51 +1179,82 @@ func PostInputPluginRecords(inputPluginRecords []map[interface{}]interface{}) in
 			msgPackEntries = append(msgPackEntries, msgPackEntry)
 		}
 
-		if !IsWindows && len(msgPackEntries) > 0 { //for linux, mdsd route
-			Log("Info::mdsd:: using mdsdsource name for input plugin records: %s", tag)
+		if (!IsWindows || (IsWindows && IsAADMSIAuthMode)) && len(msgPackEntries) > 0 {
```
pfrcks: can we simplify this? we can check for len(msgPackEntries) seperately as a precheck

source/plugins/go/src/oms.go
```diff
@@ -1177,51 +1179,82 @@ func PostInputPluginRecords(inputPluginRecords []map[interface{}]interface{}) in
...
-						telemetryDimensions["PerfCount"] = strconv.Itoa(len(msgPackEntries))
-					} else {
-						Log("FlushInputPluginRecords::Warn::Flushed records of unknown data type %s", lowerTag)
-					}
-					numRecords := len(msgPackEntries)
-					Log("FlushInputPluginRecords::Info::Successfully flushed %d records that was %d bytes in %s", numRecords, bts, elapsed)
-					// Send telemetry to AppInsights resource
-					SendEvent(InputPluginRecordsFlushedEvent, telemetryDimensions)
+					Log("Error::mdsd::Unable to create mdsd client for InputPluginRecords. Please check error log.")
 				}
```
pfrcks: can we refactor this? looks like we have multiple instances of if Linux do this and if not do that. Instead is it possible to have one block of if Linux and one block for if Windows?

source/plugins/go/input/lib/cadvisor.go
jatakiajanvi12: I think this is dead code. It is present in the ruby plugins for differentiating what and how RS collects the data. Do you think I should just remove this code? @pfrcks

source/plugins/go/input/lib/cadvisor.go
jatakiajanvi12: Same as above

source/plugins/go/input/cmetrics_windows.go
jatakiajanvi12: No progress...I think they mentioned something in slack...Will try to address that

source/plugins/go/src/oms.go
jatakiajanvi12: Do you just want to me to create another level of if?

source/plugins/go/src/oms.go
jatakiajanvi12: That would mean a lot of code which is similar in both have to be repeated multiple times in the code? Are we okay with that ?

build/windows/installer/conf/fluent-bit-common.conf
jatakiajanvi12: Haven't done analysis on the same. But may be we can try and see what is the perf difference once we have all the changes in?

pfrcks: yeah let's add this to our TODO list

source/plugins/go/input/containerinventory/containerinventory.go
```diff
@@ -165,7 +165,9 @@ func (p containerInventoryPlugin) enumerate() []map[string]interface{} {
 					container[k] = v
 				}
 				container["State"] = "Deleted"
-				lib.DeleteCGroupCacheEntryForDeletedContainer(container["InstanceID"].(string))
+				if strings.EqualFold(osType, "windows") {
```
pfrcks: why only for windows?

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -71,10 +71,12 @@ var (
 	Log                                     *logrus.Logger
 )
 
+var osType string
```
pfrcks: pls transfer to the above var block

source/plugins/go/input/lib/cadvisor.go
pfrcks: yeah lets remove. just double check that this does not have any other effects

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -625,7 +630,7 @@ func getNodeMetricItemRate(metricInfo map[string]interface{}, hostName, metricGr
 		return nil
 	} else {
 		metricRateValue := 0.0
-		if operatingSystem == "Linux" {
+		if operatingSystem == "Linux" || (operatingSystem == "Windows" && IsAADMSIAuthMode()) { // For Linux and Windows with MSI mode
```
pfrcks, line 633: here as well

source/plugins/go/input/lib/kubernetescontainerinventory.go
```diff
@@ -225,7 +224,7 @@ func getContainersInfoMap(podItem map[string]interface{}, isWindows bool) map[st
 				}
 				containerInfoMap["Command"] = cmdValueString
 
-				if isWindows {
+				if isWindows && !IsAADMSIAuthMode() {
```
pfrcks, line 228: is this a valid scenario?

source/plugins/go/src/oms.go
pfrcks: something like
if len(msgpackentries) == 0 continue

if !Windows || Windows && MSI {
	...
}

source/plugins/go/src/oms.go
pfrcks: good pt. I'll leave it upto you. If its possible to refactor it to make it more readible please do it otherwise its ok.

source/plugins/go/input/lib/cadvisor.go
jatakiajanvi12: I was thinking lets keep it as is for now and then later we can remove when everything is set up and we know what default exactly looks like? wdyt?

source/plugins/go/input/lib/cadvisor.go
jatakiajanvi12: what do we need here ?

source/plugins/go/input/lib/kubernetescontainerinventory.go
jatakiajanvi12: DEAD CODE - will be removed later

source/plugins/go/input/containerinventory/containerinventory.go
jatakiajanvi12: Ohh good catch. It had to be only for Linux as windows doesnt have cgroup cache

source/plugins/go/src/oms.go
jatakiajanvi12: Refactored the code

source/plugins/go/input/lib/cadvisor.go
jatakiajanvi12: DEAD CODE - will be removed later

source/plugins/go/input/lib/cadvisor.go
jatakiajanvi12: DEAD CODE - will be removed later

pfrcks: sure. let's add a TODO comment here before resolving

source/plugins/go/input/lib/cadvisor.go
pfrcks: please add a TODO before resolving

build/windows/installer/conf/fluent-bit-common.conf
pfrcks: pls add a TODO comment for this

source/plugins/go/input/lib/kubernetescontainerinventory.go
pfrcks: pls add a TODO comment for this

pfrcks (approved):

PR #1115, opened 2023-11-09
add fluent-bit kubernetes filter

pfrcks:
- Uses fluent-bit Kubernetes filter to capture kubernetes metadata for container log
- Enabled via configmap. Default disabled

### DIFF
--- a/build/common/installer/scripts/tomlparser.rb
+++ b/build/common/installer/scripts/tomlparser.rb
@@ -24,6 +24,7 @@ require_relative "ConfigParseErrorLogger"
 @containerLogsRoute = "v2" # default for linux
 @adxDatabaseName = "containerinsights" # default for all configurations
 @logEnableMultiline = "false"
+@logEnableKubernetesMetadata = false
 if !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp("windows") == 0
   @containerLogsRoute = "v1" # default is v1 for windows until windows agent integrates windows ama
   # This path format is necessary for fluent-bit in windows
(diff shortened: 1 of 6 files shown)


build/common/installer/scripts/tomlparser.rb
```diff
@@ -205,6 +206,17 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
       ConfigParseErrorLogger.logError("Exception while reading config map settings for adx database name - #{errorStr}, using default #{@adxDatabaseName}, please check config map for errors")
     end
+
+    #Get Kubernetes Metadata setting
+    begin
+      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
+        @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
+        puts "config::Using config map setting for kubernetes metadata"
+
+        if @containerLogSchemaVersion.strip.casecmp("v2") != 0
```
wanlonghenry: if customer did not set containerlog schema but enabled it through DCR, we still need to support this feature

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -296,6 +297,9 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
                     if (!@clusterMultilineEnabled.nil? && !@clusterMultilineEnabled.empty?)
                       telemetryProps["multilineEnabled"] = @clusterMultilineEnabled
                     end
+                    if (!@clusterKubernetesMetadataEnabled.nil? && !@clusterKubernetesMetadataEnabled.empty?)
+                      telemetryProps["metadataEnabled"] = @clusterKubernetesMetadataEnabled
```
wanlonghenry, line 301: what are the following changes regarding this CAdvisorMetrics telemetry?

build/common/installer/scripts/tomlparser.rb
pfrcks: i can remove this check from here assuming we are doing the check later in oms.go

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
pfrcks: i am not sure I understand

wanlonghenry: I mean are there any changes needed in other places for this to collect this telemetry? Do we also need to collect telemetry such as labels and annotations when we added those later?

pfrcks: we can add those later. this is just to report whether filter is enabled or not

wanlonghenry (approved):

PR #1118, opened 2023-11-15
Gangams/fix arc k8s windows custom metrics bug

ganga1980:
For Arc k8s, in case of legacy auth mode, windows custom metrics  not ingested because of the KUBERNETES_SERVICE_HOST environment variable not available in ruby code. Fix is to set the environment variable in main.ps1 and also removed KubernetesApiClient dependency  in arc_k8s_cluster_identity.rb.

### DIFF
--- a/source/plugins/ruby/arc_k8s_cluster_identity.rb
+++ b/source/plugins/ruby/arc_k8s_cluster_identity.rb
@@ -6,7 +6,6 @@ require "uri"
 require "json"
 require "base64"
 require "time"
-require_relative "KubernetesApiClient"
 require_relative "ApplicationInsightsUtility"
 
 class ArcK8sClusterIdentity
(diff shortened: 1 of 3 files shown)


source/plugins/ruby/arc_k8s_cluster_identity.rb
```diff
@@ -6,7 +6,6 @@
 require "json"
 require "base64"
 require "time"
-require_relative "KubernetesApiClient"
```
wanlonghenry, line 9: image
do you want to change all places that had this reference?

source/plugins/ruby/arc_k8s_cluster_identity.rb
```diff
@@ -34,7 +33,8 @@ def initialize
     @isLastTokenRenewalUpdatePending = false
     @token_file_path = "/var/run/secrets/kubernetes.io/serviceaccount/token"
     @cert_file_path = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
-    @kube_api_server_url = KubernetesApiClient.getKubeAPIServerUrl
```
jatakiajanvi12, line 37: Can we remove this function completely from the source file? I dont think we use this anywhere else

source/plugins/ruby/arc_k8s_cluster_identity.rb
ganga1980: No, only this file. In this file, we are only using KubernetesApiClient for getKubeAPIServerUrl method.  Removed the dependency of KubernetesApiClient in this file to avoid any issues around dependencies.

jatakiajanvi12 (approved):

wanlonghenry (approved):

PR #1120, opened 2023-11-17
add metadata feilds including labels, annotations, uid and image

wanlonghenry:
add metadata feilds including labels, annotations, uid and image
default include all the feilds

### DIFF
--- a/build/common/installer/scripts/tomlparser.rb
+++ b/build/common/installer/scripts/tomlparser.rb
@@ -25,7 +25,7 @@ require_relative "ConfigParseErrorLogger"
 @adxDatabaseName = "containerinsights" # default for all configurations
 @logEnableMultiline = "false"
 @logEnableKubernetesMetadata = false
-@logKubernetesMetadataiIncludeFields = "podLabels,podAnnotations,poduid,image"
+@logKubernetesMetadataiIncludeFields = "podLabels,podAnnotations,podUid,image"
 if !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp("windows") == 0
   @containerLogsRoute = "v1" # default is v1 for windows until windows agent integrates windows ama
   # This path format is necessary for fluent-bit in windows
@@ -210,7 +210,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
 
     #Get Kubernetes Metadata setting
     begin
-      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
+      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].nil?
         @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
         @logKubernetesMetadataiIncludeFields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields]
         puts "config::Using config map setting for kubernetes metadata"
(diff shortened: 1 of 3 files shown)


build/common/installer/scripts/tomlparser.rb
```diff
@@ -211,6 +212,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
     begin
       if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
         @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
+        @logKubernetesMetadataiIncludeFields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields]
```
pfrcks: pls add a nil check. if a customer does not specify the list then we should not collect any metadata

build/common/installer/scripts/tomlparser.rb
```diff
@@ -25,6 +25,7 @@
 @adxDatabaseName = "containerinsights" # default for all configurations
 @logEnableMultiline = "false"
 @logEnableKubernetesMetadata = false
+@logKubernetesMetadataiIncludeFields = "podLabels,podAnnotations,poduid,image"
```
pfrcks: podUID

wanlonghenry: change all to podUid to sync with othe places: 
image

build/common/installer/scripts/tomlparser.rb

build/common/installer/scripts/tomlparser.rb
```diff
@@ -209,8 +210,9 @@ def populateSettingValuesFromConfigMap(parsedConfig)
 
     #Get Kubernetes Metadata setting
     begin
-      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
+      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].nil?
         @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
+        @logKubernetesMetadataiIncludeFields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields]
```
pfrcks: looks good however we might be missing the default scenario.

1) customer enables metadata collection and does not specify field -> default scenario where we collect all field
2) customer enables metadata collection and does specify field -> we use the include_fields property.

as of now only option 2 is tackled

wanlonghenry: I think you are talking about the scenario where enabled is true but include_fields is commented out. We can add a scenario where enabled is true and parsedConfig[:log_collection_settings][:metadata_collection][:include_fields] is nil.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -211,6 +212,8 @@ def populateSettingValuesFromConfigMap(parsedConfig)
     begin
       if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
         @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
+        if !parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].nil?
+          @logKubernetesMetadataiIncludeFields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields]
```
pfrcks, line 216: missing end for the if block. also add a puts statement inside the block "config::Using config map setting for kubernetes metadata include fields"

build/common/installer/scripts/tomlparser.rb
```diff
@@ -211,6 +212,8 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
         puts "config::Using config map setting for kubernetes metadata"
```
pfrcks, line 217: shift this line to above line 214

PR #1121, opened 2023-11-20
Add telegraf enabled/disabled info in TelegrafMetricsSentCount

jatakiajanvi12:
Add telegraf enabled/disabled info in TelegrafMetricsSentCount

### DIFF
--- a/kubernetes/windows/main.ps1
+++ b/kubernetes/windows/main.ps1
@@ -649,6 +649,15 @@ function Get-ContainerRuntime {
 
 function Start-Fluent-Telegraf {
 
+    Set-ProcessAndMachineEnvVariables "TELEMETRY_CUSTOM_PROM_MONITOR_PODS" "false"
+    # run prometheus custom config parser
+    Write-Host "**********Running config parser for custom prometheus scraping**********"
+    ruby /opt/amalogswindows/scripts/ruby/tomlparser-prom-customconfig.rb
+
+    Set-EnvironmentVariablesFromFile "/opt/amalogswindows/scripts/powershell/setpromenv.txt"
+
+    Write-Host "**********End running config parser for custom prometheus scraping**********"
+
     $containerRuntime = Get-ContainerRuntime
 
     if (![string]::IsNullOrEmpty($containerRuntime) -and [string]$containerRuntime.StartsWith('docker') -eq $false) {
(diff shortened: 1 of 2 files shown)


kubernetes/windows/main.ps1
```diff
@@ -649,6 +649,15 @@ function Get-ContainerRuntime {
 
 function Start-Fluent-Telegraf {
 
+    Set-ProcessAndMachineEnvVariables "TELEMETRY_CUSTOM_PROM_MONITOR_PODS" "false"
```
wanlonghenry, line 652: does this mean that even SIDECAR_SCRAPING_ENABLED set as disabled, those variables will be set?

source/plugins/go/src/telemetry.go
```diff
@@ -254,7 +254,12 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 				TelemetryClient.Track(logLatencyMetric)
 			}
 		}
-		TelemetryClient.Track(appinsights.NewMetricTelemetry(metricNameNumberofTelegrafMetricsSentSuccessfully, telegrafMetricsSentCount))
+		telegrafEnabled := make(map[string]string)
```
wanlonghenry, line 257: the reason why we use a map instead of boolean is because we might expand this later?

jatakiajanvi12: Yes if we want to expand anything we can do that

source/plugins/go/src/telemetry.go
```diff
@@ -254,7 +254,12 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
...
+		osType := os.Getenv("OS_TYPE")
+		if osType != "" && strings.EqualFold(osType, "windows") {
+			telegrafEnabled["IsTelegrafEnabled"] = os.Getenv("TELEMETRY_CUSTOM_PROM_MONITOR_PODS") // If TELEMETRY_CUSTOM_PROM_MONITOR_PODS, then telegraf is enabled
```
ganga1980, line 260: since this specific to Windows and TelegrafMetricsSentCount used for both Linux and Windows, we should rename from IsTelegrafEnabled to IsTelegrafEnabledInWindows or something we can identify this field for windows specific.   resolving this comment since we already have Agent version and computer field dimensions which we can use.

ganga1980 (approved):

kubernetes/windows/main.ps1
jatakiajanvi12: Yes I think that is alright because we will need the TELEMETRY_CUSTOM_PROM_MONITOR_PODS always. Let me know if you think otherwise

wanlonghenry (approved):

PR #1122, opened 2023-11-21
Telemetry optimization - Add K8sAPI and CAdvisorAPI response code telemetry in customMetrics 

jatakiajanvi12:
Adding telemetry customMetrics every telemetry interval for all the Kubernetes API and Cadvisor API response codes. 

Eg: K8sAPIStatus and CAdvisorAPIStatus

### DIFF
--- a/source/plugins/ruby/ApplicationInsightsUtility.rb
+++ b/source/plugins/ruby/ApplicationInsightsUtility.rb
@@ -358,6 +358,7 @@ class ApplicationInsightsUtility
             apiResponseCodeHash.clear
           end
         end
+        return apiResponseTelemetryTimeTracker
       rescue => err
         $log.warn("Exception in AppInsightsUtility: sendAPIResponseTelemetry failed with an error: #{err}")
       end
--- a/source/plugins/ruby/ApplicationInsightsUtility.rb
+++ b/source/plugins/ruby/ApplicationInsightsUtility.rb
@@ -326,7 +326,7 @@ class ApplicationInsightsUtility
     def sendAPIResponseTelemetry(responseCode, resource, metricName, apiResponseCodeHash, apiResponseTelemetryTimeTracker)
       begin
         if (!responseCode.nil? && !responseCode.empty?)
-          if (apiResponseCodeHash.has_key?(responseCode))
+          if (!apiResponseCodeHash.has_key?(responseCode))
             telemetryProps = {}
             telemetryProps[resource] = 1
             apiResponseCodeHash[responseCode] = telemetryProps
--- a/source/plugins/ruby/ApplicationInsightsUtility.rb
+++ b/source/plugins/ruby/ApplicationInsightsUtility.rb
@@ -323,5 +323,43 @@ class ApplicationInsightsUtility
         $log.warn("Exception in AppInsightsUtility: getWorkspaceCloud - error: #{errorStr}")
       end
     end
+
+    def sendAPIResponseTelemetry(responseCode, resource, metricName)
+      begin
+        if (!responseCode.nil? && !responseCode.empty?)
+          if (@@kubernetesApiResponseCodeHash.has_key?(response.code))
+            telemetryProps = {}
+            telemetryProps[resource] = 1
+            @@kubernetesApiResponseCodeHash[responseCode] = telemetryProps
+          else
+            telemetryProps = @@kubernetesApiResponseCodeHash[responseCode]
+            if (!telemetryProps.has_key?(resource))
+              telemetryProps[resource] = 1
+            else
+              telemetryProps[resource] += 1
+            end
+            @@kubernetesApiResponseCodeHash[responseCode] = telemetryProps
+          end
+
+          timeDifference = (DateTime.now.to_time.to_i - @@kubernetesApiResponseTelemetryTimeTracker).abs
+          timeDifferenceInMinutes = timeDifference / 60
+          if (timeDifferenceInMinutes >= Constants::TELEMETRY_FLUSH_INTERVAL_IN_MINUTES)
+            @@kubernetesApiResponseTelemetryTimeTracker = DateTime.now.to_time.to_i
+            @@kubernetesApiResponseCodeHash.each do |key, value|
+              value.each do |resource, count|
+                telemetryProps = {}
+                telemetryProps["Resource"] = resource
+                telemetryProps["ResponseCode"] = key
+                sendMetricTelemetry(metricName, count, telemetryProps)
+              end
+            end
+            @@kubernetesApiResponseCodeHash.clear
+          end
+        end
+      rescue => err
+        @Log.warn "KubernetesApiClient::sendKubernetesAPIResponseTelemetry failed with an error: #{err}"
+      end
+    end
+
   end
 end
--- a/source/plugins/ruby/ApplicationInsightsUtility.rb
+++ b/source/plugins/ruby/ApplicationInsightsUtility.rb
@@ -39,9 +39,6 @@ class ApplicationInsightsUtility
   end
 
   @@controllerType = {"daemonset" => "DS", "replicaset" => "RS"}
-  @@apiResponseCodeHash = {}
-  @@apiResponseTelemetryTimeTracker = DateTime.now.to_time.to_i
-
   def initialize
   end
 
@@ -326,15 +323,15 @@ class ApplicationInsightsUtility
       end
     end
 
-    def sendAPIResponseTelemetry(responseCode, resource, metricName)
+    def sendAPIResponseTelemetry(responseCode, resource, metricName, apiResponseCodeHash, apiResponseTelemetryTimeTracker)
       begin
         if (!responseCode.nil? && !responseCode.empty?)
-          if (@@apiResponseCodeHash.has_key?(responseCode))
+          if (apiResponseCodeHash.has_key?(responseCode))
             telemetryProps = {}
             telemetryProps[resource] = 1
-            @@apiResponseCodeHash[responseCode] = telemetryProps
+            apiResponseCodeHash[responseCode] = telemetryProps
           else
-            telemetryProps = @@apiResponseCodeHash[responseCode]
+            telemetryProps = apiResponseCodeHash[responseCode]
             if (telemetryProps.nil?)
               telemetryProps = {}
             end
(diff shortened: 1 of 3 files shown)


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -70,6 +72,7 @@ def getKubeResourceInfo(resource, api_group: nil)
               @Log.info "KubernetesAPIClient::getKubeResourceInfo : Making request to #{uri.request_uri} @ #{Time.now.utc.iso8601}"
               response = http.request(kubeApiRequest)
               @Log.info "KubernetesAPIClient::getKubeResourceInfo : Got response of #{response.code} for #{uri.request_uri} @ #{Time.now.utc.iso8601}"
+              sendKubernetesAPIResonseTelemetry(response)
```
ganga1980: Instead of passing entire response which can be very large, juts pass the response code

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -1521,6 +1525,30 @@ def sendReplicasetAgentRequestsAndLimitsTelemetry(podName, podNameSpace, contain
...
+          else
+            @@kubernetesApiResponseCodeHash[response.code] += 1
+          end
+
+          timeDifference = (DateTime.now.to_time.to_i - @@kubernetesApiResponseTelemetryTimeTracker).abs
+          timeDifferenceInMinutes = timeDifference / 60
+          if (timeDifferenceInMinutes >= Constants::TELEMETRY_FLUSH_INTERVAL_IN_MINUTES)
+            @@kubernetesApiResponseTelemetryTimeTracker = DateTime.now.to_time.to_i
+            @@kubernetesApiResponseCodeHash.each do |key, value|
+              ApplicationInsightsUtility.sendMetricTelemetry("KubernetesApiResponseCode_#{key}", value, {})
```
ganga1980: We should also add the dimension of the API call for example "/pods" or "/nodes" or "/events" so we know which API call has the issue.

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -1521,6 +1525,30 @@ def sendReplicasetAgentRequestsAndLimitsTelemetry(podName, podNameSpace, contain
       end
     end
 
+    def sendKubernetesAPIResonseTelemetry(response)
```
wanlonghenry: typo, maybe better to be 'sendKubernetesAPIResponseTelemetry'?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -1521,6 +1525,30 @@ def sendReplicasetAgentRequestsAndLimitsTelemetry(podName, podNameSpace, contain
...
+      begin
+        if (!response.nil? && !response.code.nil? && !response.code.empty?)
+          if (@@kubernetesApiResponseCodeHash[response.code].nil?)
```
ganga1980: instead of nil check, https://apidock.com/ruby/Hash/has_key%3F is better readable.

source/plugins/ruby/KubernetesApiClient.rb
ganga1980: rename from KubernetesApiResponseCode to K8sAPIStatus?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -323,5 +325,46 @@ def getWorkspaceCloud()
         $log.warn("Exception in AppInsightsUtility: getWorkspaceCloud - error: #{errorStr}")
       end
     end
+
+    def sendAPIResponseTelemetry(responseCode, resource, metricName)
+      begin
+        if (!responseCode.nil? && !responseCode.empty?)
+          if (@@apiResponseCodeHash.has_key?(responseCode))
```
ganga1980: this should be **not** of the has_key, right?
!@@apiResponseCodeHash.has_key?(responseCode)

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -323,5 +325,46 @@ def getWorkspaceCloud()
...
+            else
+              telemetryProps[resource] += 1
+            end
+            @@apiResponseCodeHash[responseCode] = telemetryProps
+          end
+
+          timeDifference = (DateTime.now.to_time.to_i - @@apiResponseTelemetryTimeTracker).abs
+          timeDifferenceInMinutes = timeDifference / 60
+          if (timeDifferenceInMinutes >= Constants::TELEMETRY_FLUSH_INTERVAL_IN_MINUTES)
+            @@apiResponseTelemetryTimeTracker = DateTime.now.to_time.to_i
```
ganga1980: can you please validate this change thoroughly?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -323,5 +325,46 @@ def getWorkspaceCloud()
...
+            @@apiResponseCodeHash.each do |key, value|
+              value.each do |resource, count|
+                telemetryProps = {}
+                telemetryProps["Resource"] = resource
+                telemetryProps["ResponseCode"] = key
+                sendMetricTelemetry(metricName, count, telemetryProps)
```
ganga1980, line 355: Did you deploy this change to test cluster to validate the ingested telemetry ?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -323,5 +322,46 @@ def getWorkspaceCloud()
         $log.warn("Exception in AppInsightsUtility: getWorkspaceCloud - error: #{errorStr}")
       end
     end
+
+    def sendAPIResponseTelemetry(responseCode, resource, metricName, apiResponseCodeHash, apiResponseTelemetryTimeTracker)
+      begin
+        if (!responseCode.nil? && !responseCode.empty?)
+          if (apiResponseCodeHash.has_key?(responseCode))
```
ganga1980: this needs to be !apiResponseCodeHash.has_key?(responseCode), right?

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -323,5 +322,46 @@ def getWorkspaceCloud()
...
+          end
+
+          timeDifference = (DateTime.now.to_time.to_i - apiResponseTelemetryTimeTracker).abs
+          timeDifferenceInMinutes = timeDifference / 60
+          if (timeDifferenceInMinutes >= Constants::TELEMETRY_FLUSH_INTERVAL_IN_MINUTES)
+            apiResponseTelemetryTimeTracker = DateTime.now.to_time.to_i
+            apiResponseCodeHash.each do |key, value|
+              value.each do |resource, count|
+                telemetryProps = {}
+                telemetryProps["Resource"] = resource
```
ganga1980: what would be resource names here? is it "pods", "nodes" etc..?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -70,6 +73,7 @@ def getKubeResourceInfo(resource, api_group: nil)
               @Log.info "KubernetesAPIClient::getKubeResourceInfo : Making request to #{uri.request_uri} @ #{Time.now.utc.iso8601}"
               response = http.request(kubeApiRequest)
               @Log.info "KubernetesAPIClient::getKubeResourceInfo : Got response of #{response.code} for #{uri.request_uri} @ #{Time.now.utc.iso8601}"
+              ApplicationInsightsUtility.sendAPIResponseTelemetry(response.code, resource, "K8sAPIStatus", @@K8sApiResponseCodeHash, @@K8sApiResponseTelemetryTimeTracker)
```
ganga1980: there might scenario where response can be nil hence can we add check !response.nil? & !response.code.empty? before this call - ApplicationInsightsUtility.sendAPIResponseTelemetry

source/plugins/ruby/ApplicationInsightsUtility.rb
jatakiajanvi12: Yes I have

source/plugins/ruby/ApplicationInsightsUtility.rb
jatakiajanvi12: That would the resource uri: /pods, /stats/summary, etc

source/plugins/ruby/KubernetesApiClient.rb
jatakiajanvi12: The check is present in the function itself where we are checking if the responseCode is present or no

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -104,6 +108,7 @@ def getKubeResourceInfoV2(resource, api_group: nil)
               response = http.request(kubeApiRequest)
               responseCode = response.code
               @Log.info "KubernetesAPIClient::getKubeResourceInfoV2 : Got response of #{response.code} for #{uri.request_uri} @ #{Time.now.utc.iso8601}"
+              @@K8sApiResponseTelemetryTimeTracker = ApplicationInsightsUtility.sendAPIResponseTelemetry(response.code, resource, "K8sAPIStatus", @@K8sApiResponseCodeHash, @@K8sApiResponseTelemetryTimeTracker)
```
ganga1980: why I am seeing only this API call in the telemetry? 
image

source/plugins/ruby/ApplicationInsightsUtility.rb
```diff
@@ -323,5 +322,47 @@ def getWorkspaceCloud()
         $log.warn("Exception in AppInsightsUtility: getWorkspaceCloud - error: #{errorStr}")
       end
     end
+
+    def sendAPIResponseTelemetry(responseCode, resource, metricName, apiResponseCodeHash, apiResponseTelemetryTimeTracker)
+      begin
+        if (!responseCode.nil? && !responseCode.empty?)
```
ganga1980, line 328: For optimizing telemetry, I think, we should avoid emitting telemetry for success scenarios (>=200 & <300) and only emit telemetry for failure scenarios. what do you think?

source/plugins/ruby/KubernetesApiClient.rb
jatakiajanvi12: Hey @ganga1980  that is because the cluster I was testing on had expired token so all the plugins were not running....Here is a working cluster

image

source/plugins/ruby/KubernetesApiClient.rb
jatakiajanvi12: Added the check before the call

source/plugins/ruby/ApplicationInsightsUtility.rb
jatakiajanvi12: Yep makes sense. I have added the check where the call happens

ganga1980 (approved):

PR #1128, opened 2023-11-29
Create validation pipeline for appmonitoring

ramanujbhattacharjee93:


.pipelines/azure_pipeline_validation_appmonitoring.yaml
```diff
@@ -0,0 +1,245 @@
...
+        echo "Wait for 10s for everything to clear up..."
+        sleep 10
+
+        export JAVA_TEST_IMAGE_NAME=${{ variables.javaTestAppImageName }}
+        export DOTNET_TEST_IMAGE_NAME=${{ variables.dotnetTestAppImageName }}
+        export NODEJS_TEST_IMAGE_NAME=${{ variables.nodeTestAppImageName }}
+        export TEST_NS="${{ variables.testNamespace }}"
+        export AI_CONNECTION_STRING="${{ variables.aiConnectionString }}"
+
+        kubectl create namespace $TEST_NS
```
tokaplan, line 217: Are we creating a new namespace for every test? Are we leaving all the old ones behind?

.pipelines/azure_pipeline_validation_appmonitoring.yaml
```diff
@@ -0,0 +1,245 @@
...
+
+        cat appmonitoring-cr.yaml | envsubst | kubectl apply -f -
```
tokaplan, line 219: Are we seeing kubectl output in the pipeline logs? Will we able to troubleshoot if something fails? Are we failing the entire pipeline here if something fails? We should.

appmonitoring/helm/appmonitoring-cr.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: azmon.app.monitoring/v1
```
tokaplan: Let's not call the folder simply "helm". That suggests that it's used in a helm chart of the addon. Let's call it something that clearly indicates it's only used in the validation pipeline, not actual production code.

appmonitoring/scripts/validate-aks-charts.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
tokaplan, line 1: Will a random failure in these bash files cause the pipeline to fail? Let's ensure that's the case.

appmonitoring/scripts/validate-aks-charts.sh
```diff
@@ -0,0 +1,32 @@
...
+
+az account set --subscription 66010356-d8a5-42d3-8593-6aaa3aeb1c11
```
tokaplan, line 3: As we are doing these validations, what is the guarantee we are looking at resources created by this particular run and not some left-overs from previous runs?

.pipelines/azure_pipeline_validation_appmonitoring.yaml
ramanujbhattacharjee93: no, its the same and we delete all that came before.

ramanujbhattacharjee93: no, its the same. We delete all resources from previous run

.pipelines/azure_pipeline_validation_appmonitoring.yaml
ramanujbhattacharjee93: these are fixed resources which should not ever fail. We should fail the entire thing if anything fails here as whatever follow is hence rendered moot

appmonitoring/scripts/validate-aks-charts.sh
ramanujbhattacharjee93: this is very deterministic code, this should not fail. if it does,it would indicate a code flaw

appmonitoring/scripts/validate-aks-charts.sh
ramanujbhattacharjee93: as long as pipelines are not running concurrently, it will run properly as it deletes all resources from the previous runs

tokaplan: We can't rely on that. We need to figure out how we handle this. Do we create separate namespaces? Do we want to detect that another pipeline is running and fail? We'll get some weird unreliable results if we just ignore that.

.pipelines/azure_pipeline_validation_appmonitoring.yaml
tokaplan: So is it failing the entire pipeline if it fails? And if it does, do we see output from kubectl to understand why it failed?

.pipelines/azure_pipeline_validation_appmonitoring.yaml
tokaplan: Won't work for concurrent pipelines. We need to either figure out how to make sure different runs don't affect each other, or how to prohibit and fail them with a meaningful message. Are all our resources namespace-scoped or not? Can we support multiple runs at once?

ramanujbhattacharjee93: we cannot do concurrent tests, as many of the things being created here can only exist one at a time in a cluster, like webhook, CRD, CR etc. I will restrict the pipeline to only running it one at a time

.pipelines/azure_pipeline_validation_appmonitoring.yaml
ramanujbhattacharjee93: yes you will fail the whole thing, and it will show the reason as well, just stdout

appmonitoring/scripts/validate-aks-charts.sh
ramanujbhattacharjee93: I believe we can just make the pipeline not run multiple jobs at once and wait if another is already running

.pipelines/azure_pipeline_validation_appmonitoring.yaml
tokaplan: If you can restrict the pipeline itself somehow - that would work. If not, then let's make sure we check in the scripts and bail out if the environment is used.

Another option would be to make those cluster-scoped objects per-run, like add the id of the run to the end of the object names... any way we could do that? It would be pretty awesome to support multiple runs in parallel.

ramanujbhattacharjee93: yes it can be made to only run 1 at a time by only having 1 agent available at any given time. the only way to run concurrent tests would be to create an entire aks cluster and set it up - run tests - then delete it -> do this on every run which would take a very long time and be very complicated to do

appmonitoring/helm/appmonitoring-cr.yaml
ramanujbhattacharjee93: done

tokaplan (approved):

PR #1139, opened 2023-12-18
Longw/metadata containerlogv2 kubernetes

wanlonghenry:
Longw/metadata containerlogv2 Kubernetes
back end code for containerlogv2  Kubernetes metadata


wanlonghenry:
tested scenarios with config map log_collection_settings.metadata_collection being 1. only enabled; 2. enabled + full include fields; 3 enabled + partial include fields
image

build/common/installer/scripts/tomlparser.rb
```diff
@@ -215,7 +215,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
         @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
         if !parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].nil?
           puts "config::Using config map setting for kubernetes metadata include fields"
-          @logKubernetesMetadataiIncludeFields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields]
+          @logKubernetesMetadataiIncludeFields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].join(",")
```
pfrcks: should we check whether the input is a valid array before doing a join?

source/plugins/go/src/oms.go
```diff
@@ -1870,6 +1963,16 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
...
 	ContainerLogV2ConfigMap = (strings.Compare(ContainerLogSchemaVersion, ContainerLogV2SchemaVersion) == 0)
 
+	KubernetesMetadataConfigMap = false
+	KubernetesMetadataConfigMap = (strings.Compare(strings.ToLower(os.Getenv("AZMON_KUBERNETES_METADATA_ENABLED")), "true") == 0)
+	metadataIncludeList := os.Getenv("AZMON_KUBERNETES_METADATA_INCLUDES_FIELDS")
+	Log(fmt.Sprintf("KubernetesMetadataIncludeList from configmap: %+v\n", metadataIncludeList))
+	if KubernetesMetadataConfigMap && len(metadataIncludeList) > 0 {
+		KubernetesMetadataIncludeList = strings.Split(metadataIncludeList, ",")
+	} else {
+		KubernetesMetadataIncludeList = []string{"podLabels", "podAnnotations", "podUid", "image"}
```
pfrcks: we can set this as default value and override it if the if condition is true.

source/plugins/go/src/oms.go
```diff
@@ -1237,6 +1326,8 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 				LogMessage:      stringMap["LogMessage"],
 				LogSource:       stringMap["LogSource"],
 				AzureResourceId: stringMap["AzureResourceId"],
+				KubernetesMetadata: stringMap["KubernetesMetadata"],
```
pfrcks: are we adding this to ADX as well?

build/common/installer/scripts/tomlparser.rb

source/plugins/go/src/oms.go
wanlonghenry: changed and add condition to handle        [log_collection_settings.metadata_collection]
          enabled = true
          include_fields = [""] scenario

source/plugins/go/src/oms.go
wanlonghenry: removed

source/plugins/go/src/oms.go
```diff
@@ -1870,6 +1968,17 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
...
 	ContainerLogV2ConfigMap = (strings.Compare(ContainerLogSchemaVersion, ContainerLogV2SchemaVersion) == 0)
 
+	KubernetesMetadataConfigMap = false
+	KubernetesMetadataConfigMap = (strings.Compare(strings.ToLower(os.Getenv("AZMON_KUBERNETES_METADATA_ENABLED")), "true") == 0)
+	metadataIncludeList := os.Getenv("AZMON_KUBERNETES_METADATA_INCLUDES_FIELDS")
+	Log(fmt.Sprintf("KubernetesMetadataIncludeList from configmap: %+v\n", metadataIncludeList))
+	KubernetesMetadataIncludeList = []string{"podLabels", "podAnnotations", "podUid", "image"}
+	if KubernetesMetadataConfigMap && len(metadataIncludeList) > 0 {
+		KubernetesMetadataIncludeList = strings.Split(metadataIncludeList, ",")
+	} else if KubernetesMetadataConfigMap {
```
pfrcks: if a customer does not provide include_list that is let's say they enabled it via
```

       [log_collection_settings.metadata_collection]
          # if enabled will collect kubernetes metadata for ContainerLogv2 schema. Default false
          enabled = true
          # include_fields = ["podLabels","podAnnotations","podUid","image"]
```

what will the value of metadataIncludeList be in this case?

source/plugins/go/src/oms.go
```diff
@@ -161,6 +161,10 @@ var (
 	ContainerLogSchemaV2 bool
 	// container log schema version from config map
 	ContainerLogV2ConfigMap bool
+	// Kubernetes Metadata enabled through configmap flag
+	KubernetesMetadataConfigMap bool
```
pfrcks: I think this should be renamed. Something like KubernetesMetadataEnabled?

source/plugins/go/src/oms.go
wanlonghenry: the default value will be "podLabels","podAnnotations","podUid","image"

source/plugins/go/src/oms.go
wanlonghenry: The reason why I put this name is because this feature might be expanded to enable through ux in future, let me know your think

pfrcks:
can you rebase?

source/plugins/go/src/oms.go
pfrcks: not sure I understand. can you expand?

wanlonghenry: Like in future if we added the kuberneteMetadata enabled through portal/dcr, we will have two flags deciding whether  kuberneteMetadata  enabled is true or not

pfrcks: i see. i think we should be fine with using KubernetesMetadataEnabled and can tackle UX case in the future if it happens considering we are planning on converging to one of the two options

source/plugins/go/src/oms.go
```diff
@@ -272,6 +276,8 @@ type DataItemLAv2 struct {
 	PodNamespace  string `json:"PodNamespace"`
 	LogMessage    string `json:"LogMessage"`
 	LogSource     string `json:"LogSource"`
+	KubernetesMetadata    string `json:"KubernetesMetadata"`
```
pfrcks: can you space align this struct?

source/plugins/go/src/oms.go
```diff
@@ -1251,6 +1347,8 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 					PodNamespace:  stringMap["PodNamespace"],
 					LogMessage:    stringMap["LogMessage"],
 					LogSource:     stringMap["LogSource"],
+					KubernetesMetadata: stringMap["KubernetesMetadata"],
```
pfrcks, line 1350: can you space align this as well?

source/plugins/go/src/oms.go
wanlonghenry: addressed

source/plugins/go/src/oms.go
wanlonghenry: addressed

pfrcks (approved):

PR #1140, opened 2023-12-18
make log filtering opt in explicit.

pfrcks:
- adds configmap option filter_using_annotations to turn on annotation based log filtering
- enables kubernetes fluent-bit filter if needed

### DIFF
--- a/build/common/installer/scripts/fluent-bit-conf-customizer.rb
+++ b/build/common/installer/scripts/fluent-bit-conf-customizer.rb
@@ -44,6 +44,7 @@ def substituteFluentBitPlaceHolders
     ignoreOlder = ENV["FBIT_TAIL_IGNORE_OLDER"]
     multilineLogging = ENV["AZMON_MULTILINE_ENABLED"]
     kubernetesMetadataCollection = ENV["AZMON_KUBERNETES_METADATA_ENABLED"]
+    annotationBasedLogFiltering = ENV["AZMON_ANNOTATION_BASED_LOG_FILTERING"]
 
     serviceInterval = (!interval.nil? && is_number?(interval) && interval.to_i > 0) ? interval : @default_service_interval
     serviceIntervalSetting = "Flush         " + serviceInterval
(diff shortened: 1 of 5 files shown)


build/linux/installer/conf/fluent-bit.conf
```diff
@@ -54,5 +54,5 @@
 #${KubernetesFilterEnabled}    Kubelet_Port        10250
 #${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
 #${KubernetesFilterEnabled}    tls.verify          Off
-#${KubernetesFilterEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
```
wanlonghenry, line 57: is there any space needed for the two commented out section? fluent-bit might raise some error if the dent is not right

build/common/installer/scripts/tomlparser.rb
```diff
@@ -219,6 +220,14 @@ def populateSettingValuesFromConfigMap(parsedConfig)
         end
       end
     end
+
+    #Get annotation based log filtering setting
+    begin
+      if !parsedConfig[:log_collection_settings][:filter_using_annotations].nil? && !parsedConfig[:log_collection_settings][:filter_using_annotations][:enabled].nil?
```
wanlonghenry, line 226: do we want to set this if @logEnableKubernetesMetadata is false?

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -84,6 +85,12 @@ def substituteFluentBitPlaceHolders
       new_contents = new_contents.gsub("#${KubernetesFilterEnabled}", "")
     end
 
+    if !annotationBasedLogFiltering.nil? && annotationBasedLogFiltering.to_s.downcase == "true"
```
wanlonghenry, line 88: do we need to add condition check for KubernetesFilterEnabled? majorly how do we tie the log filterring feature to KubernetesFilterEnabled

pfrcks: no. here's why. if KubernetesFilterEnabled is enabled then it will replace all the comments from fluent-bit.conf and therefore line 90 will be a noop

if KubernetesFilterEnabled  is false then line 90 will remove the comments but since KubernetesFilterEnabled  is disabled oms.go will not process any metadata

this way filtering will work without doing any additional metadata related tasks

build/common/installer/scripts/tomlparser.rb
pfrcks: refer to the reply of the previous comment

wanlonghenry (approved):

PR #1141, opened 2023-12-19
Disable component based on logs profile

pfrcks:
- Does not start fluentd and telegraf in DS if Logs and Events only profile set
- Updates livenessprobe to check for DCR updates and restart container if the logs profile changes

PR #1145, opened 2024-01-04
Disable component based on logs profile (#1141)

pfrcks:
- Does not start fluentd and telegraf in DS if Logs and Events only profile set
- Updates livenessprobe to check for DCR updates and restart container if the logs profile changes


build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -0,0 +1,52 @@
...
+]
+@logs_and_events_only = false
+
+return if !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp('windows').zero?
+return unless ENV['USING_AAD_MSI_AUTH'].strip.casecmp('true').zero?
+
+if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp('daemonset').zero? \
+  && @containerType.nil?
+  begin
+    file_path = Dir.glob(@dcrConfigFilePattern).first
```
ganga1980: In scenarios like with syslog or private link or customer has arbitery DCRs, first config json file may not  be container insights extension config and this could be possible syslog config (with syslog) and in case of AMPLS scenario, this could be DCE config etc.. we would need enumerate all the files which matching the criterion and get the path for container insights extension config JSON file.

build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    raise 'Invalid JSON structure: Missing required keys' unless data.is_a?(Hash) && data.key?('dataSources')
+
+    # Extract the stream values
+    streams = data['dataSources'].select { |ds| ds['id'] == 'ContainerInsightsExtension' }
+                                 .flat_map { |ds| ds['streams'] if ds.key?('streams') }
+                                 .compact
+                                 .map { |stream| stream['stream'] if stream.key?('stream') }
+                                 .compact
+    streams -= @logs_and_events_streams
```
ganga1980: I think, we should also handle following possible scenarios
1.   
CONTAINER_LOG_BLOB
  CONTAINERINSIGHTS_CONTAINERLOGV2
  KUBE_EVENTS_BLOB
2.
CONTAINER_LOG_BLOB
  CONTAINERINSIGHTS_CONTAINERLOGV2
3.

  CONTAINERINSIGHTS_CONTAINERLOGV2

build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -0,0 +1,52 @@
...
+    if streams.empty?
+      # Write the settings to file, so that they can be set as environment variables
+      puts 'DCR config matches Log and Events only profile. Setting LOGS_AND_EVENTS_ONLY to true'
+      @logs_and_events_only = true
+      file = File.open('dcr_env_var', 'w')
+      file.write("LOGS_AND_EVENTS_ONLY=#{@logs_and_events_only}\n")
+      file.close
+    end
+  rescue StandardError => e
```
ganga1980: To be safe, I think, we should use generic "Exception" instead of "StandardError" to cover all the possible exceptions. This way, any of exception, and doesnt get crashed.

build/linux/installer/scripts/livenessprobe.sh
```diff
@@ -14,6 +14,14 @@ if [[ "${CONTROLLER_TYPE}" == "DaemonSet" && "${CONTAINER_TYPE}" != "PrometheusS
                   [ -f "/var/run/mdsd-ci/70-rsyslog-forward-mdsd-ci.conf" ] && rm /var/run/mdsd-ci/70-rsyslog-forward-mdsd-ci.conf && echo "remove" > /var/run/mdsd-ci/update.status
           fi
   fi
+
+  CURRENT_LOGS_AND_EVENTS_ONLY=${LOGS_AND_EVENTS_ONLY}
```
ganga1980, line 18: This needs to be executed only when controller type is Daemonset & main container, right?

pfrcks: yes, it is inside the if block which checks that

ganga1980 (dismissed):

build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -0,0 +1,62 @@
+#!/usr/local/bin/ruby
+
+require 'fileutils'
```
ganga1980, line 3: Add telemetry to track on which clusters this feature activated.

ganga1980 (approved):

PR #1138, opened 2023-12-12
collect logs from pods from excluded system ns

pfrcks:
- enabled log collection of specific pods from excluded system namespaces
- use daemonset and deployment names to identify pods whose logs need to be collected.
- extracts potential ds and deployment names from podName and matches against user provided input in configmap option include_system_pods

eg. user sets include_system_pods = ["coredns"] where coredns can be deployment name or daemonset name which will result in pod name coredns-77sd7 or coredns-65cbf4f7bc-jcn6z.

PR #1146, opened 2024-01-05
collect logs from pods from excluded system ns (#1138)

pfrcks:
- enabled log collection of specific pods from system namespaces.
- use daemonset and deployment names to identify pods whose logs need to be collected.
- extracts potential ds and deployment names from podName and matches against user provided input in configmap option include_system_pods

eg. user sets include_system_pods = ["coredns"] where coredns can be deployment name or daemonset name which will result in pod name coredns-77sd7 or coredns-65cbf4f7bc-jcn6z.

### DIFF
--- a/build/common/installer/scripts/tomlparser.rb
+++ b/build/common/installer/scripts/tomlparser.rb
@@ -98,10 +98,16 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           # Using is_a? for type checking and directly checking if the array is not empty
           filtered_entries = stdoutSystemPods.each_with_object([]) do |pod, entries|
             namespace, _controller = pod.split(':') # Split once and use the result
-            if namespace && @allowed_system_namespaces.include?(namespace) && @stdoutExcludeNamespaces.include?(namespace)
+            if namespace && @allowed_system_namespaces.include?(namespace) && !@stdoutExcludeNamespaces.include?(namespace)
               entries << pod
             else
               puts "config:: invalid entry for collect_system_pod_logs: #{pod}"
+              unless @allowed_system_namespaces.include?(namespace)
+                puts "config:: collect_system_pod_logs only works for system namespaces #{@allowed_system_namespaces}"
+              end
+              if @stdoutExcludeNamespaces.include?(namespace)
+                puts "config:: please remove #{namespace} from exclude_namespaces to use collect_system_pod_logs"
+              end
             end
           end
 
@@ -159,10 +165,16 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           # Using is_a? for type checking and directly checking if the array is not empty
           filtered_entries = stderrSystemPods.each_with_object([]) do |pod, entries|
             namespace, _controller = pod.split(':') # Split once and use the result
-            if namespace && @allowed_system_namespaces.include?(namespace) && @stderrExcludeNamespaces.include?(namespace)
+            if namespace && @allowed_system_namespaces.include?(namespace) && !@stderrExcludeNamespaces.include?(namespace)
               entries << pod
             else
               puts "config:: invalid entry for collect_system_pod_logs: #{pod}"
+              unless @allowed_system_namespaces.include?(namespace)
+                puts "config:: collect_system_pod_logs only works for system namespaces #{@allowed_system_namespaces}"
+              end
+              if @stdoutExcludeNamespaces.include?(namespace)
+                puts "config:: please remove #{namespace} from exclude_namespaces to use collect_system_pod_logs"
+              end
             end
           end
 
--- a/build/common/installer/scripts/tomlparser.rb
+++ b/build/common/installer/scripts/tomlparser.rb
@@ -30,6 +30,9 @@ require_relative "ConfigParseErrorLogger"
 @logEnableKubernetesMetadata = false
 @logKubernetesMetadataIncludeFields = "podlabels,podannotations,poduid,image,imageid,imagerepo,imagetag"
 @annotationBasedLogFiltering = false
+@allowed_system_namespaces = ['kube-system', 'gatekeeper-system', 'calico-system', 'azure-arc', 'kube-public', 'kube-node-lease']
+
+
 if !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp("windows") == 0
   @containerLogsRoute = "v1" # default is v1 for windows until windows agent integrates windows ama
   # This path format is necessary for fluent-bit in windows
@@ -91,12 +94,21 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           end
         end
 
-        if @collectStdoutLogs && !stdoutSystemPods.nil? && stdoutSystemPods.kind_of?(Array)
-          # Checking only for the first element to be string because toml enforces the arrays to contain elements of same type
-          if stdoutSystemPods.length > 0 && stdoutSystemPods[0].kind_of?(String)
-            @stdoutIncludeSystemPods = stdoutSystemPods.join(",")
-            puts "config::Using config map setting for stdout log collection to include system pods"
+        if @collectStdoutLogs && stdoutSystemPods.is_a?(Array) && !stdoutSystemPods.empty?
+          # Using is_a? for type checking and directly checking if the array is not empty
+          filtered_entries = stdoutSystemPods.each_with_object([]) do |pod, entries|
+            namespace, _controller = pod.split(':') # Split once and use the result
+            if namespace && @allowed_system_namespaces.include?(namespace) && @stdoutExcludeNamespaces.include?(namespace)
+              entries << pod
+            else
+              puts "config:: invalid entry for collect_system_pod_logs: #{pod}"
+            end
           end
+
+          @stdoutIncludeSystemPods = filtered_entries.join(",")
+          puts "config::Using config map setting for stdout log collection to include system pods" if filtered_entries.any?
+        else
+          puts "config::Stdout log collection is not enabled or stdoutSystemPods is not properly configured." unless @collectStdoutLogs
         end
 
       end
@@ -143,12 +155,21 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           end
         end
 
-        if @collectStderrLogs && !stderrSystemPods.nil? && stderrSystemPods.kind_of?(Array)
-          # Checking only for the first element to be string because toml enforces the arrays to contain elements of same type
-          if stderrSystemPods.length > 0 && stderrSystemPods[0].kind_of?(String)
-            @stderrIncludeSystemPods = stderrSystemPods.join(",")
-            puts "config::Using config map setting for stderr log collection to include system pods"
+        if @collectStderrLogs && stderrSystemPods.is_a?(Array) && !stderrSystemPods.empty?
+          # Using is_a? for type checking and directly checking if the array is not empty
+          filtered_entries = stderrSystemPods.each_with_object([]) do |pod, entries|
+            namespace, _controller = pod.split(':') # Split once and use the result
+            if namespace && @allowed_system_namespaces.include?(namespace) && @stderrExcludeNamespaces.include?(namespace)
+              entries << pod
+            else
+              puts "config:: invalid entry for collect_system_pod_logs: #{pod}"
+            end
           end
+
+          @stderrIncludeSystemPods = filtered_entries.join(",")
+          puts "config::Using config map setting for stderr log collection to include system pods" if filtered_entries.any?
+        else
+          puts "config::stderr log collection is not enabled or stderrSystemPods is not properly configured." unless @collectStderrLogs
         end
 
       end
--- a/build/common/installer/scripts/tomlparser.rb
+++ b/build/common/installer/scripts/tomlparser.rb
@@ -98,7 +98,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           # Using is_a? for type checking and directly checking if the array is not empty
           filtered_entries = stdoutSystemPods.each_with_object([]) do |pod, entries|
             namespace, controller = pod.split(':') # Split once and use the result
-            if namespace && @allowed_system_namespaces.include?(namespace) && !@stdoutExcludeNamespaces.include?(namespace) && !controller&.empty?
+            if namespace && @allowed_system_namespaces.include?(namespace) && !@stdoutExcludeNamespaces.include?(namespace) && controller && !controller.empty?
               entries << pod
             else
               puts "config:: invalid entry for collect_system_pod_logs: #{pod}"
@@ -108,7 +108,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
               if @stdoutExcludeNamespaces.include?(namespace)
                 puts "config:: please remove #{namespace} from exclude_namespaces to use collect_system_pod_logs"
               end
-              if controller&.empty?
+              if !controller || controler.empty?
                 puts "config:: Please provide valid controller name. controller name is empty"
               end
             end
@@ -168,7 +168,7 @@ def populateSettingValuesFromConfigMap(parsedConfig)
           # Using is_a? for type checking and directly checking if the array is not empty
           filtered_entries = stderrSystemPods.each_with_object([]) do |pod, entries|
             namespace, controller = pod.split(':') # Split once and use the result
-            if namespace && @allowed_system_namespaces.include?(namespace) && !@stderrExcludeNamespaces.include?(namespace) && !controller&.empty?
+            if namespace && @allowed_system_namespaces.include?(namespace) && !@stderrExcludeNamespaces.include?(namespace) && controller && !controller.empty?
               entries << pod
             else
               puts "config:: invalid entry for collect_system_pod_logs: #{pod}"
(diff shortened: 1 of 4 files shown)


kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -20,6 +20,8 @@ data:
           # If you want to continue to disable kube-system,gatekeeper-system log collection keep the namespaces in the following setting and add any other namespace you want to disable log collection to the array.
           # In the absense of this configmap, default value for exclude_namespaces = ["kube-system","gatekeeper-system"]
           exclude_namespaces = ["kube-system","gatekeeper-system"]
+          # If you want to collect logs from system pods that have been excluded using exclude_namespaces setting, add them to the following setting. Provide daemonset or deployment name of the system pod. Eg adding coredns to list can capture logs from pods with name like coredns-77sd7 or coredns-65cbf4f7bc-jcn6z. NOTE: this setting is only for system pods
+          # include_system_pods = ["coredns"]
```
ganga1980: reading the comment,  "include_system_pods" works on excluded system namespaces, is that right? if its case, what's behavior when the system namespace included and also pod names in "include_system_pods"?  I was thinking the system_pods filter works on included system namespaces and collects only the logs for the specified system pods filter if there is filter. lets discuss more on the behavior we should enable it.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -116,6 +138,20 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+          # Checking only for the first element to be string because toml enforces the arrays to contain elements of same type
+          if stderrSystemPods.length > 0 && stderrSystemPods[0].kind_of?(String)
+            @stderrIncludeSystemPods = stderrSystemPods.join(",")
+            puts "config::Using config map setting for stderr log collection to include system pods"
+          end
+        end
+
+        # If we have to collect logs from system pods belonging to exlcuded namespaces from either of the streams we need to reset exclude path to noop
+        if !@stdoutIncludeSystemPods.empty? || !@stderrIncludeSystemPods.empty?
+          @excludePath = "*.csv2"
```
ganga1980: are we resetting excludePath if customer excluded non system namespaces (in both stderr & stdout) and using the include_system_pods? if yes, what's the behavior if the non-system excluded namespaces? are we using the pod filter in the fluent-bit Exclude_Path?   Did we test all the possible scenarios for any regressions?

source/plugins/go/src/oms.go
```diff
@@ -1148,12 +1180,18 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 		logEntrySource := ToString(record["stream"])
 
 		if strings.EqualFold(logEntrySource, "stdout") {
-			if containerID == "" || containsKey(StdoutIgnoreNsSet, k8sNamespace) {
-				continue
+			if containerID == "" { continue }
+			if containsKey(StdoutIgnoreNsSet, k8sNamespace) {
+				if len(StdoutIncludeSystemPodsSet) == 0 { continue }
+				dsName, deploymentName := GetDSNameAndDeploymentNameFromK8sPodName(k8sPodName)
```
ganga1980: This will be costly doing for every log record and  also, we will not be covering all other controller types for example if  jobs, cronjobs and statefulset etc. instead we can change from include_system_pods to include_system_namespace_containers and use the container name specified in the configmap against the container name in the log entry.

kubernetes/container-azm-ms-agentconfig.yaml
ganga1980: is this feature works for both ContainerLog & ContainerLogV2 or ContainerLogV2? does it work for LA and Geneva mode?

source/plugins/go/src/oms.go
```diff
@@ -206,8 +206,14 @@ var (
 	NameIDMap map[string]string
 	// StdoutIgnoreNamespaceSet set of  excluded K8S namespaces for stdout logs
 	StdoutIgnoreNsSet map[string]bool
+	// StdoutIncludeSystemPodsSet set of included system pods for stdout logs
+	StdoutIncludeSystemPodsSet map[string]bool
```
ganga1980: just we need to list of system namespace containers for stdout & stderr, right?

pfrcks: Used hashmap for efficient search

source/plugins/go/src/oms.go
pfrcks: I thought about this and the reason I went with this route is because different pods across namespaces can have the same container name. In that case we might end up collecting logs from system namespaces we do not want. what do you think?

build/common/installer/scripts/tomlparser.rb
pfrcks: at present if namespace was present in both stderr and stdout we added it to exclude path. we can no longer do this if we want to collect logs from pods in such namespaces. therefore I reset excludePath. 

 for non-system namespaces in this scenario the records will get dropped in oms.go. unless we input both system namespace name and pod name in configmap it will be hard to avoid this

kubernetes/container-azm-ms-agentconfig.yaml
pfrcks: this works for all supported logs since processing happens before type decision

pfrcks: if a namespace is already included then this is a noop. include_system_pods only works if the system pod is in the exlude namespace list.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -59,6 +61,11 @@ def populateSettingValuesFromConfigMap(parsedConfig)
         @collectStdoutLogs = parsedConfig[:log_collection_settings][:stdout][:enabled]
         puts "config::Using config map setting for stdout log collection"
         stdoutNamespaces = parsedConfig[:log_collection_settings][:stdout][:exclude_namespaces]
+        
+        stdoutSystemPods = Array.new
+        if !parsedConfig[:log_collection_settings][:stdout][:collect_system_pod_logs].nil?
+          stdoutSystemPods = parsedConfig[:log_collection_settings][:stdout][:collect_system_pod_logs]
```
ganga1980, line 74: what would be the behavior if the customer specify the collect_system_pods with not expected format  ? For example, customer specify  ["coredns","metricsserver"] . we would need to validate and log the warn and ignore the invalid entries.

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -20,6 +20,8 @@ data:
           # If you want to continue to disable kube-system,gatekeeper-system log collection keep the namespaces in the following setting and add any other namespace you want to disable log collection to the array.
           # In the absense of this configmap, default value for exclude_namespaces = ["kube-system","gatekeeper-system"]
           exclude_namespaces = ["kube-system","gatekeeper-system"]
+          # If you want to collect logs from only selective pods inside system namespaces add them to the following setting. Provide namepace:controllerName of the system pod. NOTE: this setting is only for system pods
+          # collect_system_pod_logs = ["kube-system:coredns"]
```
ganga1980, line 25: For this to work, system namespaces corresponding to the pods in collect_system_pod_logs  to be the exclude_namespaces list or not?  Instructions in the configmap not clear to me.

source/plugins/go/src/oms.go
```diff
@@ -502,6 +520,46 @@ func populateExcludedStderrNamespaces() {
...
+func populateIncludedStdoutSystemResource() {
+	collectStdoutLogs := os.Getenv("AZMON_COLLECT_STDOUT_LOGS")
+	var stdoutIncludedSystemResourceList []string
+	includeList := os.Getenv("AZMON_STDOUT_INCLUDED_SYSTEM_PODS")
+	if (collectStdoutLogs == "true") && (len(includeList) > 0) {
+		stdoutIncludedSystemResourceList = strings.Split(includeList, ",")
+		for _, resource := range stdoutIncludedSystemResourceList {
+			colonLoc := strings.Index(resource, ":")
+			if colonLoc == -1 {
+				Log("Skipping invalid system resource namespace:controller %s for stdout log collection", resource)
```
ganga1980: This validation should happen as part of the configmap parsing so that error/warning visible for customer as part of the ama-logs pod logs.

source/plugins/go/src/oms.go
```diff
@@ -502,6 +520,46 @@ func populateExcludedStderrNamespaces() {
...
+			}
+			namespace := resource[:colonLoc]
+			Log("Including system resource namespace:controller %s for stdout log collection", resource)
+			StdoutIncludeSystemResourceSet[strings.TrimSpace(resource)] = true
+			StdoutIncludeSystemNamespaceSet[strings.TrimSpace(namespace)] = true
+		}
+	}
+}
+
+func populateIncludedStderrSystemResource() {
```
ganga1980, line 557: since the logic common for both stderr and stdout. I think, we should merge both these functions populateIncludedStdoutSystemResource & populateIncludedStderrSystemResource.

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -20,6 +20,8 @@ data:
           # If you want to continue to disable kube-system,gatekeeper-system log collection keep the namespaces in the following setting and add any other namespace you want to disable log collection to the array.
           # In the absense of this configmap, default value for exclude_namespaces = ["kube-system","gatekeeper-system"]
           exclude_namespaces = ["kube-system","gatekeeper-system"]
+          # If you want to collect logs from only selective pods inside system namespaces add them to the following setting. Provide namepace:controllerName of the system pod. NOTE: this setting is only for system pods
```
ganga1980: we should add the namespaces which are considered as system namespace to the instructions.

source/plugins/go/src/oms.go
```diff
@@ -502,6 +520,46 @@ func populateExcludedStderrNamespaces() {
 	}
 }
 
+func populateIncludedStdoutSystemResource() {
```
ganga1980: we should refactor the code in populateIncludedStdoutSystemResource & populateIncludedStderrSystemResource to single common function since the logic same in both the functions.

source/plugins/go/src/oms.go
```diff
@@ -1308,6 +1382,9 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 
 	numContainerLogRecords := 0
 
+	Log("StdoutIncludeSystemResourceSet : %v, StderrIncludeSystemResourceSet: %v, StdoutIncludeSystemNamespaceSet: %v, StderrIncludeSystemNamespaceSet: %v, PodNameToDSDeploymentNameMap: %v",
```
ganga1980: We should add telemetry to track what system namespaces and pods customer using this feature.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -79,6 +86,15 @@ def populateSettingValuesFromConfigMap(parsedConfig)
             end
           end
         end
+
+        if @collectStdoutLogs && !stdoutSystemPods.nil? && stdoutSystemPods.kind_of?(Array)
+          # Checking only for the first element to be string because toml enforces the arrays to contain elements of same type
+          if stdoutSystemPods.length > 0 && stdoutSystemPods[0].kind_of?(String)
+            @stdoutIncludeSystemPods = stdoutSystemPods.join(",")
```
ganga1980: If customer uses the namespaces other than system namespaces list, we should log warn message and ignore the entry.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -116,6 +138,15 @@ def populateSettingValuesFromConfigMap(parsedConfig)
             end
           end
         end
+
+        if @collectStderrLogs && !stderrSystemPods.nil? && stderrSystemPods.kind_of?(Array)
+          # Checking only for the first element to be string because toml enforces the arrays to contain elements of same type
+          if stderrSystemPods.length > 0 && stderrSystemPods[0].kind_of?(String)
+            @stderrIncludeSystemPods = stderrSystemPods.join(",")
+            puts "config::Using config map setting for stderr log collection to include system pods"
```
ganga1980: same here .. If customer uses the namespaces other than system namespaces list, we should log warn message and ignore the entry.

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -29,6 +31,8 @@ data:
...
```
ganga1980: we should specify what namespaces are considered as system namespaces.

source/plugins/go/src/oms.go
```diff
@@ -1151,10 +1209,26 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 			if containerID == "" || containsKey(StdoutIgnoreNsSet, k8sNamespace) {
 				continue
 			}
+			if containsKey(SystemNamespaces, k8sNamespace) && containsKey(StdoutIncludeSystemNamespaceSet, k8sNamespace) {
```
ganga1980: This check "containsKey(SystemNamespaces, k8sNamespace)"  will be redundant if system namespace allow list check in the tomlparser

source/plugins/go/src/oms.go
```diff
@@ -118,6 +118,14 @@ const ContainerTypeEnv = "CONTAINER_TYPE"
 // Default ADX destination database name, can be overriden through configuration
 const DefaultAdxDatabaseName = "containerinsights"
 
+var SystemNamespaces = map[string]bool{
+	"kube-system":       true,
+	"azure-arc":         true,
+	"gatekeeper-system": true,
+	"kube-public":       true,
+	"kube-node-lease":   true,
```
ganga1980: calico-system is other system namespace and probable, we should consider this .

source/plugins/go/src/oms.go
```diff
@@ -1622,6 +1699,25 @@ func GetContainerIDK8sNamespacePodNameFromFileName(filename string) (string, str
 	return id, ns, podName, containerName
 }
 
+// GetDSNameAndDeploymentNameFromK8sPodName tries to extract potential deployment name and daemonset name from the pod name. If its not possible to extract then sends empty string
+func GetDSNameAndDeploymentNameFromK8sPodName(podName string) (string, string) {
```
ganga1980: This should be GetControllerNameFromK8sPodName and also this should cover other controller types jobs, cronjobs and statefulset etc.

source/plugins/go/src/oms.go
```diff
@@ -1151,10 +1209,26 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+				}
+			}
 		} else if strings.EqualFold(logEntrySource, "stderr") {
 			if containerID == "" || containsKey(StderrIgnoreNsSet, k8sNamespace) {
 				continue
 			}
+			if containsKey(SystemNamespaces, k8sNamespace) && containsKey(StderrIncludeSystemNamespaceSet, k8sNamespace) {
+				if len(StderrIncludeSystemResourceSet) != 0 {
+					dsName, deploymentName := GetDSNameAndDeploymentNameFromK8sPodName(k8sPodName)
+					if !containsKey(StderrIncludeSystemResourceSet, k8sNamespace+":"+dsName) && !containsKey(StderrIncludeSystemResourceSet, k8sNamespace+":"+deploymentName) {
```
ganga1980: We dont need to check both ds and deployment name. checking for controller name should consider all the possible controller types.

kubernetes/container-azm-ms-agentconfig.yaml
ganga1980: Please add the instructions about namespaces which are considered as system namespaces.

kubernetes/container-azm-ms-agentconfig.yaml
pfrcks: include_system_pods only works if the system pod is in the exlude namespace list.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -83,6 +93,30 @@ def populateSettingValuesFromConfigMap(parsedConfig)
             end
           end
         end
+
+        if @collectStdoutLogs && stdoutSystemPods.is_a?(Array) && !stdoutSystemPods.empty?
+          # Using is_a? for type checking and directly checking if the array is not empty
+          filtered_entries = stdoutSystemPods.each_with_object([]) do |pod, entries|
+            namespace, _controller = pod.split(':') # Split once and use the result
```
ganga1980: are we treating "<namespace>:"  is a valid entry?

ganga1980 (dismissed):

source/plugins/go/src/oms.go
```diff
@@ -1945,6 +2004,25 @@ func GetContainerIDK8sNamespacePodNameFromFileName(filename string) (string, str
 	return id, ns, podName, containerName
 }
 
+// GetControllerNameFromK8sPodName tries to extract potential deployment name and daemonset name from the pod name. If its not possible to extract then sends empty string
+func GetControllerNameFromK8sPodName (podName string) (string, string) {
+	if values, ok := PodNameToDSDeploymentNameMap[podName]; ok {
+		return values[0], values[1]
```
ganga1980, line 2018: this needs to be renamed to "PodNameToControllerNameMap"

ganga1980 (approved):

PR #1147, opened 2024-01-08
Telemetry optimization: Sending mdsd error traces as metrics

jatakiajanvi12:
Converted mdsd error traces to be sent as customMetrics every telemetryPushInterval instead of sending all the statements as traces

### DIFF
--- a/source/plugins/go/src/telemetry.go
+++ b/source/plugins/go/src/telemetry.go
@@ -539,12 +539,16 @@ func PushToAppInsightsTraces(records []map[interface{}]interface{}, severityLeve
 			populateKubeMonAgentEventHash(record, PromScrapingError)
 		} else if strings.Contains(logEntry, "Lifetime validation failed. The token is expired.") {
 			UpdateMdsdErrorMetrics("MdsdTokenExpired")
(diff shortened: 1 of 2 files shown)


source/plugins/go/src/telemetry.go
```diff
@@ -500,6 +537,14 @@ func PushToAppInsightsTraces(records []map[interface{}]interface{}, severityLeve
 			populateKubeMonAgentEventHash(record, ConfigError)
 		} else if strings.Contains(logEntry, "E! [inputs.prometheus]") {
 			populateKubeMonAgentEventHash(record, PromScrapingError)
+		} else if strings.Contains(logEntry, "Lifetime validation failed. The token is expired.") {
```
ganga1980, line 540: Are we covering all the possible error scenarios of the MDSD here? Can you please look at CI agent log dashboards and telemetry to make sure we have metric for all key error scenarios for example In case of mode, Agent registration/onboarding failure and MSI mode - no DCR or invalid DCR etc.

jatakiajanvi12: Added a couple more but these were the only error types I found...Once we convert these to metrics will again review and keep on adding more scenarios if found. Let me know if you think otherwise

ganga1980 (approved):

PR #1148, opened 2024-01-08
Containerlogv2 Metadata and annotations based filtering

wanlonghenry:
Containerlogv2 Metadata and annotations based filtering.

image
image

For scale test result, please reach out to me for the doc.


pfrcks:
Filtering perf results. Enabling filtering with annotations has lower perf consumtion than filterting disabled.

cpu1
mem1

pfrcks:
@ganga1980 @wanlonghenry current I am using Kube_Meta_Cache_TTL as 60s only for filtering scenario. From fluent-bit docs

`configurable TTL for K8s cached metadata. By default, it is set to 0 which means TTL for cache entries is disabled and cache entries are evicted at random when capacity is reached. In order to enable this option, you should set the number to a time interval. For example, set this value to 60 or 60s and cache entries which have been created more than 60s will be evicted.`

Setting a value is needed for filtering scenario because otherwise we dont know when metadata will be updated for a pod once the customer adds the annotation. This way the fluent-bit filter starts dropping the logs within 60 seconds of adding the annotation. Let me know if 60s is too aggressive and needs to be increased to let's say 5minutes?

Also should this setting be applicable for metadata collection as well? I imagine that if not set then updates to labels/annotations done by the customer will take some time in getting reflected in the LA table column. @wanlonghenry is this something you have tested?

build/common/installer/scripts/tomlparser.rb
```diff
@@ -25,6 +25,9 @@
 @adxDatabaseName = "containerinsights" # default for all configurations
 @logEnableMultiline = "false"
 @stacktraceLanguages = "go,java,python,dotnet"
+@logEnableKubernetesMetadata = false
+@logKubernetesMetadataiIncludeFields = "podLabels,podAnnotations,podUid,image"
```
ganga1980: typo. There extra **i**. should be "logKubernetesMetadataIncludeFields"

ganga1980: we should also add support for "userPodLabels" and "systemPodLabels", "userPodAnnotations" and "systemPodAnnotations". Customers might prefer to exclude "systemPodLabels" for optimizing their cost.

wanlonghenry: changed

wanlonghenry: will go after private preview

build/common/installer/scripts/tomlparser.rb
```diff
@@ -25,6 +25,9 @@
 @adxDatabaseName = "containerinsights" # default for all configurations
 @logEnableMultiline = "false"
 @stacktraceLanguages = "go,java,python,dotnet"
+@logEnableKubernetesMetadata = false
+@logKubernetesMetadataIncludeFields = "podLabels,podAnnotations,podUid,image"
```
ganga1980: As we discussed during bug bash, we would need to address items which we have identified during bug bash as part of the public preview.

build/linux/installer/conf/fluent-bit.conf
```diff
@@ -44,3 +44,16 @@
...
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    Kube_Meta_Cache_TTL 60
```
ganga1980: I think, 60 is default good and lets make this configurable through config map  if this needs to be smaller value for the highly dynamic workload.

build/windows/installer/conf/fluent-bit.conf
```diff
@@ -43,3 +43,16 @@
...
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    Kube_Meta_Cache_TTL 60
```
ganga1980: I think, 60 is default good and lets make this configurable through config map if this needs to be smaller value for the highly dynamic workload.

build/linux/installer/conf/fluent-bit.conf
```diff
@@ -44,3 +44,16 @@
 #${MultilineEnabled}    Match oms.container.log.la.*
 #${MultilineEnabled}    multiline.key_content log
 #${MultilineEnabled}    multiline.parser #${MultilineLanguages}
+
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
```
ganga1980, line 56: can you please validate the metadata and annotation based filtering with multi-line enabled?

source/plugins/go/src/oms.go
```diff
@@ -1171,6 +1254,8 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 		}
 
 		logEntry := ToString(record["log"])
+		//filter loglevel and define here
+		//logLevel := ToString(record["LogLevel"])
```
ganga1980: are we plan to enable logLevel code when this PR ready to merge?

source/plugins/go/src/oms.go
```diff
@@ -1116,6 +1122,62 @@ func UpdateNumTelegrafMetricsSentTelemetry(numMetricsSent int, numSendErrors int
...
+			}
+			if image, ok := kubernetesMetadataMap["container_image"]; ok {
+				includedMetadata["image"] = image
+			}
+		}
+	}
+	return includedMetadata
+}
+
+func convertKubernetesMetadata(kubernetesMetadataJson interface{}) (map[string]interface{}, error) {
```
ganga1980, line 1262: why do we need this convertKubernetesMetadata function? cant we use json.Unmarshal function to convert the JSON string to struct?

source/plugins/go/src/oms.go
```diff
@@ -1146,6 +1208,27 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 	for _, record := range tailPluginRecords {
 		containerID, k8sNamespace, k8sPodName, containerName := GetContainerIDK8sNamespacePodNameFromFileName(ToString(record["filepath"]))
 		logEntrySource := ToString(record["stream"])
+		kubernetesMetadata := ""
+		if KubernetesMetadataEnabled {
+			if kubernetesMetadataJson, exists := record["kubernetes"]; exists {
```
ganga1980, line 1442: lets have metric to track processing time of the Kubernetes metadata.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -227,6 +230,29 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
     end
+
+    #Get Kubernetes Metadata setting
+    begin
+      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
+        puts "config::Using config map setting for kubernetes metadata"
+        @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
+        if !parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].nil?
+          puts "config::Using config map setting for kubernetes metadata include fields"
+          include_fields = parsedConfig[:log_collection_settings][:metadata_collection][:include_fields]
```
ganga1980, line 242: we should validate the provided field in the include_fields is supported list or not? If its not in supported list, log warn message

source/plugins/go/src/oms.go
```diff
@@ -1116,6 +1122,62 @@ func UpdateNumTelegrafMetricsSentTelemetry(numMetricsSent int, numSendErrors int
 	ContainerLogTelemetryMutex.Unlock()
 }
 
+func processIncludes(kubernetesMetadataMap map[string]interface{}, includesList []string) map[string]interface{} {
+	includedMetadata := make(map[string]interface{})
+	for _, include := range includesList {
+		switch include {
+		case "podUid":
```
ganga1980: we should convert the include fields in lower case and use the lower case for comparison so that  we can support even if customer having different casing for the field names in include_list

source/plugins/go/src/oms.go
```diff
@@ -266,14 +270,16 @@ type DataItemLAv1 struct {
...
+	TimeGenerated      string `json:"TimeGenerated"`
+	Computer           string `json:"Computer"`
+	ContainerId        string `json:"ContainerId"`
+	ContainerName      string `json:"ContainerName"`
+	PodName            string `json:"PodName"`
+	PodNamespace       string `json:"PodNamespace"`
+	LogMessage         string `json:"LogMessage"`
+	LogSource          string `json:"LogSource"`
+	KubernetesMetadata string `json:"KubernetesMetadata"`
+	//LogLevel     string `json:"LogLevel"`
```
ganga1980: either LogLevel code in this PR or remove related code if plan to address in different PR.

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -47,6 +47,8 @@ def substituteFluentBitPlaceHolders
     ignoreOlder = ENV["FBIT_TAIL_IGNORE_OLDER"]
```
ganga1980: Please add the agent resource usage charts with and without kubernetes metadata to the PR to validate and refer the resource usage consumption.

source/plugins/go/src/oms.go
```diff
@@ -1881,6 +1979,17 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
...
 
+	KubernetesMetadataEnabled = false
+	KubernetesMetadataEnabled = (strings.Compare(strings.ToLower(os.Getenv("AZMON_KUBERNETES_METADATA_ENABLED")), "true") == 0)
+	metadataIncludeList := os.Getenv("AZMON_KUBERNETES_METADATA_INCLUDES_FIELDS")
+	Log(fmt.Sprintf("KubernetesMetadataIncludeList from configmap: %+v\n", metadataIncludeList))
+	KubernetesMetadataIncludeList = []string{"podLabels", "podAnnotations", "podUid", "image"}
+	if KubernetesMetadataEnabled && len(metadataIncludeList) > 0 {
+		KubernetesMetadataIncludeList = strings.Split(metadataIncludeList, ",")
+	} else if KubernetesMetadataEnabled {
+		KubernetesMetadataIncludeList = []string{}
```
ganga1980: why do need this? what would be behavior when customer enabled metadata with empty include_fields?

source/plugins/go/src/oms.go
```diff
@@ -1881,6 +1979,17 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
 	ContainerLogSchemaV2 = false //default is v1 schema
 	ContainerLogV2ConfigMap = (strings.Compare(ContainerLogSchemaVersion, ContainerLogV2SchemaVersion) == 0)
 
+	KubernetesMetadataEnabled = false
+	KubernetesMetadataEnabled = (strings.Compare(strings.ToLower(os.Getenv("AZMON_KUBERNETES_METADATA_ENABLED")), "true") == 0)
+	metadataIncludeList := os.Getenv("AZMON_KUBERNETES_METADATA_INCLUDES_FIELDS")
+	Log(fmt.Sprintf("KubernetesMetadataIncludeList from configmap: %+v\n", metadataIncludeList))
+	KubernetesMetadataIncludeList = []string{"podLabels", "podAnnotations", "podUid", "image"}
```
ganga1980: why do we need this default? isnt that default should be []string{}

source/plugins/go/src/oms.go
```diff
@@ -1881,6 +1979,17 @@ func InitializePlugin(pluginConfPath string, agentVersion string) {
...
+	if KubernetesMetadataEnabled && len(metadataIncludeList) > 0 {
+		KubernetesMetadataIncludeList = strings.Split(metadataIncludeList, ",")
```
ganga1980, line 2214: where are we validating the customer provided input  include_fields is valid or not?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -53,7 +53,13 @@ data:
           # If enabled will also stitch together container logs split by docker/cri due to size limits(16KB per log line) up to 64 KB.
           # enabled = "false"
           # stacktrace_languages = []
-
+       #[log_collection_settings.metadata_collection]
+          # if enabled will collect kubernetes metadata for ContainerLogv2 schema. Default false
+          # enabled = false
+          # include_fields = ["podLabels","podAnnotations","podUid","image"]
```
ganga1980: as part of the private preview, we should validate with the customer if they need any other fields.

ganga1980:
> @ganga1980 @wanlonghenry current I am using Kube_Meta_Cache_TTL as 60s only for filtering scenario. From fluent-bit docs
> 
> `configurable TTL for K8s cached metadata. By default, it is set to 0 which means TTL for cache entries is disabled and cache entries are evicted at random when capacity is reached. In order to enable this option, you should set the number to a time interval. For example, set this value to 60 or 60s and cache entries which have been created more than 60s will be evicted.`
> 
> Setting a value is needed for filtering scenario because otherwise we dont know when metadata will be updated for a pod once the customer adds the annotation. This way the fluent-bit filter starts dropping the logs within 60 seconds of adding the annotation. Let me know if 60s is too aggressive and needs to be increased to let's say 5minutes?
> 
> Also should this setting be applicable for metadata collection as well? I imagine that if not set then updates to labels/annotations done by the customer will take some time in getting reflected in the LA table column. @wanlonghenry is this something you have tested?

> @pfrcks , @wanlonghenry  - 60s seems good default. I think, we should also make this configurable and we can suggest the customer to configure if we need any of their scenario.

source/plugins/go/src/oms.go
wanlonghenry: yes Amol will be doing that after private preview

source/plugins/go/src/oms.go
wanlonghenry: this is if customer put all empty fields, it won't collect anything, we have tested this in bugbush 
image

source/plugins/go/src/oms.go
wanlonghenry: if this option is used, default collect all 
image

kubernetes/container-azm-ms-agentconfig.yaml
wanlonghenry: Will request this feedback from Ibrar

build/linux/installer/conf/fluent-bit.conf
wanlonghenry: image
verified

source/plugins/go/src/oms.go
wanlonghenry: because this kind of map structure is not directly compatible with the standard way JSON is unmarshalled into Go structs or maps (which typically use map[string]interface{}). So we need to convert it ourselves. I previously tested to directly use json.Unmarshal.

build/common/installer/scripts/fluent-bit-conf-customizer.rb
wanlonghenry: included in the scale test doc

source/plugins/go/src/oms.go
wanlonghenry: right now we just drop it if provided are not match 
image

build/common/installer/scripts/tomlparser.rb
wanlonghenry: if the include_fields does not match the data will be dropped

source/plugins/go/src/oms.go
wanlonghenry: The name in this format is to align with KubePodInventory table,
Though we can change from "case 'podUid':" to "case 'poduid':"

build/common/installer/scripts/tomlparser.rb
wanlonghenry: validation added but warning pending implementation

source/plugins/go/src/oms.go
wanlonghenry: added

build/common/installer/scripts/tomlparser.rb
wanlonghenry: seperated image into four fields but configmap for each field is pending implementation

source/plugins/go/src/oms.go
```diff
@@ -161,6 +161,10 @@ var (
 	ContainerLogSchemaV2 bool
 	// container log schema version from config map
 	ContainerLogV2ConfigMap bool
+	// Kubernetes Metadata enabled through configmap flag
+	KubernetesMetadataEnabled bool
+	// Kubernetes Metadata enabled exclude list
```
ganga1980: this should be "include" not "exclude" in the comment.

source/plugins/go/src/oms.go
```diff
@@ -1146,6 +1248,31 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+				}
+				kubernetesMetadata = string(kubernetesMetadataBytes)
+			} else {
+				message := fmt.Sprintf("Error while fetching kubernetesMetadataJson")
+				Log(message)
+				continue
+			}
+			elapsed := time.Since(start)
+			processingTimeMs := elapsed.Milliseconds();
+			SendMetric("K8sMetadataProcessingMs", processingTimeMs)
```
ganga1980: SendMetric takes 3 arguments not to 2. This needs to be fixed.

build/linux/installer/conf/fluent-bit.conf
```diff
@@ -44,3 +44,16 @@
 #${MultilineEnabled}    Match oms.container.log.la.*
 #${MultilineEnabled}    multiline.key_content log
 #${MultilineEnabled}    multiline.parser #${MultilineLanguages}
+
+#${KubernetesFilterEnabled}[FILTER]
```
ganga1980, line 50: we should make these work on geneva path also as internal teams are asking for this.

kubernetes/container-azm-ms-agentconfig.yaml
ganga1980: Add comment what's the behavior if the include_fields commented or not specified.

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -53,7 +53,13 @@ data:
           # If enabled will also stitch together container logs split by docker/cri due to size limits(16KB per log line) up to 64 KB.
           # enabled = "false"
           # stacktrace_languages = []
-
+       #[log_collection_settings.metadata_collection]
+          # if enabled will collect kubernetes metadata for ContainerLogv2 schema. Default false
+          # enabled = false
```
ganga1980, line 60: Looks like this is not working on MSI clusters, did you verify that, it works on MSI clusters?

kubernetes/container-azm-ms-agentconfig.yaml
ganga1980: Lets figure out what other fields we can add support as part of this feature.

source/plugins/go/src/oms.go
ganga1980: Similar to AgentLogProcessingMaxLatencyMs, we should compute . This metric should be  emitted from telemetry.go  so that enabletelemetry flag being honored.

kubernetes/container-azm-ms-agentconfig.yaml
wanlonghenry: added

kubernetes/container-azm-ms-agentconfig.yaml
wanlonghenry: on happens for eastus workspace

build/common/installer/scripts/tomlparser.rb

kubernetes/container-azm-ms-agentconfig.yaml
wanlonghenry: raw data contains majorly all those fields

build/common/installer/scripts/tomlparser.rb
wanlonghenry: warning added

build/linux/installer/conf/fluent-bit.conf
wanlonghenry: added for geneva

source/plugins/go/src/oms.go
wanlonghenry: SendMetric will be in seperate PR

wanlonghenry:
> > @ganga1980 @wanlonghenry current I am using Kube_Meta_Cache_TTL as 60s only for filtering scenario. From fluent-bit docs
> > `configurable TTL for K8s cached metadata. By default, it is set to 0 which means TTL for cache entries is disabled and cache entries are evicted at random when capacity is reached. In order to enable this option, you should set the number to a time interval. For example, set this value to 60 or 60s and cache entries which have been created more than 60s will be evicted.`
> > Setting a value is needed for filtering scenario because otherwise we dont know when metadata will be updated for a pod once the customer adds the annotation. This way the fluent-bit filter starts dropping the logs within 60 seconds of adding the annotation. Let me know if 60s is too aggressive and needs to be increased to let's say 5minutes?
> > Also should this setting be applicable for metadata collection as well? I imagine that if not set then updates to labels/annotations done by the customer will take some time in getting reflected in the LA table column. @wanlonghenry is this something you have tested?
> 
> > @pfrcks , @wanlonghenry  - 60s seems good default. I think, we should also make this configurable and we can suggest the customer to configure if we need any of their scenario.

For kubernetes metadata displaying the data get refreshed within 60s, so we should be good for not applying this to metadata collection as well. Tested both MSI and legacy cluster.
image
image

source/plugins/go/src/oms.go
```diff
@@ -1116,6 +1119,146 @@ func UpdateNumTelegrafMetricsSentTelemetry(numMetricsSent int, numSendErrors int
...
+	colonLocation := strings.Index(image, ":")
+	atLocation := strings.Index(image, "@")
+
+	// Exclude the digest part for imageRepo/image/tag parsing, if present
+	if atLocation != -1 {
+		image = image[:atLocation]
+	}
+
+	// Process Image Name, Repo, and Tag based on the original logic
+	if colonLocation != -1 {
```
jatakiajanvi12: Can't we just make colonLocation as the end of the list if it is not found? to avoid all the duplicate code?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -54,7 +54,14 @@ data:
           # Requires ContainerLogV2 schema to be enabled. See https://aka.ms/ContainerLogv2 for more details.
           # enabled = "false"
           # stacktrace_languages = []
-
+       #[log_collection_settings.metadata_collection]
+          # if enabled will collect kubernetes metadata for ContainerLogv2 schema. Default is false.
```
ganga1980: kube_meta_cache_ttl entry here  with comments about the default value and link to fluent-bit doc about this field.

source/plugins/go/src/oms.go
wanlonghenry: changed

kubernetes/container-azm-ms-agentconfig.yaml
wanlonghenry: added

Documentation/Internal/ContainerLogV2-Linux.xml
```diff
@@ -25,6 +25,8 @@
       <Column name="Computer" type="str" mdstype="mt:wstr" />
       <Column name="ContainerName" type="str" mdstype="mt:wstr" />
```
ganga1980, line 26: 1. bump the version 1.0 to 1.1
2.  lets validate these changes works for Geneva path.

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -62,12 +63,25 @@ def populateSettingValuesFromConfigMap(parsedConfig)
   end
 end
 
+def getKubernetesMetadataCacheTTL(parsedConfig)
+  begin
+    if !parsedConfig.nil? && !parsedConfig[:agent_settings].nil?
+      k8s_metadata_config = parsedConfig[:agent_settings][:k8s_metadata_config]
+      if !k8s_metadata_config.nil? && !k8s_metadata_config[:kube_meta_cache_ttl].nil?
```
ganga1980: since the units of this field is seconds, hence have the seconds prefix of this config entry, kube_meta_cache_ttl_secs

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -9,6 +9,7 @@
 @configSchemaVersion = ""
 
 @disableTelemetry = false
+@logEnableKubernetesMetadataCacheTTL = "60"
```
ganga1980: 1. have the seconds suffix for the variable name, i.e.  @logEnableKubernetesMetadataCacheTTLSeconds
2. use the integer instead of string.

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -62,12 +63,25 @@ def populateSettingValuesFromConfigMap(parsedConfig)
   end
 end
 
+def getKubernetesMetadataCacheTTL(parsedConfig)
+  begin
+    if !parsedConfig.nil? && !parsedConfig[:agent_settings].nil?
+      k8s_metadata_config = parsedConfig[:agent_settings][:k8s_metadata_config]
+      if !k8s_metadata_config.nil? && !k8s_metadata_config[:kube_meta_cache_ttl].nil?
+        @logEnableKubernetesMetadataCacheTTL = k8s_metadata_config[:kube_meta_cache_ttl]
```
ganga1980: check the provided  value in the config is integer or not  and also check for the value is >=0

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -62,12 +63,25 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+  end
+end
+
 @configSchemaVersion = ENV["AZMON_AGENT_CFG_SCHEMA_VERSION"]
 puts "****************Start Config Processing********************"
 if !@configSchemaVersion.nil? && !@configSchemaVersion.empty? && @configSchemaVersion.strip.casecmp("v1") == 0 #note v1 is the only supported schema version , so hardcoding it
   configMapSettings = parseConfigMap
   if !configMapSettings.nil?
     populateSettingValuesFromConfigMap(configMapSettings)
+    getKubernetesMetadataCacheTTL(configMapSettings)
```
ganga1980: you dont need this and populateSettingValuesFromConfigMap should take care of this.

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -87,6 +101,8 @@ def get_command_windows(env_variable_name, env_variable_value)
       commands = get_command_windows("DISABLE_TELEMETRY", @disableTelemetry)
       file.write(commands)
     end
+    commands = get_command_windows("AZMON_KUBERNETES_METADATA_CACHE_TTL", @logEnableKubernetesMetadataCacheTTL)
```
ganga1980: rename this to AZMON_KUBERNETES_METADATA_CACHE_TTL_SECONDS

build/common/installer/scripts/tomlparser.rb
```diff
@@ -228,6 +231,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+          if include_fields.empty?
+            puts "config::Include fields specified for Kubernetes metadata is empty, disabling Kubernetes metadata"
```
ganga1980: fix the log message to include warn for example, "config::WARN Include fields specified for Kubernetes metadata is empty, disabling Kubernetes metadata"

build/common/installer/scripts/tomlparser.rb
```diff
@@ -228,6 +231,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+            any_field_match = include_fields.any? { |field| predefined_fields.include?(field) }
+            if any_field_match
+              @logKubernetesMetadataIncludeFields = include_fields.join(",")
+            else
+              puts "config:: WARN: Include fields specified for Kubernetes metadata does not match any predefined fields, disabling Kubernetes metadata"
+              @logEnableKubernetesMetadata = false
+            end
+          end
+        end
+      end
```
ganga1980, line 258: add rescue block and log the exception if there is any.

build/common/installer/scripts/tomlparser.rb
```diff
@@ -228,6 +231,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+    end
+
+    #Get annotation based log filtering setting
+    begin
+      if !parsedConfig[:log_collection_settings][:filter_using_annotations].nil? && !parsedConfig[:log_collection_settings][:filter_using_annotations][:enabled].nil?
+        puts "config::Using config map setting for annotation based log filtering"
+        @annotationBasedLogFiltering = parsedConfig[:log_collection_settings][:filter_using_annotations][:enabled]
```
ganga1980, line 267: add rescue block and log the exception if there is any.

build/linux/installer/conf/fluent-bit-geneva.conf
```diff
@@ -33,6 +33,19 @@
...
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     oms.container.log.la.var.log.containers.
```
ganga1980: this has to be  **geneva.container.log.containers.**

build/windows/installer/conf/fluent-bit-geneva.conf
```diff
@@ -30,6 +30,19 @@
...
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     oms.container.log.la.var.log.containers.
+
```
ganga1980: this has to be  **geneva.container.log.containers.**

build/linux/installer/conf/fluent-bit-geneva.conf
```diff
@@ -33,6 +33,19 @@
     Record Computer ${HOSTNAME}
     Record AzureResourceId ${AKS_RESOURCE_ID}
 
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
```
ganga1980: this has to be **geneva.container.log.***

build/windows/installer/conf/fluent-bit-geneva.conf
```diff
@@ -30,6 +30,19 @@
     Record Computer ${HOSTNAME}
     Record AzureResourceId ${AKS_RESOURCE_ID}
 
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
```
ganga1980: this has to be **geneva.container.log.***

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -304,6 +307,15 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
                         telemetryProps["multilineLanguages"] = @clusterMultilineLanguages
                       end
                     end
+                    if (!@clusterKubernetesMetadataEnabled.nil? && !@clusterKubernetesMetadataEnabled.empty?)
+                      telemetryProps["metadataEnabled"] = @clusterKubernetesMetadataEnabled
+                    end
+                    if (!@clusterKubernetesMetadataiIncludeFields.nil? && !@clusterKubernetesMetadataiIncludeFields.empty?)
+                      telemetryProps["metadataIncludeFields"] = @clusterKubernetesMetadataiIncludeFields
+                    end
```
ganga1980, line 316: typo and extra "i". should be clusterKubernetesMetadataIncludeFields

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -214,6 +222,10 @@ data:
     # The following settings are "undocumented", we don't recommend uncommenting them unless directed by Microsoft.
     # [agent_settings.telemetry_config]
     #   disable_telemetry = false  # if this is not applied, default value is false
+    
+    # The following setting is for KubernetesMetadata CacheTTL Settings, default is 60s.
+    # [agent_settings.k8s_metadata_config]
+    #   kube_meta_cache_ttl = 60  # if this is not applied, default value is false
```
ganga1980: 1. rename the field name to have suffix seconds i.e. kube_meta_cache_ttl_seconds
2. fix the comment, "if this is not applied, default value is false" since this is not correct.

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -113,6 +115,16 @@ def substituteFluentBitPlaceHolders
       new_contents = new_contents.gsub("\n    ${TAIL_IGNORE_OLDER}\n", "\n")
     end
 
+    if !kubernetesMetadataCollection.nil? && kubernetesMetadataCollection.to_s.downcase == "true"
+      new_contents = new_contents.gsub("#${KubernetesFilterEnabled}", "")
+    end
+
+    if !annotationBasedLogFiltering.nil? && annotationBasedLogFiltering.to_s.downcase == "true"
+      # enabled kubernetes fluent-bit plugin if not already enabled
```
ganga1980: should be kubernetes **filter** plugin

build/common/installer/scripts/fluent-bit-geneva-conf-customizer.rb
```diff
@@ -82,6 +84,16 @@ def substituteFluentBitPlaceHolders(configFilePath)
       new_contents = new_contents.gsub("\n    ${TAIL_THREADED}\n", "\n")
     end
 
+    if !kubernetesMetadataCollection.nil? && kubernetesMetadataCollection.to_s.downcase == "true"
+      new_contents = new_contents.gsub("#${KubernetesFilterEnabled}", "")
+    end
+
+    if !annotationBasedLogFiltering.nil? && annotationBasedLogFiltering.to_s.downcase == "true"
+      # enabled kubernetes fluent-bit plugin if not already enabled
```
ganga1980: should be kubernetes filter plugin

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -304,6 +307,15 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
                         telemetryProps["multilineLanguages"] = @clusterMultilineLanguages
                       end
                     end
+                    if (!@clusterKubernetesMetadataEnabled.nil? && !@clusterKubernetesMetadataEnabled.empty?)
+                      telemetryProps["metadataEnabled"] = @clusterKubernetesMetadataEnabled
+                    end
+                    if (!@clusterKubernetesMetadataiIncludeFields.nil? && !@clusterKubernetesMetadataiIncludeFields.empty?)
```
ganga1980: typo and extra "i". should be clusterKubernetesMetadataIncludeFields

please fix across all the places

build/common/installer/scripts/tomlparser.rb
```diff
@@ -228,6 +231,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
       ConfigParseErrorLogger.logError("Exception while reading config map settings for adx database name - #{errorStr}, using default #{@adxDatabaseName}, please check config map for errors")
     end
+
+    #Get Kubernetes Metadata setting
+    begin
+      if !parsedConfig[:log_collection_settings][:metadata_collection].nil? && !parsedConfig[:log_collection_settings][:metadata_collection][:enabled].nil?
+        puts "config::Using config map setting for kubernetes metadata"
+        @logEnableKubernetesMetadata = parsedConfig[:log_collection_settings][:metadata_collection][:enabled]
+        if !parsedConfig[:log_collection_settings][:metadata_collection][:include_fields].nil?
+          puts "config::Using config map setting for kubernetes metadata include fields"
```
ganga1980: missing "INFO"  in the log message should be "config::INFO Using config map setting for kubernetes metadata include fields"

build/common/installer/scripts/tomlparser.rb
ganga1980: missing "INFO"  in the log message

build/common/installer/scripts/tomlparser.rb
```diff
@@ -228,6 +231,40 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+            end
+          end
+        end
+      end
+    end
+
+    #Get annotation based log filtering setting
+    begin
+      if !parsedConfig[:log_collection_settings][:filter_using_annotations].nil? && !parsedConfig[:log_collection_settings][:filter_using_annotations][:enabled].nil?
+        puts "config::Using config map setting for annotation based log filtering"
```
ganga1980: missing "INFO"  in the log message

build/windows/installer/conf/fluent-bit.conf
```diff
@@ -45,5 +45,18 @@
     Match oms.container.log.la.*
     Exclude stream ${AZMON_LOG_EXCLUSION_REGEX_PATTERN}
 
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
```
ganga1980, line 53: Did you validate that, this works for windows containers? if not, lets validate this.

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -54,7 +54,15 @@ data:
...
           # enabled = "false"
           # stacktrace_languages = []
-
+       #[log_collection_settings.metadata_collection]
+          # kube_meta_cache_ttl is a configurable option for K8s cached metadata. Default is 60s. You may adjust it in below section [agent_settings.k8s_metadata_config]. Reference link: https://docs.fluentbit.io/manual/pipeline/filters/kubernetes#configuration-parameters
+          # if enabled will collect kubernetes metadata for ContainerLogv2 schema. Default is false. 
+          # enabled = false
+          # if include_fields commented out or empty, all fields will be included. If include_fields is set, only the fields listed will be included.
+          # include_fields = ["podLabels","podAnnotations","podUid","image","imageID","imageRepo","imageTag"]
+       #[log_collection_settings.filter_using_annotations]
```
ganga1980, line 63: @pfrcks , we would need to validate this works for Geneva path also since  cosmic team asking for this.

ganga1980, line 63: Please add fluent-bit link to supported annotations or give examples for  the annotations.

source/plugins/go/src/oms.go
```diff
@@ -1149,6 +1152,141 @@ func UpdateNumTelegrafMetricsSentTelemetry(numMetricsSent int, numSendErrors int
...
+
+	for _, include := range includesList {
+		switch include {
+		case "poduid":
+			if val, ok := kubernetesMetadataMap["pod_id"]; ok {
+				includedMetadata["podUid"] = val
+			}
+		case "podlabels":
+			if val, ok := kubernetesMetadataMap["labels"]; ok {
+				includedMetadata["podLabels"] = val
```
ganga1980, line 1185: Did you test what max labels size can be ingested? is it 64KB? if yes, then we should cap up to  64KB in the agent  and then discard more than 64kB.

source/plugins/go/src/oms.go
```diff
@@ -1149,6 +1152,141 @@ func UpdateNumTelegrafMetricsSentTelemetry(numMetricsSent int, numSendErrors int
...
+		case "podannotations":
+			if val, ok := kubernetesMetadataMap["annotations"]; ok {
+				if annotationsMap, ok := val.(map[string]interface{}); ok {
+					filteredAnnotations := make(map[string]interface{})
+					for key, annotationValue := range annotationsMap {
+						if !strings.Contains(key, "kubernetes.io/config") {
+							filteredAnnotations[key] = annotationValue
+						}
+					}
+					includedMetadata["podAnnotations"] = filteredAnnotations
```
ganga1980, line 1196: similarly for this as well, Did you test what max annotation size can be ingested? is it 64KB? if yes, then we should cap up to  64KB in the agent  and then discard annotations which are more than 64KB.

source/plugins/go/src/oms.go
```diff
@@ -1149,6 +1152,141 @@ func UpdateNumTelegrafMetricsSentTelemetry(numMetricsSent int, numSendErrors int
 	ContainerLogTelemetryMutex.Unlock()
 }
 
+func processIncludes(kubernetesMetadataMap map[string]interface{}, includesList []string) map[string]interface{} {
```
ganga1980, line 1155: Did you test with short lived (for example, 10s, 30s) Kubernetes jobs/cronjobs to see if filter plugin able to collect metadata?

build/windows/installer/conf/fluent-bit.conf
wanlonghenry: yes A value of 0 results in no limit, and the buffer will expand as-needed. For windows container, the data is flowing good

Documentation/Internal/ContainerLogV2-Linux.xml
wanlonghenry: changed

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: changed

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: changed

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: changed

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: the getKubernetesMetadataCacheTTLSeconds is a separate function

build/common/installer/scripts/tomlparser.rb
wanlonghenry: changed

build/common/installer/scripts/tomlparser.rb
wanlonghenry: changed

build/common/installer/scripts/tomlparser.rb
wanlonghenry: changed

build/linux/installer/conf/fluent-bit-geneva.conf
wanlonghenry: changed

build/windows/installer/conf/fluent-bit-geneva.conf
wanlonghenry: changed

build/linux/installer/conf/fluent-bit-geneva.conf
wanlonghenry: changed

build/windows/installer/conf/fluent-bit-geneva.conf
wanlonghenry: changed

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
wanlonghenry: changed

kubernetes/container-azm-ms-agentconfig.yaml
wanlonghenry: changed

source/plugins/go/src/oms.go
wanlonghenry: yes it works fine 
image
image

source/plugins/go/src/oms.go
wanlonghenry: A kubernete label key/value must be no longer than 63 characters

source/plugins/go/src/oms.go
wanlonghenry: for annotations, the Kubernetes api restrict max to 256 KB and LA ingestion is 64KB

wanlonghenry: image

build/linux/installer/conf/fluent-bit-geneva.conf
```diff
@@ -33,6 +33,19 @@
...
+#${KubernetesFilterEnabled}    Match               geneva.container.log.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL_SECONDS}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     geneva.container.log.containers.
```
wanlonghenry: needs to adjust tag

build/windows/installer/conf/fluent-bit-geneva.conf
```diff
@@ -30,6 +30,19 @@
...
+#${KubernetesFilterEnabled}    Match               geneva.container.log.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL_SECONDS}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     geneva.container.C.log.containers.
```
wanlonghenry: needs to adjust tag

build/linux/installer/conf/fluent-bit-geneva-logs_tenant.conf
```diff
@@ -29,3 +29,16 @@
     Workers                         ${OUTPUT_FORWARD_WORKERS_COUNT}
     Retry_Limit                     ${OUTPUT_FORWARD_RETRY_LIMIT}
     Require_ack_response            ${REQUIRE_ACK_RESPONSE}
+
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               geneva.container.log.*
```
pfrcks: should be geneva.container.log.tenant.*


also move this between input and output

build/linux/installer/conf/fluent-bit-geneva-logs_infra.conf
```diff
@@ -25,3 +25,16 @@
     Retry_Limit                     10
     TelemetryPushIntervalSeconds    300
     Match                           *_<INFRA_NAMESPACE>_*
+
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               geneva.container.log.*
```
pfrcks: should be geneva.container.log.infra.*


also move this between input and output

build/windows/installer/conf/fluent-bit-geneva-logs_infra.conf
```diff
@@ -25,3 +25,16 @@
     Retry_Limit                     10
     TelemetryPushIntervalSeconds    300
     Match                           *_<INFRA_NAMESPACE>_*
+
+#${KubernetesFilterEnabled}[FILTER]
```
pfrcks: same comment as in linux

build/windows/installer/conf/fluent-bit-geneva-logs_tenant.conf
```diff
@@ -29,3 +29,16 @@
     Workers                         ${OUTPUT_FORWARD_WORKERS_COUNT}
     Retry_Limit                     ${OUTPUT_FORWARD_RETRY_LIMIT}
     Require_ack_response            ${REQUIRE_ACK_RESPONSE}
+
+#${KubernetesFilterEnabled}[FILTER]
```
pfrcks: same comment as in linux

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -28,6 +28,9 @@ class CAdvisorMetricsAPIClient
   @clusterContainerLogSchemaVersion = ENV["AZMON_CONTAINER_LOG_SCHEMA_VERSION"]
   @clusterMultilineEnabled = ENV["AZMON_MULTILINE_ENABLED"]
   @clusterMultilineLanguages = ENV["AZMON_MULTILINE_LANGUAGES"]
+  @clusterKubernetesMetadataEnabled = ENV["AZMON_KUBERNETES_METADATA_ENABLED"]
```
pfrcks, line 31: can you add telemetry for annotation filtering as well?

build/linux/installer/conf/fluent-bit-geneva-logs_tenant.conf
wanlonghenry: changed

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
wanlonghenry: added

pfrcks (dismissed):
lgtm

build/common/installer/scripts/fluent-bit-conf-customizer.rb
```diff
@@ -77,6 +77,8 @@ def substituteFluentBitPlaceHolders
     stacktraceLanguages = ENV["AZMON_MULTILINE_LANGUAGES"]
     resourceOptimizationEnabled = ENV["AZMON_RESOURCE_OPTIMIZATION_ENABLED"]
     windowsFluentBitDisabled = ENV["AZMON_WINDOWS_FLUENT_BIT_DISABLED"]
+    kubernetesMetadataCollection = ENV["AZMON_KUBERNETES_METADATA_ENABLED"]
```
ganga1980, line 80: I think, changes required in the xml - https://github.com/microsoft/Docker-Provider/blob/ci_prod/kubernetes/linux/mdsd.xml  for new fields. Did you validate these changes for ContainerLogV2 in legacy mode for both Linux and windows?

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -62,12 +63,32 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+  begin
+    if !parsedConfig.nil? && !parsedConfig[:agent_settings].nil?
+      k8s_metadata_config = parsedConfig[:agent_settings][:k8s_metadata_config]
+      if !k8s_metadata_config.nil? && !k8s_metadata_config[:kube_meta_cache_ttl_secs].nil?
+        ttl_value = k8s_metadata_config[:kube_meta_cache_ttl_secs]
+        if ttl_value.is_a?(Integer) && ttl_value >= 0
+          @logEnableKubernetesMetadataCacheTTLSeconds = ttl_value
+          puts "config::INFO: Using config map value: kube_meta_cache_ttl_secs = #{@logEnableKubernetesMetadataCacheTTLSeconds}"
+        else
+          puts "config::WARN: Invalid kube_meta_cache_ttl_secs value: Must be an integer and >= 0"
```
ganga1980: are we not taking default if customer provide value less than 0 or non-integer? If yes, we should fix the log message "config::WARN: Using the default value for kube_meta_cache_ttl_secs since provided config value is invalid"

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -62,12 +63,32 @@ def populateSettingValuesFromConfigMap(parsedConfig)
...
+  end
+end
+
 @configSchemaVersion = ENV["AZMON_AGENT_CFG_SCHEMA_VERSION"]
 puts "****************Start Config Processing********************"
 if !@configSchemaVersion.nil? && !@configSchemaVersion.empty? && @configSchemaVersion.strip.casecmp("v1") == 0 #note v1 is the only supported schema version , so hardcoding it
   configMapSettings = parseConfigMap
   if !configMapSettings.nil?
     populateSettingValuesFromConfigMap(configMapSettings)
+    getKubernetesMetadataCacheTTLSeconds(configMapSettings)
```
ganga1980: You dont need this and code under populateSettingValuesFromConfigMap can be moved to populateSettingValuesFromConfigMap function.

build/common/installer/scripts/tomlparser-common-agent-config.rb
```diff
@@ -9,6 +9,7 @@
 @configSchemaVersion = ""
 
 @disableTelemetry = false
+@logEnableKubernetesMetadataCacheTTLSeconds = 60
```
ganga1980, line 12: is this default used when there is no configmap as well?

build/linux/installer/conf/fluent-bit-geneva-logs_infra.conf
```diff
@@ -18,6 +18,19 @@
...
+#${KubernetesFilterEnabled}    Match               geneva.container.log.infra.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL_SECONDS}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     geneva.container.log.infra.var.log.containers.
```
ganga1980, line 32: Doesnt this needs to contain "C.var.log.container"  like in other places?

build/linux/installer/conf/fluent-bit-geneva-logs_tenant.conf
```diff
@@ -18,6 +18,19 @@
...
+#${KubernetesFilterEnabled}    Match               geneva.container.log.tenant.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL_SECONDS}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     geneva.container.log.tenant.var.log.containers.
```
ganga1980: similarly here.. Doesnt this needs to contain "C.var.log.container"  like in other places?

build/linux/installer/conf/fluent-bit.conf
```diff
@@ -47,5 +47,18 @@
...
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
+#${KubernetesFilterEnabled}    Merge_Log           Off
+#${KubernetesFilterEnabled}    Buffer_Size         0
+#${KubernetesFilterEnabled}    Use_Kubelet         true
+#${KubernetesFilterEnabled}    Kubelet_Port        10250
+#${KubernetesFilterEnabled}    Kubelet_Host        ${NODE_IP}
+#${KubernetesFilterEnabled}    tls.verify          Off
+#${KubernetesFilterEnabled}#${AnnotationBasedLogFilteringEnabled}    K8S-Logging.Exclude On
+#${KubernetesFilterEnabled}    Kube_Meta_Cache_TTL ${AZMON_KUBERNETES_METADATA_CACHE_TTL_SECONDS}
+#${KubernetesFilterEnabled}    Kube_Tag_Prefix     oms.container.log.la.var.log.containers.
```
ganga1980, line 62: same here .. Doesnt this needs to contain "C.var.log.container"  like in other places?

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: when there is no config map, this feild will not be used

build/linux/installer/conf/fluent-bit.conf
wanlonghenry: for win only

build/linux/installer/conf/fluent-bit-geneva-logs_tenant.conf
wanlonghenry: for win only

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: changed

build/common/installer/scripts/tomlparser-common-agent-config.rb
wanlonghenry: changed

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
```diff
@@ -304,6 +308,18 @@ def getContainerCpuMetricItems(metricJSON, hostName, cpuMetricNameToCollect, met
                         telemetryProps["multilineLanguages"] = @clusterMultilineLanguages
                       end
                     end
+                    if (!@clusterKubernetesMetadataEnabled.nil? && !@clusterKubernetesMetadataEnabled.empty?)
+                      telemetryProps["metadataEnabled"] = @clusterKubernetesMetadataEnabled
+                    end
```
jatakiajanvi12, line 313: Why are we sending the telemetry from here? Shouldn't this all be sent as part of ContainerLog event or something? 
Also, if you decide to add the telemetry here, now that this is all moved to golang as well, you will have to add the telemetry there as well

build/common/installer/scripts/fluent-bit-conf-customizer.rb
wanlonghenry: changed

source/plugins/ruby/CAdvisorMetricsAPIClient.rb
wanlonghenry: discussed with Ganga this is a valid point, will keep this and add the telemetry in go as well

build/linux/installer/conf/fluent-bit-geneva-logs_infra.conf
```diff
@@ -18,6 +18,19 @@
     Exclude_Path ${AZMON_CLUSTER_LOG_TAIL_EXCLUDE_PATH}
     ${TAIL_THREADED}
 
+#${KubernetesFilterEnabled}[FILTER]
```
ganga1980: This should be moved to fluent-bit-geneva.conf after multi-line and this similar to multi-line filter

build/linux/installer/conf/fluent-bit-geneva-logs_tenant.conf
```diff
@@ -18,6 +18,19 @@
     Exclude_Path ${AZMON_CLUSTER_LOG_TAIL_EXCLUDE_PATH}
     ${TAIL_THREADED}
 
+#${KubernetesFilterEnabled}[FILTER]
```
ganga1980: This should be moved to fluent-bit-geneva.conf after multi-line and this similar to multi-line filter . We only need one filter all up geneva.

build/windows/installer/conf/fluent-bit-geneva-logs_infra.conf
```diff
@@ -18,6 +18,19 @@
     Exclude_Path ${AZMON_CLUSTER_LOG_TAIL_EXCLUDE_PATH}
     ${TAIL_THREADED}
 
+#${KubernetesFilterEnabled}[FILTER]
```
ganga1980: This should be moved to fluent-bit-geneva.conf after multi-line and this similar to multi-line filter

build/windows/installer/conf/fluent-bit-geneva-logs_tenant.conf
```diff
@@ -18,6 +18,19 @@
     Exclude_Path ${AZMON_CLUSTER_LOG_TAIL_EXCLUDE_PATH}
     ${TAIL_THREADED}
 
+#${KubernetesFilterEnabled}[FILTER]
```
ganga1980: This should be moved to fluent-bit-geneva.conf after multi-line and this similar to multi-line filter

Documentation/Internal/ContainerLogV2-Linux.xml
```diff
@@ -25,6 +25,8 @@
...
       <Column name="AzureResourceId" type="str" mdstype="mt:wstr" />
+      <Column name="KubernetesMetadata" type="str" mdstype="mt:wstr" />
+      <Column name="LogLevel" type="str" mdstype="mt:wstr" />
```
ganga1980: LogLevel is not supported in geneva and this can be removed.

build/windows/installer/conf/fluent-bit.conf
```diff
@@ -45,5 +45,18 @@
     Match oms.container.log.la.*
     Exclude stream ${AZMON_LOG_EXCLUSION_REGEX_PATTERN}
 
+#${KubernetesFilterEnabled}[FILTER]
+#${KubernetesFilterEnabled}    Name                kubernetes
+#${KubernetesFilterEnabled}    Match               oms.container.log.la.*
```
ganga1980, line 51: Add the  alias "#${KubernetesFilterEnabled} Alias oms_kubernetes_filter"  after Name . This will be useful for the metrics and troubleshooting.

build/linux/installer/conf/fluent-bit-geneva-logs_infra.conf
ganga1980: Also add the alias  "#${KubernetesFilterEnabled} Alias geneva_kubernetes_filter" after Name . This will be useful for the metrics and troubleshooting.

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -1051,6 +1051,18 @@ func getContainerCpuMetricItems(metricInfo map[string]interface{}, hostName, met
 						if len(os.Getenv("AZMON_RESOURCE_OPTIMIZATION_ENABLED")) > 0 {
 							telemetryProps["resoureceOptimizationEnabled"] = os.Getenv("AZMON_RESOURCE_OPTIMIZATION_ENABLED")
 						}
+						if len(os.Getenv("AZMON_KUBERNETES_METADATA_ENABLED")) > 0 {
```
ganga1980, line 1054: add dashboard tile to track the adoption.

source/plugins/go/src/oms.go
```diff
@@ -1299,6 +1437,27 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
...
+				if err != nil {
+					message := fmt.Sprintf("Error while Marshalling kubernetesMetadataBytes to json bytes: %s", err.Error())
+					Log(message)
+					SendException(message)
+				}
+				kubernetesMetadata = string(kubernetesMetadataBytes)
+			} else {
+				message := fmt.Sprintf("Error while fetching kubernetesMetadataJson")
+				Log(message)
+				continue
```
ganga1980: we shouldn't skip the ingesting the log record if failed to fetch Kubernetes metadata.

wanlonghenry: remove the continue and add SendException

Documentation/Internal/ContainerLogV2-Linux.xml
wanlonghenry: changed

source/plugins/go/src/oms.go
pfrcks: is this Log too verbose?

Documentation/Internal/ContainerLogV2-Linux.xml
wanlonghenry: removed

build/linux/installer/conf/fluent-bit-geneva-logs_infra.conf
wanlonghenry: added

build/windows/installer/conf/fluent-bit.conf
wanlonghenry: added

source/plugins/go/src/oms.go
wanlonghenry: changed to Error while getting kubernetesMetadata

source/plugins/go/input/lib/cadvisor.go
wanlonghenry: will add that

ganga1980 (approved):

PR #1150, opened 2024-01-10
make dotnet conform better to stacktrace format

pfrcks:
the start_state was unnecessarily broad in scope. Reduced its scope to match .NET stacktraces better. 

Eg. it shouldn't match java exceptions which are similar pattern eg: redis.clients.jedis.exceptions.JedisConnectionException: java.net.SocketException: Broken pipe (Write failed)

New start_state regex rubular link https://rubular.com/r/JJmfqcpw6eTacN  

Tested on CI/CD multiline cluster


jatakiajanvi12 (commented):
Can you add examples of what used to match earlier and how that has changed?

PR #1154, opened 2024-01-13
Gangams/fips updates

ganga1980:
Changes for the FIPs updates. Verified on FIPS enabled nodes - Azure Linux and Ubuntu and also Windows nodes (2019 & 2022).   Update to use the openssl shipped with distroless image instead of copying the binaries from base to distroless image.  For FIPs validation, Mariner team introduced HMAC verification on the binaries and openssl binaries will have associated .hmac files.  Copying just binaries will cause the Openssl FIP test failures and recommendation for Mariner team to use the Openssl shipped in distroless image.

### DIFF
--- a/kubernetes/linux/Dockerfile.multiarch
+++ b/kubernetes/linux/Dockerfile.multiarch
@@ -103,7 +103,7 @@ COPY --from=builder /opt/microsoft/azure-mdsd/lib/libsymcrypt.so.103 /opt/micros
 # logrotate dependencies
 COPY --from=builder /lib/libselinux.so.1 /lib/libpopt.so.0 /lib/libc.so.6 /lib/libpcre.so.1 /lib/
 # curl dependencies
-COPY --from=builder /lib/libcurl.so.4 /lib/libz.so.1 /usr/lib/libc.so.6 /lib/libnghttp2.so.14 /lib/libssl.so.1.1 /lib/libssh2.so.1 /lib/libcrypto.so.1.1 /lib/libgssapi_krb5.so.2 /lib/libzstd.so.1 /lib/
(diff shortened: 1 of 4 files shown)


kubernetes/linux/Dockerfile.multiarch
```diff
@@ -22,7 +22,7 @@ ENV tmpdir /opt
 RUN tdnf clean all
 RUN tdnf repolist --refresh
 RUN tdnf -y update
-RUN tdnf install -y build-essential wget openssl curl sudo net-tools cronie rsyslog dmidecode gnupg make logrotate busybox gawk tar && rm -rf /var/lib/apt/lists/*
```
pfrcks, line 25: does this mean openssl components are already present in the distroless image?

ganga1980: yes, openssl already available in distroless. Mariner team recommends to use the version in distroless instead of copying binaries from base to distroless.

source/plugins/ruby/in_kube_nodes.rb
```diff
@@ -603,6 +603,10 @@ def getNodeTelemetryProps(item)
         end
         properties["NODES_CHUNK_SIZE"] = @NODES_CHUNK_SIZE
         properties["NODES_EMIT_STREAM_BATCH_SIZE"] = @NODES_EMIT_STREAM_BATCH_SIZE
+        nodeLabels = item["metadata"]["labels"]
+        if !nodeLabels.nil? && !nodeLabels.empty? && nodeLabels.key?("kubernetes.azure.com/fips_enabled") && nodeLabels["kubernetes.azure.com/fips_enabled"] == "true"
```
wanlonghenry, line 607: nodeLabels.key? might not be needed, as nil == "true" will return false

ganga1980: yes, its not needed but checking there is no harm as well.

wanlonghenry (approved):

PR #1159, opened 2024-01-24
Gangams/telegraf update to 1.28.5

ganga1980:
This PR has following updates
1.  Linux Agent, Telegraf update from 1.27.3 to 1.28.5
3.  update from response_timeout to timeout since response_timeout deprecated and new field name is timeout
4.  comment bearer_token since this is not required and also to address customer incident
5. re-enable timebound token

scenarios validaed
1.  Pod annotation scraping for both Linux (prom sidecar) & Windows
2.  Cluster level service endpoint scraping through Replica
3.  default metrics for both Linux and windows

Note: Telegraf upgrade for windows requires more investigation as the latest versions doesnt work.


ganga1980:
ama-logs-rs memory usage(p99)
image
ama-logs-rs cpu usage(p99)
image
ama-logs ds memory usage (p99)
image
ama-logs ds cpu usage (p99)
image

build/windows/installer/conf/telegraf.conf
```diff
@@ -124,7 +124,7 @@
 #Prometheus Custom Metrics
 [[inputs.prometheus]]
   interval = "$AZMON_TELEGRAF_CUSTOM_PROM_INTERVAL"
-  
+
```
jatakiajanvi12: Unnecessary change in file

build/linux/installer/conf/telegraf-prom-side-car.conf
```diff
@@ -184,12 +192,12 @@
   # kube_config = "/path/to/kubernetes.config"
 
   ## Use bearer token for authorization. ('bearer_token' takes priority)
-  bearer_token = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+  # bearer_token = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
jatakiajanvi12, line 195: Should we just remove instead of commenting? Or does it get used somewhere?

build/linux/installer/conf/telegraf-prom-side-car.conf
```diff
@@ -66,7 +66,15 @@
   ## Run telegraf in quiet mode (error log messages only).
   quiet = true
   ## Specify the log file name. The empty string means to log to stderr.
-  logfile = ""
+  logfile = "/var/opt/microsoft/docker-cimprov/log/telegraf_error.log"
+
+  ## Maximum number of rotated archives to keep, any older logs are deleted. If set to -1, no archives are removed.
+  logfile_rotation_max_archives = 3
```
jatakiajanvi12, line 72: Do we need 3 log files as archives?

build/linux/installer/conf/telegraf-prom-side-car.conf
jatakiajanvi12: How was it affecting the customer?

charts/azuremonitor-containers/templates/ama-logs-daemonset-windows.yaml
```diff
@@ -98,6 +98,9 @@ spec:
        - name: SIDECAR_SCRAPING_ENABLED
          value: {{ .Values.amalogs.sidecarscraping | quote }}
        volumeMounts:
+        - name: kube-api-access
+          mountPath: /var/run/secrets/kubernetes.io/serviceaccount
```
jatakiajanvi12: Does this timebound token work for the scenario of the sev2? Can you add that in the testing scenarios you mentioned as well?

jatakiajanvi12:
> ama-logs-rs memory usage(p99) image ama-logs-rs cpu usage(p99) image ama-logs ds memory usage (p99) image ama-logs ds cpu usage (p99) image

Why is cpu increasing for RS?

build/windows/installer/conf/telegraf.conf
ganga1980: > Unnecessary change in file

Yes, this is number of log files to keep and 3 should be fine with each 1MB. other places we have 5 - https://github.com/microsoft/Docker-Provider/blob/ci_prod/kubernetes/linux/main.sh#L940.

build/linux/installer/conf/telegraf-prom-side-car.conf
ganga1980: > Do we need 3 log files as archives?

Yes, this is number of log files to keep and 3 should be fine with each 1MB. other places we have 5 - https://github.com/microsoft/Docker-Provider/blob/ci_prod/kubernetes/linux/main.sh#L940.

ganga1980:
> > ama-logs-rs memory usage(p99) image ama-logs-rs cpu usage(p99) image ama-logs ds memory usage (p99) image ama-logs ds cpu usage (p99) image
> 
> Why is cpu increasing for RS?

> the cpu increase is very little i.e. 25millis to 45millis and this can be ignored. This increase because of the telegraf.

charts/azuremonitor-containers/templates/ama-logs-daemonset-windows.yaml
ganga1980: > Does this timebound token work for the scenario of the sev2? Can you add that in the testing scenarios you mentioned as well?

> With the updated telegraf has the fix for refreshing the token. Yes, this works.  There is no special testing scenario and just applying the yaml to this cluster should restrict the token expiry for an hour and all the components should refresh the token before the expiry.

build/linux/installer/conf/telegraf-prom-side-car.conf
ganga1980: > How was it affecting the customer?

> customer app validating this token and this token shouldnt be sent as part of the request since all pod based scraping happens without any auth.

jatakiajanvi12 (dismissed):

jatakiajanvi12 (approved):

PR #1173, opened 2024-02-21
Gangams/geneva msi mode bug fix

ganga1980:
This PR has following changes
1. Geneva config xml for the cluster resourceId, region and computer fields for identity
2.  HOSTNAME environment variable addition for windows container
3. bug fix for the msi mode for the linux

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -825,6 +825,10 @@ if [ "${CONTAINER_TYPE}" != "PrometheusSidecar" ] && isGenevaMode; then
     echo "export MDSD_FLUENT_SOCKET_PORT=$MDSD_FLUENT_SOCKET_PORT" >> ~/.bashrc
     export SSL_CERT_FILE="/etc/pki/tls/certs/ca-bundle.crt"
     echo "export SSL_CERT_FILE=$SSL_CERT_FILE" >> ~/.bashrc
+    if [ "${USING_AAD_MSI_AUTH}" == "true" ]; then
+       export AAD_MSI_AUTH_MODE=true
+       echo "export AAD_MSI_AUTH_MODE=true" >> ~/.bashrc
+    fi
 else
       if [ "${USING_AAD_MSI_AUTH}" == "true" ]; then
             echo "*** setting up oneagent in aad auth msi mode ***"
--- a/source/plugins/go/src/extension/extension_linux.go
+++ b/source/plugins/go/src/extension/extension_linux.go
@@ -3,12 +3,14 @@
 package extension
 
 import (
-	"fmt"
 	"os"
 	"strings"
 
 	"github.com/ugorji/go/codec"
 )
+
+const FluentSocketAddress = "/var/run/mdsd-ci/default_fluent.socket"
+const FluentSocketAddressPrometheysSidecar = "/var/run/mdsd-PrometheusSidecar/default_fluent.socket"
 
 func getExtensionConfigResponse(jsonBytes []byte) ([]byte, error) {
 	var data []byte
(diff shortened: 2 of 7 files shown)


source/plugins/go/src/extension/extension_linux.go
```diff
@@ -17,8 +19,11 @@ func getExtensionConfigResponse(jsonBytes []byte) ([]byte, error) {
...
 	fs := &FluentSocket{}
 	fs.sockAddress = "/var/run/mdsd-ci/default_fluent.socket"
+	genevaLogsIntegrationEnabled := strings.TrimSpace(strings.ToLower(os.Getenv("GENEVA_LOGS_INTEGRATION")))
 	if containerType != "" && strings.Compare(strings.ToLower(containerType), "prometheussidecar") == 0 {
 		fs.sockAddress = fmt.Sprintf("/var/run/mdsd-%s/default_fluent.socket", containerType)
+	} else if genevaLogsIntegrationEnabled != "" && strings.Compare(strings.ToLower(genevaLogsIntegrationEnabled), "true") == 0 {
+		fs.sockAddress = "/var/run/mdsd-PrometheusSidecar/default_fluent.socket"
 	}
 	responseBytes, err := FluentSocketWriter.writeAndRead(fs, data)
 	defer FluentSocketWriter.disconnect(fs)
```
jatakiajanvi12, line 30: Hey @ganga1980  do you want to remove the additional log message you mentioned here?

source/plugins/go/src/extension/extension_linux.go
```diff
@@ -17,8 +19,11 @@ func getExtensionConfigResponse(jsonBytes []byte) ([]byte, error) {
 
 	fs := &FluentSocket{}
 	fs.sockAddress = "/var/run/mdsd-ci/default_fluent.socket"
+	genevaLogsIntegrationEnabled := strings.TrimSpace(strings.ToLower(os.Getenv("GENEVA_LOGS_INTEGRATION")))
 	if containerType != "" && strings.Compare(strings.ToLower(containerType), "prometheussidecar") == 0 {
 		fs.sockAddress = fmt.Sprintf("/var/run/mdsd-%s/default_fluent.socket", containerType)
+	} else if genevaLogsIntegrationEnabled != "" && strings.Compare(strings.ToLower(genevaLogsIntegrationEnabled), "true") == 0 {
+		fs.sockAddress = "/var/run/mdsd-PrometheusSidecar/default_fluent.socket"
```
pfrcks: can we make this a const?

kubernetes/linux/main.sh
```diff
@@ -825,6 +825,10 @@ if [ "${CONTAINER_TYPE}" != "PrometheusSidecar" ] && isGenevaMode; then
     echo "export MDSD_FLUENT_SOCKET_PORT=$MDSD_FLUENT_SOCKET_PORT" >> ~/.bashrc
     export SSL_CERT_FILE="/etc/pki/tls/certs/ca-bundle.crt"
     echo "export SSL_CERT_FILE=$SSL_CERT_FILE" >> ~/.bashrc
+    if [ "${USING_AAD_MSI_AUTH}" == "true" ]; then
```
pfrcks, line 828: looks like both this parent if and else block have this check?

source/plugins/go/src/extension/extension_linux.go
jatakiajanvi12: Doesn't the above if also make sure the address is mdsd-prometheussidecar? Since we are checking that the containerType is prometheussidecar? We can combine the if and else right?

kubernetes/linux/main.sh
ganga1980: > looks like both this parent if and else block have this check?

No, parent if doesnt have this check. This is for geneva path only.

source/plugins/go/src/extension/extension_linux.go
ganga1980: > can we make this a const?

Fixed.

source/plugins/go/src/extension/extension_linux.go
ganga1980: > Hey @ganga1980 do you want to remove the additional log message you mentioned here?

Done.

source/plugins/go/src/extension/extension_linux.go
ganga1980: > Doesn't the above if also make sure the address is mdsd-prometheussidecar? Since we are checking that the containerType is prometheussidecar? We can combine the if and else right?

done.

jatakiajanvi12 (approved):

PR #1175, opened 2024-02-23
update v2 schema configmap

pfrcks:


build/common/installer/scripts/tomlparser.rb
```diff
@@ -158,11 +158,6 @@ def populateSettingValuesFromConfigMap(parsedConfig)
         @logEnableMultiline = parsedConfig[:log_collection_settings][:enable_multiline_logs][:enabled]
         puts "config::Using config map setting for multiline logging"
 
-        if @containerLogSchemaVersion.strip.casecmp("v2") != 0
-          puts "config:: WARN: container logs V2 is disabled and is required for multiline logging. Disabling multiline logging"
-          @logEnableMultiline = "false"
```
jatakiajanvi12, line 163: Don't we still want to do this if someone has multline enabled but they changed configmap to use v1?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -42,11 +42,11 @@ data:
...
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
-       #[log_collection_settings.schema]
+       [log_collection_settings.schema]
           # In the absence of this configmap, default value for containerlog_schema_version is "v1"
           # Supported values for this setting are "v1","v2"
           # See documentation at https://aka.ms/ContainerLogv2 for benefits of v2 schema over v1 schema before opting for "v2" schema
-          # containerlog_schema_version = "v2"
+          containerlog_schema_version = "v2"
        #[log_collection_settings.enable_multiline_logs]
```
jatakiajanvi12, line 50: Should we also mention here that v2 is needed for multiline?

pfrcks: our public docs already mention this and there are v2 usecases outside of multiline.

build/common/installer/scripts/tomlparser.rb
pfrcks: any new onboarding is v2 and therefore checking is NOOP. however for existing v1 customers yes they can in a way enable multiline with this change however that is unsupported and not publicized. 

we need to remove this check because someone who onboards via UX now has v2 enabled via DCR and assuming they then enable multiline thinking v2 is already enabled it wont work

kubernetes/container-azm-ms-agentconfig.yaml
jatakiajanvi12: No I meant should we mention in multiline that v2 is needed ?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -42,11 +42,11 @@ data:
           # When the setting is set to false, only the kube events with !normal event type will be collected
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
-       #[log_collection_settings.schema]
+       [log_collection_settings.schema]
           # In the absence of this configmap, default value for containerlog_schema_version is "v1"
           # Supported values for this setting are "v1","v2"
           # See documentation at https://aka.ms/ContainerLogv2 for benefits of v2 schema over v1 schema before opting for "v2" schema
```
ganga1980: since v2 will be default with this change in config map, this statement can be removed "before opting for "v2" schema", right?

kubernetes/container-azm-ms-agentconfig.yaml
```diff
@@ -42,11 +42,11 @@ data:
           # When the setting is set to false, only the kube events with !normal event type will be collected
           enabled = false
           # When this is enabled (enabled = true), all kube events including normal events will be collected
-       #[log_collection_settings.schema]
+       [log_collection_settings.schema]
           # In the absence of this configmap, default value for containerlog_schema_version is "v1"
```
ganga1980: we should rephrase this statement from 
 In the absence of this configmap, default value for containerlog_schema_version is "v1"
To
 In the absence of this configmap, default value for containerlog_schema_version is "v1" if "v2" not enabled through onboarding

ganga1980 (approved):

kubernetes/container-azm-ms-agentconfig.yaml
pfrcks: added to multiline section

PR #1183, opened 2024-02-29
update mdsd for aarch64 patch

pfrcks:
- Update mdsd from 1.29.4 to 1.29.7 to address aarch64 crashes

### DIFF
--- a/kubernetes/linux/setup.sh
+++ b/kubernetes/linux/setup.sh
@@ -37,7 +37,7 @@ mv /usr/lib/ruby/gems/3.1.0/specifications/default/uri-0.11.0.gemspec /usr/lib/r
 gem uninstall time --version 0.2.0
 gem uninstall uri --version 0.11.0
 
-sudo tdnf install -y azure-mdsd-1.29.4
+sudo tdnf install -y azure-mdsd-1.29.7
 cp -f $TMPDIR/mdsd.xml /etc/mdsd.d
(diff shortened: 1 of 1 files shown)


kubernetes/linux/setup.sh
```diff
@@ -37,7 +37,7 @@ mv /usr/lib/ruby/gems/3.1.0/specifications/default/uri-0.11.0.gemspec /usr/lib/r
 gem uninstall time --version 0.2.0
 gem uninstall uri --version 0.11.0
 
-sudo tdnf install -y azure-mdsd-1.29.4
+sudo tdnf install -y azure-mdsd-1.29.7
```
ganga1980, line 40: 1. can you please double check this version works on both ARM64 and x64 nodes?
2.  Lets ask the MDSD team to have the release notes for this version so we and other team can refer the release notes for this. In the past, AKS team indicated that, we were using not released version by looking at MDSD release notes.

pfrcks: 1) confirmed with Aaron that it's for both
2) will be live in an hour or so

jatakiajanvi12 (commented):
Do we need to test for cpu/memory usage  for this new version? Also make sure there isn't any other regression?

pfrcks:
compares 3.1.17 with this PR. later values are for this PR changes
legacy
legacy_cpu
legacy_mem

msi
msi_cpu
msi_mem

ganga1980 (approved):

PR #1200, opened 2024-03-11
fix bugs in fluent-bit input plugins

pfrcks:
* [`source/plugins/go/input/lib/cadvisor.go`](diffhunk://#diff-0713d5f8005c34704ac1b85fcd20fca2ebf1df47a0decf8690b8953e3e0929a8L528-R564): Modified `getContainerCpuMetricItemRate` function to add missing cpu metric for windows
* [`source/plugins/go/input/lib/cadvisor.go`](diffhunk://#diff-0713d5f8005c34704ac1b85fcd20fca2ebf1df47a0decf8690b8953e3e0929a8R901): Changed `getContainerStartTimeMetricItems` fix bug by parsing the string as unix epoch time. [[1]](diffhunk://#diff-0713d5f8005c34704ac1b85fcd20fca2ebf1df47a0decf8690b8953e3e0929a8R901) [[2]](diffhunk://#diff-0713d5f8005c34704ac1b85fcd20fca2ebf1df47a0decf8690b8953e3e0929a8L876-R911)


source/plugins/go/input/lib/cadvisor.go
```diff
@@ -1207,3 +1241,30 @@ func getPersistentVolumeMetrics(metricInfo map[string]interface{}, hostName, met
 
 	return metricItems
 }
+
+func ResetWinContainerIdCache() {
+	winContainerIdCache = make(map[string]bool)
```
ganga1980: This can cause memory growth until gc kicks in,  instead of creating new map object every time, rather we should delete the entries in the map.

 for key := range winContainerIdCache {
        delete(winContainerIdCache , key)
    }

source/plugins/go/input/perf/perf.go
```diff
@@ -125,6 +126,14 @@ func (p perfPlugin) Collect(ctx context.Context, ch chan<- plugin.Message) error
 				}
 				lib.SendTelemetry("Perf", telemetryProperties)
 			}
+
+			cleanupTimeDifference := int(math.Abs(float64(time.Now().Unix() - cleanupRoutineTimeTracker)))
+			cleanupTimeDifferenceInMinutes := cleanupTimeDifference / 60
```
ganga1980, line 131: Please validate this change at least 15 to 30 mins to validate the clean up routine logic works as expected.

ganga1980 (approved):

PR #1208, opened 2024-03-20
Enable disable addon test pipeline

jatakiajanvi12:
This pull request includes changes to the `.pipelines/daily_addons_enablement_test.yaml` file. The changes primarily involve setting up a new scheduled job that checks the status of the OMSAgent addon on an AKS cluster and performs various operations based on its status.

Here are the most important changes:

* Added a new scheduled job that runs daily at midnight on the `ci_prod` branch. The job uses the Azure CLI to interact with an AKS cluster.
* Defined several variables, including the Azure subscription ID, the resource group name, the AKS cluster name, and the Log Analytics workspace ID.
* Added a script that checks if the OMSAgent addon is enabled on the AKS cluster. If it is, the script disables and then re-enables the addon, waits for logs to start flowing, and then checks if there is data in the Log Analytics workspace. If the addon is not enabled or if there is no data in the workspace, the script exits with an error.

### DIFF
--- a/.pipelines/daily_addons_enablement_test.yaml
+++ b/.pipelines/daily_addons_enablement_test.yaml
@@ -1,0 +1,64 @@
+schedules:
+- cron: '0 0 * * *'
+  displayName: Daily midnight test
+  branches:
+    include:
+    - ci_prod
+jobs:
+    - job: test
+      variables:
+        armServiceConnectionName: 'ci-1es-acr-connection'
+        subscription: '9b96ebbd-c57a-42d1-bbe9-b69296e4c7fb'
+      pool:
+        name: Azure-Pipelines-CI-Test-EO
+      steps:
+          - task: AzureCLI@2
+            inputs:
+              azureSubscription: '${{ variables.armServiceConnectionName }}'
+              scriptType: 'bash'
+              scriptLocation: 'inlineScript'
+              inlineScript: |
+                az --version
+                az account show
+                az account set -s ${{ variables.subscription }}
+                
+                # Check if the omsagent addon is enabled
+                addonEnabled=$(az aks show --resource-group $(RESOURCEGROUP) --name $(AKSCLUSTERNAME) --query addonProfiles.omsagent.enabled -o tsv)
+                
+                if [ "$addonEnabled" == "true" ]; then
+                    echo "OMSAgent addon is enabled on the AKS cluster."
+                    disable=$(az aks disable-addons --addons monitoring --name $(AKSCLUSTERNAME) --resource-group $(RESOURCEGROUP))
+                    omsagentEnabled=$(echo $disable | jq '.addonProfiles.omsagent.enabled')
+                    if [ $omsagentEnabled == "true" ]; then
+                      echo "Omsagent is not disabled."
+                      exit -1
+                    fi
+                    
+                    echo "Re-enabling the OMSAgent addon..."
+                    enable=$(az aks enable-addons --addons monitoring --name $(AKSCLUSTERNAME) --resource-group $(RESOURCEGROUP) --workspace-resource-id $(LOGANALYTICSWORKSPACE))
+                    omsagentEnabled=$(echo $enable | jq '.addonProfiles.omsagent.enabled')
+                    if [ $omsagentEnabled == "false" ]; then
+                      echo "Omsagent is not enabled"
+                      exit -1
+                    fi
+                    omsagentWorkspace=$(az aks show --resource-group $(RESOURCEGROUP) --name $(AKSCLUSTERNAME) --query addonProfiles.omsagent.config.logAnalyticsWorkspaceResourceID -o tsv)
+                    if [ "$omsagentWorkspace" != "$(LOGANALYTICSWORKSPACE)" ]; then
+                      echo "Omsagent is not enabled with the correct log analytics workspace - $omsagentWorkspace"
+                      exit -1
+                    fi
+                    
+                    echo "Sleeping for everything to get set up and logs start flowing"
+                    sleep $(SLEEPDURATION)
+                    # Check if there is data in the Log Analytics workspace
+                    dataCount=$(az monitor log-analytics query --workspace $(WORKSPACEID) --analytics-query "ContainerInventory | where TimeGenerated > ago(5m) and _ResourceId contains 'daily_enable_disable_test' | count")
+                    countValue=$(echo "$dataCount" | jq -r '.[0].Count')
+                    if [ $countValue -gt 0 ]; then
+                      echo "Data is present in the Log Analytics workspace."
+                    else
+                      echo "No data is present in the Log Analytics workspace."
+                      exit -1
+                    fi
+                else
+                    echo "OMSAgent addon is not enabled on the AKS cluster."
+                    exit -1
+                fi


pfrcks:
what is the intended use case for this pipeline? does it check enablement via az cli? also since it is supposed to be nightly will it be hooked up to some alerts/dashboards?

jatakiajanvi12:
> what is the intended use case for this pipeline? does it check enablement via az cli? also since it is supposed to be nightly will it be hooked up to some alerts/dashboards?

Yes it is checking the addon enablement using az cli. checking the addon is getting enabled or not. I was checking the alerts open but not able to create a team alert or something as I dont have permissions. I have created a personal notification. I will check more on this

.pipelines/daily_addons_enablement_test.yaml
```diff
@@ -0,0 +1,56 @@
...
+              azureSubscription: '${{ variables.armServiceConnectionName }}'
+              scriptType: 'bash'
+              scriptLocation: 'inlineScript'
+              inlineScript: |
+                az --version
+                az account show
+                az account set -s ${{ variables.subscription }}
+                
+                # Define the resource group and AKS cluster name
+                resourceGroup="ci-logs-prod-aks"
```
ganga1980: use the pipeline variables for all the internal specific aksclusterName, logAnalyticsworkspace etc. and instead of adding these in to github.

.pipelines/daily_addons_enablement_test.yaml
```diff
@@ -0,0 +1,56 @@
...
+                logAnalyticsWorkspace="/subscriptions/9b96ebbd-c57a-42d1-bbe9-b69296e4c7fb/resourcegroups/ci-logs-prod-aks/providers/microsoft.operationalinsights/workspaces/daily-enable-disable-test-la"
+                workspaceId="0b525f9a-755c-4a1a-b6fe-f96a23ea8a4e"
+
+                # Check if the omsagent addon is enabled
+                addonEnabled=$(az aks show --resource-group $resourceGroup --name $aksClusterName --query addonProfiles.omsagent.enabled -o tsv)
+                
+                if [ "$addonEnabled" == "true" ]; then
+                    echo "OMSAgent addon is enabled on the AKS cluster."
+                    az aks disable-addons --addons monitoring --name $aksClusterName --resource-group $resourceGroup
+                    sleep 10
```
ganga1980: 10s delay between disable and enable not enough as the AKS does the reconciler and it could take more time .

.pipelines/daily_addons_enablement_test.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+                    echo "Sleeping 6 minutes for everything to get set up and logs start flowing"
+                    sleep 360
+                    # Check if there is data in the Log Analytics workspace
+                    dataCount=$(az monitor log-analytics query --workspace $workspaceId --analytics-query "ContainerInventory | where TimeGenerated > ago(5m) and _ResourceId contains 'daily_enable_disable_test' | count")
+                    countValue=$(echo "$dataCount" | jq -r '.[0].Count')
+                    if [ $countValue -gt 0 ]; then
+                      echo "Data is present in the Log Analytics workspace."
+                    else
+                      echo "No data is present in the Log Analytics workspace."
```
ganga1980, line 58: how will get notified if this failed? do we need to look at the pipeline everyday?

.pipelines/daily_addons_enablement_test.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+                # Check if the omsagent addon is enabled
+                addonEnabled=$(az aks show --resource-group $resourceGroup --name $aksClusterName --query addonProfiles.omsagent.enabled -o tsv)
+                
+                if [ "$addonEnabled" == "true" ]; then
+                    echo "OMSAgent addon is enabled on the AKS cluster."
+                    az aks disable-addons --addons monitoring --name $aksClusterName --resource-group $resourceGroup
+                    sleep 10
+                    echo "Re-enabling the OMSAgent addon..."
+                    az aks enable-addons --addons monitoring --name $aksClusterName --resource-group $resourceGroup --workspace-resource-id $logAnalyticsWorkspace
```
ganga1980: we should also validate addonconfig after enable and disablement.

.pipelines/daily_addons_enablement_test.yaml
```diff
@@ -0,0 +1,56 @@
...
+                    echo "OMSAgent addon has been re-enabled."
+
+                    echo "Sleeping 6 minutes for everything to get set up and logs start flowing"
+                    sleep 360
```
ganga1980: make these intervals as pipeline variables so that these can tuned without changing the script.

.pipelines/daily_addons_enablement_test.yaml
jatakiajanvi12: This is actually not needed....The disable is completed only after the cluster is in running state

.pipelines/daily_addons_enablement_test.yaml
```diff
@@ -0,0 +1,64 @@
...
+                      exit -1
+                    fi
+                    omsagentWorkspace=$(az aks show --resource-group $(RESOURCEGROUP) --name $(AKSCLUSTERNAME) --query addonProfiles.omsagent.config.logAnalyticsWorkspaceResourceID -o tsv)
+                    if [ "$omsagentWorkspace" != "$(LOGANALYTICSWORKSPACE)" ]; then
+                      echo "Omsagent is not enabled with the correct log analytics workspace - $omsagentWorkspace"
+                      exit -1
+                    fi
+                    
+                    echo "Sleeping for everything to get set up and logs start flowing"
+                    sleep $(SLEEPDURATION)
```
ganga1980, line 51: SLEEPDURATION has to be greater than 5m which is query interval and otherwise we will be querying the data before disable.  whats the default value of SLEEPDURATION  pipeline variable?

.pipelines/daily_addons_enablement_test.yaml
ganga1980: as discussed, add the automated alert to track this.

.pipelines/daily_addons_enablement_test.yaml
jatakiajanvi12: It is greater than 5 min. It is 6mi

.pipelines/daily_addons_enablement_test.yaml
jatakiajanvi12: Yes will add that in the other repo

ganga1980 (dismissed):

pfrcks:
> > what is the intended use case for this pipeline? does it check enablement via az cli? also since it is supposed to be nightly will it be hooked up to some alerts/dashboards?
> 
> Yes it is checking the addon enablement using az cli. checking the addon is getting enabled or not. I was checking the alerts open but not able to create a team alert or something as I dont have permissions. I have created a personal notification. I will check more on this

Makes sense. Since we are just checking enablement disablement then this pipeline should only run once daily at midnight. Is that or something similar the plan?

Also to avoid flaky errors might be worth to have some workloads running on this pipeline's test cluster if there isnt already

ganga1980 (approved):

PR #1211, opened 2024-03-21
Improve performance of ruby get_hostname

roharon:
Hi Microsoft Team!
I improved some method located in `source/plugins/ruby/oms_common.rb` that get_hostname used.

In ruby, `Regexp#=~` returns the integer index of the first match in string, otherwise nil.
`Regex#match?` returns true or false to indicate whether the regex expression is matched or not. `Regex#match?` executes a regexp match without creating a back reference object to reduce object allocation and changing $~ to reduce object allocation.

In light of this, I think this pr will be helpful for improving ruby plugin performance


ganga1980:
@roharon , thanks for making this PR. This file not being used and  we would need to clean up, can you please update the PR with the change to delete the file?

pfrcks:
hi @roharon thanks for the PR.

we plan on removing this file from our codebase soon and therefore will be closing this PR.

PR #1216, opened 2024-03-26
Add telemetry for KubernetesMetadata size

wanlonghenry:
This PR majorly add telemetry for KubernetesMetadata size

### DIFF
--- a/source/plugins/go/src/oms.go
+++ b/source/plugins/go/src/oms.go
@@ -2010,7 +2010,7 @@ func GetContainerIDK8sNamespacePodNameFromFileName(filename string) (string, str
 // For sample pod name kube-proxy-dgcx7 it returns kube and kube-proxy.
 // For sample pod name coredns-77d8fb66dd-hsgbb returns coredns-77d8fb66dd and coredns
 // The returned values are matched against configmap inputs eg kube-system:coredns and kube-system:kube-proxy
-func GetControllerNameFromK8sPodName (podName string) (string, string) {
+func GetControllerNameFromK8sPodName(podName string) (string, string) {
 	// clear cache if it exceeds the cache size
 	if len(PodNameToControllerNameMap) > PodNameToControllerNameMapCacheSize {
 		Log("Clearing PodNameToControllerNameMap cache")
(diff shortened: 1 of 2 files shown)


source/plugins/go/src/oms.go
```diff
@@ -1603,6 +1603,9 @@ func PostDataHelper(tailPluginRecords []map[interface{}]interface{}) int {
 		var msgPackEntry MsgPackEntry
 
 		FlushedRecordsSize += float64(len(stringMap["LogEntry"]))
+		if ContainerLogSchemaV2 == true {
```
ganga1980: 1. Instead of ContainerLogSchemaV2 and we should check for KubernetesMetadataEnabled , rght?
2. Isnt that KubernetesMetadataEnabled  only for ContainerLogSchemaV2? if yes, then we can add check for KubernetesMetadataEnabled  otherwise we should add check for both  - 

if KubernetesMetadataEnabled  &&  ContainerLogSchemaV2 {
FlushedMetadataSize += float64(len(stringMap["KubernetesMetadata"]))
}

source/plugins/go/src/telemetry.go
```diff
@@ -262,11 +267,16 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
...
 				TelemetryClient.Track(flushRateMetric)
+
 				logRateMetric := appinsights.NewMetricTelemetry(metricNameAvgLogGenerationRate, logRate)
 				logSizeMetric := appinsights.NewMetricTelemetry(metricNameLogSize, logSizeRate)
 				TelemetryClient.Track(logRateMetric)
 				Log("Log Size Rate: %f\n", logSizeRate)
 				TelemetryClient.Track(logSizeMetric)
+
+				metadataSizeMetric := appinsights.NewMetricTelemetry(metricNameMetadataSize, metadataSizeRate)
+				TelemetryClient.Track(metadataSizeMetric)
```
ganga1980: we should emit this metric only when KubernetesMetadataEnabled .

if KubernetesMetadataEnabled  {
    metadataSizeMetric := appinsights.NewMetricTelemetry(metricNameMetadataSize, metadataSizeRate)
				TelemetryClient.Track(metadataSizeMetric)
}

source/plugins/go/src/oms.go
wanlonghenry: changed to collect when KubernetesMetadataEnabled

ganga1980 (approved):

PR #1224, opened 2024-04-10
Add AI onboarding ARM templates

jatakiajanvi12:
This pull request introduces two new JSON files for Azure Resource Manager (ARM) templates, which are used to automate deployments in Microsoft Azure. The first file, `existingClusterOnboarding.json`, defines the deployment template for an existing Azure Kubernetes Service (AKS) cluster. The second file, `existingClusterParam.json`, provides parameter values for the deployment template.

Here are the key changes:

* [`AIOnboardingARMTemplates/existingClusterOnboarding.json`](diffhunk://#diff-5ae6a92df76f06138ee7303ac1749a16992396c6be31d302dcce240faf06f63cR1-R89): This new file is an ARM template for onboarding an existing AKS cluster. It includes parameters for the AKS resource ID and location, ports for OpenTelemetry logs and metrics, and tags for AKS, ContainerInsights, and DataCollectionRule resources. The template also includes a deployment for an AKS monitoring add-on, which enables auto-instrumentation and OpenTelemetry logs and metrics.

* [`AIOnboardingARMTemplates/existingClusterParam.json`](diffhunk://#diff-8abce7b09c25189ac9ac6bee36c93c6f2c05ff2f35cf08f4b96357a5502cea1fR1-R25): This new file provides parameter values for the `existingClusterOnboarding.json` template. It includes placeholders for the AKS resource ID and location, OpenTelemetry logs and metrics ports, and resource tags.

### DIFF
--- a/AIOnboardingARMTemplates/existingClusterOnboarding.json
+++ b/AIOnboardingARMTemplates/existingClusterOnboarding.json
@@ -11,7 +11,7 @@
         "aksResourceLocation": {
             "type": "string",
             "metadata": {
-              "description": "Location of the AKS resource e.g. \"East US\""
+              "description": "Location of the AKS resource e.g. \"eastus\""
             }
         },
         "openTelemetryLogsPort": {
--- a/AIOnboardingARMTemplates/existingClusterOnboarding.json
+++ b/AIOnboardingARMTemplates/existingClusterOnboarding.json
@@ -1,0 +1,89 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "aksResourceId": {
+            "type": "string",
+            "metadata": {
+                "description": "AKS Cluster Resource ID"
+            }
+        },
+        "aksResourceLocation": {
+            "type": "string",
+            "metadata": {
+              "description": "Location of the AKS resource e.g. \"East US\""
+            }
+        },
+        "openTelemetryLogsPort": {
+            "type": "int",
+            "metadata": {
+              "description": "Port to use for OpenTelemetry Logs"
+            },
+            "defaultValue": "28331"
+        },
+        "openTelemetryMetricsPort": {
+            "type": "int",
+            "metadata": {
+              "description": "Port to use for OpenTelemetry Metrics"
+            },
+            "defaultValue": "28333"
+        },
+        "resourceTagValues": {
+            "type": "object",
+            "metadata": {
+              "description": "Existing or new tags to use on AKS, ContainerInsights and DataCollectionRule Resources"
+            }
+        }
+    },
+    "variables": {
+        "clusterSubscriptionId": "[split(parameters('aksResourceId'),'/')[2]]",
+        "clusterResourceGroup": "[split(parameters('aksResourceId'),'/')[4]]",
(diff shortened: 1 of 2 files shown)


AIOnboardingARMTemplates/existingClusterOnboarding.json
```diff
@@ -0,0 +1,89 @@
...
+        "aksResourceId": {
+            "type": "string",
+            "metadata": {
+                "description": "AKS Cluster Resource ID"
+            }
+        },
+        "aksResourceLocation": {
+            "type": "string",
+            "metadata": {
+              "description": "Location of the AKS resource e.g. \"East US\""
```
tokaplan: Is this the format we are following? Not "eastus"?

AIOnboardingARMTemplates/existingClusterOnboarding.json
```diff
@@ -0,0 +1,89 @@
...
+                "contentVersion": "1.0.0.0",
+                "parameters": {},
+                "variables": {},
+                "resources": [
+                    {
+                        "name": "[variables('clusterName')]",
+                        "type": "Microsoft.ContainerService/managedClusters",
+                        "location": "[parameters('aksResourceLocation')]",
+                        "tags": "[parameters('resourceTagValues')]",
+                        "apiVersion": "2024-02-02-preview",
```
tokaplan, line 64: Is this the new _api-version_ that has the latest changes?

AIOnboardingARMTemplates/existingClusterOnboarding.json
```diff
@@ -0,0 +1,89 @@
...
+            "defaultValue": "28331"
+        },
+        "openTelemetryMetricsPort": {
+            "type": "int",
+            "metadata": {
+              "description": "Port to use for OpenTelemetry Metrics"
+            },
+            "defaultValue": "28333"
+        },
+        "resourceTagValues": {
```
tokaplan, line 31: What is the usage for these tags?

AIOnboardingARMTemplates/existingClusterOnboarding.json
jatakiajanvi12: yes this is the new version

AIOnboardingARMTemplates/existingClusterOnboarding.json
jatakiajanvi12: I think it is just used by customers to tag their deployments and resources. It is same as Azure tags: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources

AIOnboardingARMTemplates/existingClusterOnboarding.json
jatakiajanvi12: yes makes sense - my bad. Let me make this update

tokaplan: @jatakiajanvi12 let's make sure it's in line with what ARM typically uses. In my experience, templates usually follow the "eastus" format, but my experience is limited.

tokaplan (approved):

AIOnboardingARMTemplates/existingClusterOnboarding.json
jatakiajanvi12: yes you are right. Our current ARM templates also follow that and I tested with "eastus" only but there was a description error

PR #1228, opened 2024-04-16
update dcr optimization error messages

pfrcks:
This pull request includes updates to the error messages in the `dcr-config-parser.rb` script. The changes make the error messages more informative, providing additional context for troubleshooting. Specifically, the error message for when no JSON file is found now suggests checking if mdsd is running in MSI mode. Similarly, the error message for when the required key 'dataSources' is missing from the JSON structure now suggests checking if DCR is valid.

### DIFF
--- a/build/linux/installer/scripts/dcr-config-parser.rb
+++ b/build/linux/installer/scripts/dcr-config-parser.rb
@@ -33,12 +33,12 @@ if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.cas
     end
 
     # Raise an error if no JSON file is found
-    raise 'No JSON file found in the specified directory' unless file_path
+    raise 'No JSON file found in the specified directory. Check if mdsd is running in MSI mode' unless file_path
 
     file_contents = File.read(file_path)
     data = JSON.parse(file_contents)
 
-    raise 'Invalid JSON structure: Missing required keys' unless data.is_a?(Hash) && data.key?('dataSources')
+    raise 'Invalid JSON structure: Missing required key: dataSources. Check if DCR is valid' unless data.is_a?(Hash) && data.key?('dataSources')
 
     # Extract the stream values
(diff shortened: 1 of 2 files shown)


jatakiajanvi12 (dismissed):

build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -33,12 +33,12 @@
     end
 
     # Raise an error if no JSON file is found
-    raise 'No JSON file found in the specified directory' unless file_path
+    raise 'No JSON file found in the specified directory. Check if mdsd is running in MSI mode' unless file_path
```
ganga1980, line 37: If the Geneva Integration enabled, we shouldnt get into this error since there is no DCR. Do you know why this error triggered  when the geneva integration enabled?

pfrcks: yes, we check whether geneva integration is enabled and return if it is: https://github.com/microsoft/Docker-Provider/blob/05e094df2470e8a6dcedd73b220c4fd8b7b27131/build/linux/installer/scripts/dcr-config-parser.rb#L18

it looks like for some reason cluster was in geneva mode but GENEVA_LOGS_INTEGRATION was not set

ganga1980: Ok, lets investigate and also, we should also add similar check on GENEVA_LOGS_INTEGRATION_SERVICE_MODE as well. return if ENV['GENEVA_LOGS_INTEGRATION_SERVICE_MODE ']&.strip&.casecmp?('true')

ganga1980: @pfrcks , do you know why GENEVA_LOGS_INTEGRATION flag not being set even though geneva enabled? can you please check Nont cluster and see what's the issue here?

ganga1980 (approved):

PR #1229, opened 2024-04-17
update fluent-bit to 2.2.2 in linux

pfrcks:
The change updates the version of `fluent-bit` that is installed from `2.1.10` to `2.2.2`.


pfrcks:
legacy cpu ds
Screenshot 2024-04-16 200246


legacy memory ds
Screenshot 2024-04-16 200213


rs cpu no visible change
rs memory 2.2.2 shows 15MB increase.

jatakiajanvi12 (commented):
Should we also compare for MSI mode?

pfrcks:
> Should we also compare for MSI mode?

for msi the values are very similar, ds memory is slightly worse 5-10MB

in order: fb update vs 3.1.19
image

jatakiajanvi12 (dismissed):

kubernetes/linux/setup.sh
```diff
@@ -70,7 +70,7 @@ docker_cimprov_version=$(sudo tdnf list installed | grep docker-cimprov | awk '{
 echo "DOCKER_CIMPROV_VERSION=$docker_cimprov_version" >> packages_version.txt
 
 #install fluent-bit
-sudo tdnf install fluent-bit-2.1.10 -y
+sudo tdnf install fluent-bit-2.2.2 -y
```
ganga1980, line 73: can you please double check these (1) verify if there are any errors/warnings in fluent-bit.log (2) did you verify with some workloads and see if fluent-bit able to detect the new log files and log updates (3) can you please review the release notes and issues in github and see if there are any issues reported on this version

ganga1980 (approved):

PR #1237, opened 2024-04-29
Add scan tools to the build pipeline

jatakiajanvi12:
Added the following tools: 
- credscan - can index and scan for credentials & other sensitive content in source code, other data sources such as logs, telemetry, and data stores, and documents.
- binskim - validates compiler/linker settings and other security-relevant binary characteristics. https://github.com/Microsoft/binskim
- policheck - scans files and directories for sensitive geopolitical terms, profanity, and other sensitive terms
- armory - scan for common vulnerabilities in ARM template json files.
- antimalware: allows users to run an Anti-Malware Scan ion Windows agents, using Windows Defender


.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -81,6 +81,9 @@ jobs:
       echo "##vso[task.setvariable variable=windowsAMAUrl;isOutput=true]$windowsAMAUrl"
     name: setup
 
+  - task: CredScan@3
```
pfrcks, line 84: shouldn't this be run before "git push" since our repo is public? if it runs after git push the code is already publicly available

pfrcks: we might need to setup credscam as a github webhook

jatakiajanvi12: yeah true. I added credscan that is built my msft. we will have to find something which github uses as well

pfrcks (approved):

PR #1240, opened 2024-05-01
Fix Certificate housekeeper job to fix issues with secrets

ramanujbhattacharjee93:


appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -174,6 +174,7 @@ export class CertificateManager {
             }
         } catch (error) {
             logger.error(error, operationId, this.requestMetadata);
+            logger.error(JSON.stringify(error), operationId, this.requestMetadata);
```
tokaplan, line 176: Surely we don't need both?

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -232,6 +233,53 @@ export class CertificateManager {
...
+    private static IsValidCertificate(operationId: string, mwhcCaBundle: string, webhookCertData: WebhookCertData, clusterArmId: string, clusterArmRegion: string): boolean {
+        try {
+            forge.pki.certificateFromPem(mwhcCaBundle);
+            forge.pki.certificateFromPem(webhookCertData.caCert);
+            forge.pki.certificateFromPem(webhookCertData.tlsCert);
+            forge.pki.privateKeyFromPem(webhookCertData.tlsKey);
+            return true;
+        } catch (error) {
+            logger.error('Error occured while trying to validate certificates!', operationId, this.requestMetadata);
+            logger.error(JSON.stringify(error), operationId, this.requestMetadata);
```
tokaplan, line 244: We want a single error trace, why split it into 2?

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -232,6 +233,53 @@ export class CertificateManager {
...
+            logger.SendEvent("CertificateValidationFailed", operationId, null, clusterArmId, clusterArmRegion, true, error);
+            return false;
+        }
+    }
+
+    private static async CheckCertificateJobStatus(kubeConfig: k8s.KubeConfig, operationId: string, clusterArmId: string, clusterArmRegion: string): Promise<boolean> {
+        kubeConfig.loadFromDefault();
```
tokaplan: I thought we decided this wasn't needed?

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -232,6 +233,53 @@ export class CertificateManager {
...
+
+        const k8sApi = kubeConfig.makeApiClient(k8s.BatchV1Api);
+        const requestMetadata = this.requestMetadata;
+
+        async function checkJobStatus(namespace, jobName) : Promise<boolean> {
```
tokaplan: Why do we need a local function that is only invoked once immediately below?

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -318,58 +367,50 @@ export class CertificateManager {
...
-        const selector = "app-monitoring-webhook"
+    private static async RestartWebhookDeployment(operationId: string, kc: k8s.KubeConfig, clusterArmId: string, clusterArmRegion: string) {
         let name = null;
-
+    
         try {
-            const replicaSets: k8s.V1ReplicaSetList = (await k8sApi.listNamespacedReplicaSet(NamespaceName)).body;
+            const k8sApi = kc.makeApiClient(k8s.AppsV1Api);
+            const selector = "app-monitoring-webhook"
+            const deployments: k8s.V1DeploymentList = (await k8sApi.listNamespacedDeployment(NamespaceName)).body;
```
tokaplan: Let's remove the potential train wreck please.

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -318,58 +367,50 @@ export class CertificateManager {
...
 
-            const matchingReplicaSets: k8s.V1ReplicaSet[] = replicaSets.items.filter(rs => 
-                rs.spec.selector && rs.spec.selector.matchLabels && selector.localeCompare(rs.spec.selector.matchLabels.app) === 0
+            if (!deployments)
+            {
+                throw new Error(`No Deployments found in ${NamespaceName} namespace!`);
+            }
+    
+            const matchingDeployments: k8s.V1Deployment[] = deployments.items.filter(deployment => 
+                deployment.spec.selector && deployment.spec.selector.matchLabels && selector.localeCompare(deployment.spec.selector.matchLabels.app) === 0
```
tokaplan: We do know the webhook deployment is the only deployment with the same app label, correct?

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -318,58 +367,50 @@ export class CertificateManager {
...
             );
 
-            const replicaSet: k8s.V1ReplicaSet = matchingReplicaSets[0];
-            const annotations = replicaSet.spec.template.metadata.annotations || {};
+            if (matchingDeployments.length != 1) {
+                throw new Error(`Expected 1 Deployment with selector ${selector}, but found ${matchingDeployments.length}`);
+            }
+
```
tokaplan: Extra empty line?

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -318,58 +367,50 @@ export class CertificateManager {
...
-
-            logger.info(`Restarting ReplicaSet ${name}...`, operationId, this.requestMetadata);
-            logger.SendEvent("CertificateRestartingReplicaSet", operationId, null, clusterArmId, clusterArmRegion);
-            await k8sApi.replaceNamespacedReplicaSet(name, NamespaceName, replicaSet);
-            console.log(`Successfully restarted ReplicaSet ${name}`);
-            logger.SendEvent("CertificateRestartedReplicaSet", operationId, null, clusterArmId, clusterArmRegion);
-
+            deployment.spec.template.metadata.annotations = annotations;
+            name = deployment.metadata.name;
+
```
tokaplan: Random empty line and random tab

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -318,58 +367,50 @@ export class CertificateManager {
...
-
+            deployment.spec.template.metadata.annotations = annotations;
+            name = deployment.metadata.name;
+    
+            logger.info(`Restarting Deployment ${name}...`, operationId, this.requestMetadata);
+            logger.SendEvent("DeploymentRestarting", operationId, null, clusterArmId, clusterArmRegion);
+            await k8sApi.replaceNamespacedDeployment(name, NamespaceName, deployment);
+            console.log(`Successfully restarted Deployment ${name}`);
+            logger.SendEvent("DeploymentRestarted", operationId, null, clusterArmId, clusterArmRegion);
+
```
tokaplan: Random tab

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -318,58 +367,50 @@ export class CertificateManager {
...
+            logger.info(`Restarting Deployment ${name}...`, operationId, this.requestMetadata);
+            logger.SendEvent("DeploymentRestarting", operationId, null, clusterArmId, clusterArmRegion);
+            await k8sApi.replaceNamespacedDeployment(name, NamespaceName, deployment);
+            console.log(`Successfully restarted Deployment ${name}`);
+            logger.SendEvent("DeploymentRestarted", operationId, null, clusterArmId, clusterArmRegion);
+    
         } catch (err) {
-            logger.error(`Failed to get ReplicaSet ${name}: ${err}`, operationId, this.requestMetadata);
-            logger.SendEvent("CertificateRestartFailedReplicaSet", operationId, null, clusterArmId, clusterArmRegion, true, err);
+            logger.error(`Failed to get Deployment ${name}: ${err}`, operationId, this.requestMetadata);
```
tokaplan: Failed to get? Or failed to restart?

tokaplan (changes requested):

appmonitoring/ts/src/CertificateGenerator.ts
ramanujbhattacharjee93: This is definitely needed :)

appmonitoring/ts/src/CertificateGenerator.ts
ramanujbhattacharjee93: I dont think this comes out as 2? I need that distinct message, and not have it be muddled with the stacktrace

tokaplan: You call logger.error twice, how would it not come out as 2? You can do it this way if that's what you want, but the output will be just as muddled, except now there are two error messages that pertain to exactly the same thing.

appmonitoring/ts/src/CertificateGenerator.ts
tokaplan: I thought we decided we would only call `loadFromDefault` once immediately after initialization? This appears to be one of those extra calls that we don't need, correct?

appmonitoring/ts/src/CertificateGenerator.ts
ramanujbhattacharjee93: Just my preference really, makes the logic easier to read

appmonitoring/ts/src/CertificateGenerator.ts
ramanujbhattacharjee93: Yes

appmonitoring/ts/src/CertificateGenerator.ts
tokaplan: Let's not do it please unless there's an actual reason, it's weird and arbitrary. If you feel you need a separate method - just make it a normal method. If not - just have the content of the method directly in the body of the top level method.

tokaplan (changes requested):

appmonitoring/ts/src/CertificateGenerator.ts
ramanujbhattacharjee93: ah you mean only this line, somehow I thought you were asking if this entire method was needed :D

appmonitoring/ts/src/CertificateGenerator.ts
ramanujbhattacharjee93: wait thats not the line I thought you were talking about

tokaplan: I'm referring to the existence of the `checkJobStatus` local function.

tokaplan (changes requested):

appmonitoring/ts/src/CertificateGenerator.ts
tokaplan, line 1: Make sure the file and the class match when it comes to the name

tokaplan (changes requested):

appmonitoring/ts/src/CertificateGenerator.ts
```diff
@@ -232,6 +232,50 @@ export class CertificateManager {
...
+        const jobName = 'appmonitoring-cert-manager-hook-install';
+        const namespace = NamespaceName;
+
+        try {
+            const res = await k8sApi.readNamespacedJobStatus(jobName, namespace);
+            const jobStatus = res.body.status;
+            
+            if (jobStatus.conditions) {
+                for (const condition of jobStatus.conditions) {
+                    if (condition.type === 'Complete' && condition.status === 'True') {
```
tokaplan, line 273: For both these conditions, can we maybe use case-insensitive comparison? A little scary to have it like this...

tokaplan (approved):

PR #1241, opened 2024-05-01
Add life cycle metadata to container insights dev images

wanlonghenry:
This change adds life cycle metadata to the container insights dev images.

This adds metadata end-of-life annotation at the image build time so that it does not get flagged in downstream.
image
image

And it only adds metadata for dev images only: without telemetry tag indicated vs with telemetry tag
image


.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -238,6 +239,15 @@ jobs:
     displayName: "ORAS Push Artifacts in $(Build.ArtifactStagingDirectory)/linux/"
     condition: eq(variables.IS_PR, false)
 
+  - bash: |
+        set -euxo pipefail
+        oras attach ${{ variables.repoImageName }}:$(linuxImagetag) \
+            --artifact-type 'application/vnd.microsoft.artifact.lifecycle' \
+            --annotation "vnd.microsoft.artifact.lifecycle.end-of-life.date=$(date -u -d '-1 hour' +"%Y-%m-%dT%H:%M:%SZ")"
```
ganga1980, line 246: 1. why -1 hour, isnt that +1 hour? 
2.  after this timestamp, is image cant be pulled from MCR?
3. can we have this configurable through pipeline variable instead of hardcoding?

wanlonghenry: 1. Yes -1 means that it is already end of life as they are dev images
2. Yes it is just marked as end of life
3. it is doable but if the goal is to mark it EOL before push to mcr, I don't think it will require change, let me know your thoughts

ganga1980: From #1 and #2, response,  dev images cant be used since these are end of life as soon we push the images. is that right?

wanlonghenry: yes it is not recommended to use by customers and it won't be scanned for vulnerabilities

ganga1980 (approved):

pfrcks (approved):

PR #1246, opened 2024-05-09
Longw/signing payload drop

wanlonghenry:
This change migrate from using az acr import to use oras copy to avoid current limitation of az acr import as it is only copying the images without any artifacts. 

Please note that the oras copy will copy all artifacts including signing payload and lifecycle management.
image
without override flag:
image
with: 
image


deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
```diff
@@ -66,17 +64,27 @@ fi
...
+  echo "az logged in successfully"
 else
   echo "-e error failed to login to az with managed identity credentials"
   exit 1
-fi     
+fi
 
-echo "Pushing ${AGENT_IMAGE_FULL_PATH} to ${ACR_NAME} with force option set to ${AZ_ACR_IMPORT_FORCE}"
-az acr import --name $ACR_NAME --source $SOURCE_IMAGE_FULL_PATH --image $AGENT_IMAGE_FULL_PATH $AZ_ACR_IMPORT_FORCE
+az acr login -n containerinsights
```
ganga1980: this is same acr name coming as the value of ${ACR_NAME}, right? If yes, let use the ${ACR_NAME} and this way in future if we can change the ACR_NAME pipeline variable value and that should work.

deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
```diff
@@ -66,17 +64,27 @@ fi
...
+if [ "$OVERRIDE_TAG" == "true" ] || [ "$TAG_EXISTS_STATUS" -ne 0 ]; then
+  echo "Copying ${SOURCE_IMAGE_FULL_PATH} to ${ACR_NAME}/${AGENT_IMAGE_FULL_PATH}"
+  oras copy -r $SOURCE_IMAGE_FULL_PATH $ACR_NAME/$AGENT_IMAGE_FULL_PATH
+  if [ $? -eq 0 ]; then
+    echo "Retagged and pushed image and artifact successfully"
+  else
+    echo "-e error failed to retag and push image to destination ACR"
+    exit 1
+  fi
 fi
```
ganga1980: add new line

deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
```diff
@@ -35,7 +34,6 @@ echo "TAG_EXISTS_STATUS = $TAG_EXISTS_STATUS; OVERRIDE_TAG = $OVERRIDE_TAG"
 
 if [[ "$OVERRIDE_TAG" == "true" ]]; then
   echo "OverrideTag set to true. Will override ${AGENT_IMAGE_TAG_SUFFIX} image"
-  AZ_ACR_IMPORT_FORCE="--force"
```
pfrcks, line 38: pls validate that override tag ability still works by using a test image

deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
wanlonghenry: the value is not the same the ACR_NAME is containerinsights.azurecr.io  though using the full name is also fine, changed

deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
wanlonghenry: added

deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
```diff
@@ -66,17 +64,28 @@ fi
...
-  echo "-e error failed to retag and push image to destination ACR"
+  echo "-e error failed to login to az acr with managed identity credentials for containerinsights"
   exit 1
-fi
\ No newline at end of file
+fi
+
+if [ "$OVERRIDE_TAG" == "true" ] || [ "$TAG_EXISTS_STATUS" -ne 0 ]; then
+  echo "Copying ${SOURCE_IMAGE_FULL_PATH} to ${ACR_NAME}/${AGENT_IMAGE_FULL_PATH}"
+  oras copy -r $SOURCE_IMAGE_FULL_PATH $ACR_NAME/$AGENT_IMAGE_FULL_PATH --to-password $TOKEN
```
ganga1980: we should use --password-stdin  option instead of using  --to-password option. See the guidance what's best practice to pass the token
https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token

wanlonghenry: changed

ganga1980 (dismissed):

deployment/mergebranch-multiarch-agent-deployment/ServiceGroupRoot/Scripts/pushAgentToAcr.sh
wanlonghenry: tested pls check screenshot

ganga1980 (approved):

PR #1252, opened 2024-05-11
apply resourceTags on DCE also

ganga1980:


kubernetes/ama-logs.yaml
```diff
@@ -394,8 +394,8 @@ spec:
           imagePullPolicy: IfNotPresent
           resources:
             limits:
-              cpu: 500m
-              memory: 750Mi
+              cpu: 3
```
jatakiajanvi12, line 397: Did we update this in aks-rp manifest?

ganga1980: > Did we update this in aks-rp manifest?

I will be making  this as part of 3.1.21 release.

jatakiajanvi12 (approved):

PR #1253, opened 2024-05-14
win ama fix: Update the canary mcs endpoint

jatakiajanvi12:
This pull request includes a change in the `Set-AMA3PEnvironmentVariables` function in the `kubernetes/windows/main.ps1` file. The change adds a conditional statement to set a different value for the `$mcs_globalendpoint` variable if the `$aksRegion` variable is either "eastus2euap" or "centraluseuap". 

This change allows the system to use a different endpoint for the Azure Monitor for containers service in specific regions, which could be useful for testing or regional service variations.


kubernetes/windows/main.ps1
```diff
@@ -150,6 +150,9 @@ function Set-AMA3PEnvironmentVariables {
     $domain = "opinsights.azure.com"
     $mcs_endpoint = "https://monitor.azure.com/"
     $mcs_globalendpoint = "https://global.handler.control.monitor.azure.com"
+    if ($aksRegion.ToLower() -eq "eastus2euap" -or $aksRegion.ToLower() -eq "centraluseuap") {
+        $mcs_globalendpoint = "https://global.handler.canary.control.monitor.azure.com"
```
ganga1980, line 154: can you also make similar change here  https://github.com/microsoft/Docker-Provider/blob/ci_prod/source/plugins/go/src/ingestion_token_utils.go#L337 as well.

jatakiajanvi12: this change is only for the windows ama codepath....do we need to make the same changes for ingestion token?

ganga1980: > this change is only for the windows ama codepath....do we need to make the same changes for ingestion token?

> ingestion token code for windows only. we would need the same fix here as well.

source/plugins/go/src/ingestion_token_utils.go
```diff
@@ -335,6 +335,9 @@ func getAgentConfiguration(imdsAccessToken string) (configurationId string, chan
 	mcsEndpoint := os.Getenv("MCS_ENDPOINT")
 
 	AmcsEndpoint = fmt.Sprintf("https://global.handler.control.%s", mcsEndpoint)
+	if strings.Compare(strings.ToLower(resourceRegion), "eastus2euap") == 0 || strings.Compare(strings.ToLower(resourceRegion), "centraluseuap") == 0 {
+		AmcsEndpoint = fmt.Sprintf("https://global.handler.canary.control.%s", resourceRegion)
```
ganga1980: This should be mcsEndpoint not resourceRegion, right?

ganga1980 (approved):

PR #1284, opened 2024-06-26
Gangams/bugfix-high-log-scale-mode

ganga1980:
This PR has following changes
1. separate DCE for config and ingestion endpoint
2.  Config DCE only used when AMPLS configured
3. Ingestion DCE only used when Microsoft-ContainerLogV2-HighScale stream in the DCR.
4.  Config DCE resource MUST be the same region as AKS cluster
5.  Ingestion DCE resource MUST be the same region as Log Analytics workspace.
6. Ingestion DCE (if its exists) MUST be linked to the DCR.
7.  Add both Config and ingestion DCEs to AMPLS scope.

### DIFF
--- a/scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
+++ b/scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
@@ -111,9 +111,14 @@
         "enableHighLogScaleMode": "[if(contains(parameters('streams'),'Microsoft-ContainerLogV2-HighScale'), bool('true'), bool('false'))]",
         "associationName": "ContainerInsightsExtension",
         "dataCollectionRuleId": "[resourceId(variables('clusterSubscriptionId'), variables('clusterResourceGroup'), 'Microsoft.Insights/dataCollectionRules', variables('dcrName'))]",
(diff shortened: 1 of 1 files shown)


scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
```diff
@@ -333,12 +355,43 @@
         {
             "condition": "[parameters('useAzureMonitorPrivateLinkScope')]",
             "type": "Microsoft.Resources/deployments",
-            "name": "[Concat('aks-monitoring-msi-ampls-scope', '-',  uniqueString(parameters('aksResourceId')))]",
+            "name": "[Concat('aks-monitoring-msi-ampls-scope-config', '-',  uniqueString(parameters('aksResourceId')))]",
+            "apiVersion": "2017-05-10",
```
wanlonghenry, line 359: This api version is old, should we consider using the new ones?

wanlonghenry (approved):

scripts/onboarding/aks/onboarding-using-msi-auth/existingClusterOnboarding.json
ganga1980: > This api version is old, should we consider using the new ones?

We dont have dependency on the API version and I think, we should be good using GA API version.

PR #1286, opened 2024-06-27
Gangams/conformance tests mi migration

ganga1980:
add managed identity support for conformance e2e test framework

### DIFF
--- a/test/e2e/src/core/e2e_tests.sh
+++ b/test/e2e/src/core/e2e_tests.sh
@@ -73,7 +73,7 @@ waitForCIExtensionInstalled() {
 }
 
 validateCommonParameters() {
-    if [ -z $TENANT_ID ]; then
+   if [ -z $TENANT_ID ]; then
 	   echo "ERROR: parameter TENANT_ID is required." > ${results_dir}/error
 	   python3 setup_failure_handler.py
 	fi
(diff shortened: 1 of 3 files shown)


test/e2e/src/core/e2e_tests.sh
```diff
@@ -77,15 +77,25 @@ validateCommonParameters() {
...
-	if [ -z $CLIENT_SECRET ]; then
-	   echo "ERROR: parameter CLIENT_SECRET is required." > ${results_dir}/error
-	   python3 setup_failure_handler.py
-	fi
+      if [ -z $CLIENT_SECRET ]; then
+         echo "ERROR: parameter CLIENT_SECRET is required." > ${results_dir}/error
+         python3 setup_failure_handler.py
+      fi
+   else
+    if [[ -z "${CLIENT_ID}" ]]
```
pfrcks: this else clause is for when WORKLOAD_CLIENT_ID exists... why are we checking for CLIENT_ID's absence in this case?

ganga1980: > this else clause is for when WORKLOAD_CLIENT_ID exists... why are we checking for CLIENT_ID's absence in this case?

good catch. fixed.

test/e2e/src/core/e2e_tests.sh
```diff
@@ -73,19 +73,23 @@ waitForCIExtensionInstalled() {
 }
 
 validateCommonParameters() {
-    if [ -z $TENANT_ID ]; then
+   if [ -z $TENANT_ID ]; then
 	   echo "ERROR: parameter TENANT_ID is required." > ${results_dir}/error
 	   python3 setup_failure_handler.py
 	fi
```
pfrcks: misaligned?

pfrcks (approved):

PR #1297, opened 2024-07-23
add fluent-bit log files to logrotate

pfrcks:
This pull request involves changes to the `logrotate.conf` file in the `kubernetes/linux` directory. The changes include the addition of configuration settings for two new log files: `fluent-bit.log` and `fluent-bit-geneva.log`. The settings for these logs include directives for rotation, compression, size, and handling of missing files.

Key changes:

* [`kubernetes/linux/logrotate.conf`](diffhunk://#diff-560dc1db1cc54b05df1eb476da64f673361996b0ca681e448f06155459848250R40-R59): Added log rotation configuration for `fluent-bit.log` and `fluent-bit-geneva.log`. These configurations include directives for log rotation, compression, and size limit. They also include settings to handle missing log files and to delay compression until the second rotation cycle.


kubernetes/linux/logrotate.conf
```diff
@@ -37,3 +37,23 @@
...
+
+/var/opt/microsoft/docker-cimprov/log/fluent-bit-geneva.log {
+    copytruncate
+    rotate 7
+    missingok
+    notifempty
+    delaycompress
+    compress
+    size 10M
+}
```
ganga1980: add new line at the end? and also, can you please deploy and verify log rotation indeed working?

pfrcks:
tested this and after rotation fluent-bit fails to write to the fluent-bit.log file. This happens because fluent-bit.log file after rotation is a different file than the one it was writing to.

To avoid this we need to restart fluent-bit when logrotation happens

ganga1980:
> tested this and after rotation fluent-bit fails to write to the fluent-bit.log file. This happens because fluent-bit.log file after rotation is a different file than the one it was writing to.
> 
> To avoid this we need to restart fluent-bit when logrotation happens

> @pfrcks , restarting fluent-bit triggers the container restart which would loss previous log file. can you please reach out fluent-bit team over slack and find out what would be best way to rotate the fluent-bit.log?  is it through postrotate script to trigger the restart of fluent-bit on the log rotation to pick up new log file?  or is there any better way to implement this without losing previous log file?

PR #1308, opened 2024-08-19
Add windows telemetry for checking Telegraf running status when enabled via cmd

wanlonghenry:
Add windows telemetry for checking Telegraf running status when enabled via cmd

tasklist /FI "IMAGENAME eq telegraf.exe"
Image Name                     PID Session Name        Session#    Mem Usage
========================= ======== ================ =========== ============
telegraf.exe                 12756 Services                  20     66,984 K

### DIFF
--- a/source/plugins/go/src/telemetry.go
+++ b/source/plugins/go/src/telemetry.go
@@ -305,7 +305,9 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 			if telegrafEnabled["IsTelegrafEnabled"] == "true" {
 				isTelegrafRunning := isTelegrafRunning()
 				if !isTelegrafRunning {
-					telegrafEnabled["WinTelegrafNotRunning"] = "true"
+					telegrafEnabled["WinTelegrafRunning"] = "false"
+				} else {
+					telegrafEnabled["WinTelegrafRunning"] = "true"
 				}
 			}
 		}
--- a/source/plugins/go/src/telemetry.go
+++ b/source/plugins/go/src/telemetry.go
@@ -305,7 +305,7 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 			if telegrafEnabled["IsTelegrafEnabled"] == "true" {
 				isTelegrafRunning := isTelegrafRunning()
 				if !isTelegrafRunning {
-					telemetryDimensions["WinTelegrafNotRunning"] = "true"
+					telegrafEnabled["WinTelegrafNotRunning"] = "true"
 				}
 			}
 		}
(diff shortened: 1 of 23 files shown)


source/plugins/go/src/telemetry.go
```diff
@@ -302,6 +302,14 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 		osType := os.Getenv("OS_TYPE")
 		if osType != "" && strings.EqualFold(osType, "windows") {
 			telegrafEnabled["IsTelegrafEnabled"] = os.Getenv("TELEMETRY_CUSTOM_PROM_MONITOR_PODS") // If TELEMETRY_CUSTOM_PROM_MONITOR_PODS, then telegraf is enabled
```
ganga1980: change the variable name from telegrafEnabled to telegrafConfig since this is  holding multiple properties.

ganga1980: use the variable isTelegrafEnabled = os.Getenv("TELEMETRY_CUSTOM_PROM_MONITOR_PODS") and assign this value telegrafConfig["IsTelegrafEnabled"] = isTelegrafEnabled

source/plugins/go/src/telemetry.go
```diff
@@ -302,6 +302,14 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
...
+			if telegrafEnabled["IsTelegrafEnabled"] == "true" {
+				isTelegrafRunning := isTelegrafRunning()
+				if !isTelegrafRunning {
+					telegrafEnabled["WinTelegrafRunning"] = "false"
```
ganga1980: change the naming from WinTelegrafRunning to IsTelegrafRunning

source/plugins/go/src/utils_windows.go
```diff
@@ -49,3 +50,13 @@ func EnsureGenevaOr3PNamedPipeExists(namedPipeConnection *net.Conn, datatype str
 	}
 	return true
 }
+
+func isTelegrafRunning() bool {
```
ganga1980: you can make this function generic so that we can use this for other processes if required. For example

func isProcessRunning(processName string) (bool, error) {
    // Command to list all processes with the given name
    cmd := exec.Command("tasklist", "/FI", fmt.Sprintf("IMAGENAME eq %s", processName))
    output, err := cmd.Output()
    if err != nil {
        return false, err
    }

    // Check if the process name appears in the command output
    return strings.Contains(string(output), processName), nil
}

source/plugins/go/src/utils_linux.go
```diff
@@ -15,3 +19,13 @@ func EnsureGenevaOr3PNamedPipeExists(namedPipeConnection *net.Conn, datatype str
 	Log("Error::EnsureGenevaOr3PNamedPipeExists not implemented for Linux")
 	return false
 }
+
+func isTelegrafRunning() bool {
```
ganga1980: same here .. make this generic to work for any given process

PR #1310, opened 2024-08-22
Add telemetry for checking Telegraf running status when enabled

wanlonghenry:
Add telemetry for checking Telegraf running status when enabled
image

### DIFF
--- a/source/plugins/go/src/utils_linux.go
+++ b/source/plugins/go/src/utils_linux.go
@@ -2,8 +2,12 @@
 
 package main
 
-import "net"
-import "time"
+import (
+	"net"
+	"os/exec"
+	"strings"
+	"time"
+)
 
 func CreateWindowsNamedPipeClient(namedPipe string, namedPipeConnection *net.Conn) {
 	//function unimplemented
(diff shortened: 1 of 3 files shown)


ganga1980 (approved):

source/plugins/go/src/utils_linux.go
```diff
@@ -15,3 +19,14 @@ func EnsureGenevaOr3PNamedPipeExists(namedPipeConnection *net.Conn, datatype str
 	Log("Error::EnsureGenevaOr3PNamedPipeExists not implemented for Linux")
 	return false
 }
+
+func isProcessRunning(processName string) (string, error) {
```
ganga1980, line 23: did you verify this code works? linux process name will not have .exe, so you would need to trim that if the processName has .exe in it

wanlonghenry: yes verified works please check screenshot, also in the code line 303. it checked the os type before passing in telegraf.exe: if osType != "" && strings.EqualFold(osType, "windows")

ganga1980 (approved):

PR #1313, opened 2024-09-11
Do not patch or unpatch the object if it's being finalized

tokaplan:


tokaplan:
Turns out this is a wrong thing to do. We can't just

tokaplan:
Turns out this is a wrong thing to do. We cannot just let the object through the webhook if it's being finalized. That will result in unmodified customer  yaml being applied on top of a mutated yaml, which is not what we want. We must continue to mutate (or not) in accordance with our normal algorithm even for an object that is being deleted.
As an example - if we have an environment variable (e.g. NODE_NAME) that conflicts with the the same environment variable that mutation uses - we need the normal webhook algorithm to resolve the conflict (store the original value into a backup environment variable, etc.) rather than just let k8s apply customer yaml on top of mutated one.

PR #1340, opened 2025-01-13
Longw/Fix CVEs for linux vulnerabilities

wanlonghenry:
Fix CVEs for linux vulnerabilities, including base image update and address go dependency related vulnerabilities.

For verify the fix please reference latest build and CVE reports.
image
image
image


source/plugins/go/src/go.mod
```diff
@@ -39,11 +39,11 @@ require (
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/philhofer/fwd v1.1.2 // indirect
-	golang.org/x/net v0.26.0 // indirect
+	golang.org/x/net v0.33.0 // indirect
 	golang.org/x/oauth2 v0.19.0 // indirect
-	golang.org/x/sys v0.21.0 // indirect
-	golang.org/x/term v0.21.0 // indirect
-	golang.org/x/text v0.16.0 // indirect
+	golang.org/x/sys v0.28.0 // indirect
```
ganga1980, line 44: Few questions 1. dont we need these changes for other go input plugins? 2. what packages which has dependency of these packages? isnt that good idea to update the package which has these indirect dependencies? 
3. did we test this change to ensure no regressions?

wanlonghenry: These changes were introduced automatically by running go mod tidy after updating golang.org/x/net from v0.26.0 to v0.33.0. Will also share the screenshot for testing with this image.

suyadav1: The latest version for sys is [Version: v0.29.0](https://pkg.go.dev/golang.org/x/sys?tab=versions). Is there any reason we are going with v0.28.0?

wanlonghenry: the golang.org/x/net v0.33.0 depends on sys v0.28.0

suyadav1: /net latest version is also v0.34.0. Is this also dependent on something else?

wanlonghenry: good catch though it is too new, I have not tested this yet

suyadav1 (approved):

ganga1980 (approved):

PR #1349, opened 2025-02-03
Fix conformance test version conflict with AutoUpgradeMinorVersion

wanlonghenry:
Fix conformance test version conflict with AutoUpgradeMinorVersion

### DIFF
--- a/test/e2e/src/core/e2e_tests.sh
+++ b/test/e2e/src/core/e2e_tests.sh
@@ -131,7 +131,6 @@ createArcCIExtension() {
     if [ ! -z "$CI_ARC_RELEASE_TRAIN" ]; then
        basicparameters="$basicparameters  --release-train $CI_ARC_RELEASE_TRAIN"
     fi
-
     if [ ! -z "$CI_ARC_VERSION" ]; then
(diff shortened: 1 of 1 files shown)


test/e2e/src/core/e2e_tests.sh
```diff
@@ -131,9 +131,6 @@ createArcCIExtension() {
     if [ ! -z "$CI_ARC_RELEASE_TRAIN" ]; then
        basicparameters="$basicparameters  --release-train $CI_ARC_RELEASE_TRAIN"
     fi
-    if [ ! -z "$CI_ARC_VERSION" ]; then
```
ganga1980: Lets keep the version as is and instead set --auto-upgrade-minor-version false in basicparameters in https://github.com/microsoft/Docker-Provider/blob/657d856749e1843995bfbb740a304d64ed2286d4/test/e2e/src/core/e2e_tests.sh#L130 . This way, we know for sure, test will be validated against the version specified.

wanlonghenry: addressed

ganga1980 (approved):

PR #1366, opened 2025-03-07
Test automation: added tests for LA data flow

suyadav1:
Test automation: following tests are added in this PR:
1. Check if the data is flowing in all the tables.
2. Check if ContainerInventoy has any empty columns. This can be extended to any table.
3. Check if all nodes and pods' data is flowing.


test/ginkgo-e2e/querylogs/querylogs_test.go
```diff
@@ -0,0 +1,51 @@
...
+		},
+		// Check how to chose containerlog table
+		Entry("ContainerLog", "ContainerLog"),
+		Entry("ContainerLogV2", "ContainerLogV2"),
+		Entry("ContainerInventory", "ContainerInventory"),
+		Entry("ContainerNodeInventory", "ContainerNodeInventory"),
+		Entry("KubeNodeInventory", "KubeNodeInventory"),
+		Entry("KubePodInventory", "KubePodInventory"),
+		Entry("KubePVInventory", "KubePVInventory"),
+	)
```
ganga1980, line 33: Add InsightsMetrics and Perf table also.

test/ginkgo-e2e/querylogs/querylogs_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+var _ = Describe("When querying the logs for the table", func() {
+	DescribeTable("All tables should have logs",
+		func(table string) {
+			query := table + " | where TimeGenerated > ago(15m) | summarize count()"
+			err := utils.QueryLogsForCount(LogsClient, AKSResourceId, query, false)
+			Expect(err).NotTo(HaveOccurred())
+		},
+		// Check how to chose containerlog table
+		Entry("ContainerLog", "ContainerLog"),
```
ganga1980, line 27: Tables ContainerInventory, ContainerNodeInventory, Perf, ContainerLogV2/ContainerLog, InsightsMetrics should covered for both Windows and Linux.

test/ginkgo-e2e/querylogs/querylogs_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"docker-provider/test/utils"
+)
+
+var _ = Describe("When querying the logs for the table", func() {
+	DescribeTable("All tables should have logs",
+		func(table string) {
+			query := table + " | where TimeGenerated > ago(15m) | summarize count()"
+			err := utils.QueryLogsForCount(LogsClient, AKSResourceId, query, false)
+			Expect(err).NotTo(HaveOccurred())
```
ganga1980, line 23: In addition to this, we would also need to check number of records greater than 0 by parsing response.

test/ginkgo-e2e/querylogs/querylogs_test.go
ganga1980, line 27: ContainerLogV2 can be configured either through config map or DCR.  There are multiple ways to handle. May be best option to check the count is greater than 0 either for ContainerLog or ContainerLogV2.

test/ginkgo-e2e/querylogs/querylogs_test.go
```diff
@@ -0,0 +1,51 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"docker-provider/test/utils"
+)
+
+var _ = Describe("When querying the logs for the table", func() {
+	DescribeTable("All tables should have logs",
+		func(table string) {
+			query := table + " | where TimeGenerated > ago(15m) | summarize count()"
```
ganga1980, line 16: can we make this configurable?

test/ginkgo-e2e/utils/query_logs_api_utils.go
```diff
@@ -0,0 +1,147 @@
...
+	} else if logsTable == "KubePodInventory" {
+		pods, err := GetAllAgentPods(K8sClient)
+		if err != nil {
+			return err
+		}
+		for _, pod := range pods {
+			resources = append(resources, pod.Name)
+		}
+		query = logsTable + " | where TimeGenerated > ago(15m) | where ContainerName contains \"ama-logs\" | distinct Name"
+	}
```
ganga1980, line 144: we should use this query to get distinct pod count "KubePodInventory
| where TimeGenerated > ago(15m)
| distinct PodUid
| count "

test/ginkgo-e2e/querylogs/querylogs_test.go
suyadav1: configurable how? For checking the data flow, this query should work for all the tables assuming all tables has TimeGenerated column.

test/ginkgo-e2e/querylogs/querylogs_test.go
suyadav1: yes, we are doing that in logs utils.

test/ginkgo-e2e/querylogs/querylogs_test.go
suyadav1: Don't we need to check in KubeNodeInventory, KubePodInventory and KubePVInventory as well?
I'll add Perf and InsightsMetrics table.

test/ginkgo-e2e/utils/query_logs_api_utils.go
suyadav1: Checking the pod count only won't work. i.e. If a pod is destroyed and a new pod is created in last 5 mins but it's not emitting any data, the count of pods would still match which is not correct. If it does emit the data, the number of pods in the LA would be greater than the number of pods running currently so that won't match which is wrong as well.

test/ginkgo-e2e/querylogs/querylogs_test.go
ganga1980: @suyadav1 -  Kube* tables such as KubeNodeInventory, KubePodInventory and KubePVInventory  only emitted from Linux agent and replicaset.

suyadav1: our cicd clusters has linux agent so it's okay to keep these tables as well?

test/ginkgo-e2e/querylogs/querylogs_test.go
suyadav1: As discussed, will make the timestamp configurable in the next PR.

test/ginkgo-e2e/querylogs/querylogs_test.go
suyadav1: As discussed, will add a test for dataflow from each node in above mentioned tables in the next PR.

test/ginkgo-e2e/utils/query_logs_api_utils.go
suyadav1: Updated this testcase to check dataflow for all pods.

ganga1980 (approved):

PR #1379, opened 2025-03-22
Fix bug as we now introduced AZMON_WINDOWS_FLUENT_BIT_ENABLED

wanlonghenry:
Fix bug as we now introduced AZMON_WINDOWS_FLUENT_BIT_ENABLED

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -1065,8 +1065,8 @@ if [ ! -f /etc/cron.d/ci-agent ]; then
       echo "*/5 * * * * root /usr/sbin/logrotate -s /var/lib/logrotate/ci-agent-status /etc/logrotate.d/ci-agent >/dev/null 2>&1" >/etc/cron.d/ci-agent
 fi
 
-setGlobalEnvVar AZMON_RESOURCE_OPTIMIZATION_ENABLED "${AZMON_RESOURCE_OPTIMIZATION_ENABLED}"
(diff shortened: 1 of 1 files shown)


kubernetes/linux/main.sh
```diff
@@ -1065,8 +1065,8 @@ if [ ! -f /etc/cron.d/ci-agent ]; then
       echo "*/5 * * * * root /usr/sbin/logrotate -s /var/lib/logrotate/ci-agent-status /etc/logrotate.d/ci-agent >/dev/null 2>&1" >/etc/cron.d/ci-agent
 fi
 
-setGlobalEnvVar AZMON_RESOURCE_OPTIMIZATION_ENABLED "${AZMON_RESOURCE_OPTIMIZATION_ENABLED}"
```
ganga1980, line 1068: why AZMON_RESOURCE_OPTIMIZATION_ENABLED  being removed? are we not using both AZMON_RESOURCE_OPTIMIZATION_ENABLED  and AZMON_WINDOWS_FLUENT_BIT_ENABLED ?  We have bunch of places still we are using AZMON_RESOURCE_OPTIMIZATION_ENABLED  .

ganga1980 (changes requested):
left comments

kubernetes/linux/main.sh
ganga1980, line 1068: lets make sure this doesnt cause any another regressions.

wanlonghenry: AZMON_RESOURCE_OPTIMIZATION_ENABLED is for DS only   this is for controlling RS collection using fluentd thus using AZMON_WINDOWS_FLUENT_BIT_ENABLED

wanlonghenry: https://github.com/microsoft/Docker-Provider/blob/662bef02d45fd503cbda9a848627096a8b0d1161/kubernetes/linux/main.sh#L1102  this value is defined here

ganga1980 (approved):

PR #1385, opened 2025-03-28
Add converted governed release for arc canary stable

wanlonghenry:
Add converted governed release for arc canary stable


zanejohnson-azure:
two questions:
1. where does the Governed template come from? 
2. where is the arc k8s extension release yamls?

PR #1393, opened 2025-04-04
Fixing certificate housekeeper cronjob and validation pipeline

ramanujbhattacharjee93:
Granting the permission that the job was missing, reworking the way the job detects concurrent installer job, overhauling error handling.
Fixing validation pipeline so that we catch issues like that going forward.


appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -131,7 +131,7 @@ export class CertificateManager {
         }
     }
 
-    public async PatchSecretStore(operationId: string, kubeConfig: k8s.KubeConfig, certificate: WebhookCertData) {
+    public async UpdateSecretStore(operationId: string, kubeConfig: k8s.KubeConfig, certificate: WebhookCertData) {
```
tokaplan: Async methods that return "nothing" should return Promise<void>, let's update it here and for other async methods in this file while we are at it.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,22 +155,38 @@ export class CertificateManager {
...
-                };
+                try {
+                    certificate = {
+                        caCert: Buffer.from(secretsObj.data['ca.cert'], 'base64').toString('utf-8'),
+                        caKey: Buffer.from(secretsObj.data['ca.key'], 'base64').toString('utf-8'),
+                        tlsCert: Buffer.from(secretsObj.data['tls.cert'], 'base64').toString('utf-8'),
+                        tlsKey: Buffer.from(secretsObj.data['tls.key'], 'base64').toString('utf-8')
+                    };
+                } catch (error) {
+                    logger.error('Failed to decode secret data from base64!', operationId, this.requestMetadata);
```
tokaplan: A single logger.error call please, let's not split a single error message into multiple calls, that creates a false impression for the reader.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,22 +155,38 @@ export class CertificateManager {
         }
     }
 
-    public async GetSecretDetails(operationId: string, kubeConfig: k8s.KubeConfig): Promise<WebhookCertData> {
+    public async GetSecretDetails(operationId: string, kubeConfig: k8s.KubeConfig, clusterArmId: string, clusterArmRegion: string): Promise<WebhookCertData> {
         try {
             const k8sApi = kubeConfig.makeApiClient(k8s.CoreV1Api);
             const secretsObj: k8s.V1Secret = await k8sApi.readNamespacedSecret({ name: CertificateStoreName, namespace: KubeSystemNamespaceName });
+            let certificate: WebhookCertData = null;
             if (secretsObj.data) {
```
tokaplan: If secretsObj is undefined or null, what do we want? If we don't want to throw - we need to do secretsObj?.data or something.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,22 +155,38 @@ export class CertificateManager {
...
+                        tlsCert: null,
+                        tlsKey: null
+                    }
+                    logger.info('Returning empty certificate data due to decode failure.', operationId, this.requestMetadata);
+                }
 
                 return certificate;
             }
         } catch (error) {
             logger.error(JSON.stringify(error), operationId, this.requestMetadata);
```
tokaplan: Tell us what failed exactly. Which call? readNamespacedSecret?

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -186,14 +202,28 @@ export class CertificateManager {
...
             }
 
-            return Buffer.from(mutatingWebhookObject.webhooks[0].clientConfig.caBundle, 'base64').toString('utf-8');
+            let caBundle: string = null;
+
+            try {
+                caBundle = Buffer.from(mutatingWebhookObject.webhooks[0].clientConfig.caBundle, 'base64').toString('utf-8');
+            }
+            catch (error) {
+                logger.error('Failed to decode caBundle from MutatingWebhookConfiguration!', operationId, this.requestMetadata);
```
tokaplan: Single logger.error call for this error please

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -186,14 +202,28 @@ export class CertificateManager {
...
+                logger.error(JSON.stringify(error), operationId, this.requestMetadata);
+                await logger.SendEvent("MutatingWebhookCABundleDecodeFailure", operationId, null, null, null, true, error);
+
+                caBundle = null; // return null to indicate failure in decoding
+                logger.info('Returning null caBundle due to decode failure.', operationId, this.requestMetadata);
+            }
+
+            return caBundle;
         } catch (error) {
             logger.error(`Failed to get MutatingWebhookConfiguration! ${JSON.stringify(error)}`, operationId, this.requestMetadata);
```
tokaplan, line 211: SendEvent missing

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,22 +155,38 @@ export class CertificateManager {
...
             const secretsObj: k8s.V1Secret = await k8sApi.readNamespacedSecret({ name: CertificateStoreName, namespace: KubeSystemNamespaceName });
+            let certificate: WebhookCertData = null;
             if (secretsObj.data) {
-                const certificate: WebhookCertData = {
-                    caCert: Buffer.from(secretsObj.data['ca.cert'], 'base64').toString('utf-8'),
-                    caKey: Buffer.from(secretsObj.data['ca.key'], 'base64').toString('utf-8'),
-                    tlsCert: Buffer.from(secretsObj.data['tls.cert'], 'base64').toString('utf-8'),
-                    tlsKey: Buffer.from(secretsObj.data['tls.key'], 'base64').toString('utf-8')
-                };
+                try {
```
tokaplan: This is very confusing, why is this try nested into the first try? Let's have two separate tries in sequence - one for the API call, one for interpreting its results. As-is is very confusing.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -186,14 +202,28 @@ export class CertificateManager {
                 throw new Error("MutatingWebhookConfiguration not found or is malformed!");
             }
 
-            return Buffer.from(mutatingWebhookObject.webhooks[0].clientConfig.caBundle, 'base64').toString('utf-8');
+            let caBundle: string = null;
+
+            try {
```
tokaplan, line 177: This is very confusing, why is this try nested into the first try? Let's have two separate tries in sequence - one for the API call, one for interpreting its results. As-is is very confusing.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -228,6 +258,7 @@ export class CertificateManager {
         try {
             forge.pki.certificateFromPem(mwhcCaBundle);
             forge.pki.certificateFromPem(webhookCertData.caCert);
+            forge.pki.certificateFromPem(webhookCertData.caKey);
```
tokaplan: (sigh)

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -436,10 +470,12 @@ export class CertificateManager {
          * and the webhook deployment is restarted. If neither certificate is regenerated, the reconciliation is complete.
          */
         if (shouldUpdate) {
-            await this.PatchWebhookAndSecretStore(operationId, kc, webhookCertData, clusterArmId, clusterArmRegion);
+            await this.UpdateWebhookAndSecretStore(operationId, kc, webhookCertData, clusterArmId, clusterArmRegion);
             if (shouldRestartDeployment) {
                 logger.info('Restarting webhook deployment so the pods pick up new certificates...', operationId, this.requestMetadata);
+                await logger.SendEvent("CertificateDeploymentRestartStarted", operationId, null, clusterArmId, clusterArmRegion, true);
                 await this.RestartWebhookDeployment(operationId, kc, clusterArmId, clusterArmRegion);
```
tokaplan, line 491: What happens if this throws?

tokaplan (changes requested):

appmonitoring/ts/src/CertificateManager.ts
ramanujbhattacharjee93: No, if the secret doesn’t exist, the call will throw a 404 error. It wont return null from what I can tell

appmonitoring/ts/src/CertificateManager.ts
ramanujbhattacharjee93: then we must throw and fail. Otherwise the next run will think all is well

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
+                namespace: KubeSystemNamespaceName
+            });
+        } catch (error) {
+            logger.error(JSON.stringify(error), operationId, this.requestMetadata);
+            logger.SendEvent('ReadNamespacedSecretFailure', operationId, null, clusterArmId, clusterArmRegion, true, error);
+            throw error;
+        }
+
+        // If the secret has data, decode it
+        if (secretsObj && secretsObj.data) {
```
tokaplan: You mean
`secretsObj?.data`
?

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
+                certificate = {
                     caCert: Buffer.from(secretsObj.data['ca.cert'], 'base64').toString('utf-8'),
                     caKey: Buffer.from(secretsObj.data['ca.key'], 'base64').toString('utf-8'),
                     tlsCert: Buffer.from(secretsObj.data['tls.cert'], 'base64').toString('utf-8'),
                     tlsKey: Buffer.from(secretsObj.data['tls.key'], 'base64').toString('utf-8')
                 };
-
-                return certificate;
+            } catch (error) {
+                logger.error('Failed to decode secret data from base64!' + JSON.stringify(error), operationId, this.requestMetadata);
```
tokaplan: Let's use `blah ${variable}` syntax for clarify

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
         }
     }
 
-    public async GetSecretDetails(operationId: string, kubeConfig: k8s.KubeConfig): Promise<WebhookCertData> {
+    public async GetSecretDetails(operationId: string, kubeConfig: k8s.KubeConfig, clusterArmId: string, clusterArmRegion: string): Promise<WebhookCertData> {
+        let secretsObj: k8s.V1Secret;
+        let certificate: WebhookCertData = null;
```
tokaplan, line 160: You are returning null if secret returns null. Then further down the road it will throw an undefined exception in IsValid method... let's be careful and not allow these random things to occur. Positively handle this obvious edge case and make it explicit.

appmonitoring/ts/src/CertificateManager.ts
tokaplan: Yes, but also tell us the context of the failure:
`Error reading the secret: ${JSON.stringify(error)}`

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
-            {
-                throw new Error("MutatingWebhookConfiguration not found or is malformed!");
-            }
-
-            return Buffer.from(mutatingWebhookObject.webhooks[0].clientConfig.caBundle, 'base64').toString('utf-8');
+            mutatingWebhookObject = await webhookApi.readMutatingWebhookConfiguration({ name: MutatingWebhookConfigurationName });
+            
         } catch (error) {
             logger.error(`Failed to get MutatingWebhookConfiguration! ${JSON.stringify(error)}`, operationId, this.requestMetadata);
+            await logger.SendEvent("MutatingWebhookConfigurationFetchFailure", operationId, null, null, null, true, error);
```
tokaplan: Why the nulls? Pass it in if we need it, much nicer when there's context attached to each event. You can always find the entire trace by operationId, but if you don't know and you're looking by something else you'll miss random events, not good.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
+        try {
+            if (!mutatingWebhookObject
+                || !mutatingWebhookObject.webhooks
+                || mutatingWebhookObject.webhooks.length !== 1 || !mutatingWebhookObject.webhooks[0].clientConfig) {
+                throw new Error("MutatingWebhookConfiguration not found or is malformed!");
+            }
+            caBundle = Buffer.from(mutatingWebhookObject.webhooks[0].clientConfig.caBundle, 'base64').toString('utf-8');
+        } catch (error) {
+            logger.error('Failed to decode caBundle from MutatingWebhookConfiguration!' + JSON.stringify(error), operationId, this.requestMetadata);
+            await logger.SendEvent("MutatingWebhookCABundleDecodeFailure", operationId, null, null, null, true, error);
```
tokaplan: Nulls again... let's fill these out.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
                 throw new Error("MutatingWebhookConfiguration not found or is malformed!");
             }
             mutatingWebhookObject.webhooks[0].clientConfig.caBundle = Buffer.from(certificate.caCert, 'utf-8').toString('base64');
-            
+
             await webhookApi.replaceMutatingWebhookConfiguration({ name: MutatingWebhookConfigurationName, body: mutatingWebhookObject });
         } catch (error) {
-            logger.error('Failed to patch MutatingWebhookConfiguration!', operationId, this.requestMetadata);
-            logger.error(JSON.stringify(error), operationId, this.requestMetadata);
+            logger.error('Failed to patch MutatingWebhookConfiguration!' + JSON.stringify(error), operationId, this.requestMetadata);
```
tokaplan: `${}`

tokaplan (changes requested):

appmonitoring/ts/src/CertificateManager.ts
ramanujbhattacharjee93: oh I did not realize you could do `?` in ts

appmonitoring/ts/src/CertificateManager.ts
ramanujbhattacharjee93: We pass this into a validator function down the line which handles if the certificate returned is empty or not proper. The validation method `IsValidCertificate`  is meant to handle any edge case with that

appmonitoring/ts/src/CertificateManager.ts
ramanujbhattacharjee93: hmm I did not realize you could do it like that

appmonitoring/ts/src/CertificateManager.ts
tokaplan: It will work, but it's not written as if it expects _webhookCertData_ to be null, it really seems like it just happens to work by accident. If that's the case - can we explicitly handle this maybe? Like throw explicitly if webhookCertData is null? Or in the very least say `webhookCertData?.caCert` instead of `webhookCertData.caCert` to show that you're aware it can be null?

```C#
        try {
            forge.pki.certificateFromPem(mwhcCaBundle);
            forge.pki.certificateFromPem(webhookCertData.caCert);
            forge.pki.certificateFromPem(webhookCertData.caKey);
            forge.pki.certificateFromPem(webhookCertData.tlsCert);
            forge.pki.privateKeyFromPem(webhookCertData.tlsKey);
            return true;
        } catch (error) {
            logger.error(`Error occured while trying to validate certificates. ${JSON.stringify(error)}`, operationId, this.requestMetadata);
            await logger.SendEvent("CertificateValidationFailure", operationId, null, clusterArmId, clusterArmRegion, true, error);
            return false;
        }```

File: appmonitoring/ts/src/CertificateManager.ts [hunk 17]
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
+                namespace: KubeSystemNamespaceName
+            });
+        } catch (error) {
+            logger.error(`Error reading the secret: ${JSON.stringify(error)}`, operationId, this.requestMetadata);
+            logger.SendEvent('ReadNamespacedSecretFailure', operationId, null, clusterArmId, clusterArmRegion, true, error);
+            throw error;
+        }
+
+        // If the secret has data, decode it
+        if (secretsObj && secretsObj?.data) {
```
tokaplan: You mean `secretsObj?.data`? No need to have a chain of two checks.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
         } catch (error) {
             logger.error(`Failed to get MutatingWebhookConfiguration! ${JSON.stringify(error)}`, operationId, this.requestMetadata);
+            await logger.SendEvent("MutatingWebhookConfigurationFetchFailure", operationId, null, clusterArmId, clusterArmRegion, true, error);
             throw error;
         }
+
+        try {
+            if (!mutatingWebhookObject
+                || !mutatingWebhookObject.webhooks
+                || mutatingWebhookObject.webhooks.length !== 1 || !mutatingWebhookObject.webhooks[0].clientConfig) {
```
tokaplan: These 3 checks could be a single check using `?.` notation, please use it.

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,96 @@ export class CertificateManager {
...
         try {
             const webhookApi: k8s.AdmissionregistrationV1Api = kubeConfig.makeApiClient(k8s.AdmissionregistrationV1Api);
             const mutatingWebhookObject: k8s.V1MutatingWebhookConfiguration = await webhookApi.readMutatingWebhookConfiguration({ name: MutatingWebhookConfigurationName });
-            if (!mutatingWebhookObject 
-                || !mutatingWebhookObject.webhooks 
-                || mutatingWebhookObject.webhooks.length !== 1 || !mutatingWebhookObject.webhooks[0].clientConfig)
-            {
+            if (!mutatingWebhookObject
+                || !mutatingWebhookObject.webhooks
+                || mutatingWebhookObject.webhooks.length !== 1 || !mutatingWebhookObject.webhooks[0].clientConfig) {
```
tokaplan: Same here with `?.`

tokaplan (changes requested):

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -155,60 +155,93 @@ export class CertificateManager {
...
+
+            return certificate;
         }
+
+        return null;
     }
 
-    public async GetMutatingWebhookCABundle(operationId: string, kubeConfig: k8s.KubeConfig): Promise<string> {
+    public async GetMutatingWebhookCABundle(operationId: string, kubeConfig: k8s.KubeConfig, clusterArmId: string, clusterArmRegion: string): Promise<string> {
+        let caBundle: string = null;
```
tokaplan: If you're explicitly setting caBundle to null in catch, let's not initialize it here. That serves as a compiler-enforced protection against missing a branch where it's accidentally not initialized.

tokaplan (changes requested):

.pipelines/azure_pipeline_validation_appmonitoring.yaml
```diff
@@ -18,7 +18,8 @@ variables:
   armServiceConnectionName: 'Appmonitoring_Validation_Pipeline'
   subscription: '5a3b3ba4-3a42-42ae-b2cb-f882345803bc'
   containerRegistry: 'aicommon' # this build pushes the aidev image to the dev ACR (caution: ACR that has "prod" in its name is the dev one, the one without "prod" is prod AME)
```
tokaplan: Does this comment need to be updated?

.pipelines/azure_pipeline_validation_appmonitoring.yaml
```diff
@@ -222,24 +231,90 @@ jobs:
...
+        sudo chmod u+x ./validate-certs.sh
+
+        echo "Checking for charts..."
+        if ! ./test-cronjob.sh; then
+          echo "Cron Job validation failed"
+          exit 1
+        fi
+
+  - task: AzureCLI@2
+    displayName: "Restart test Apps"
```
tokaplan: This step appears to restart the webhook, not the apps. Should we clarify?

appmonitoring/validation-helm/app-monitoring-addon/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+{
```
tokaplan, line 1: Let's remember to update the release instructions to explain how to copy over the chart from AKS RP repo.
Let's also make a record in _change.log_ file describing what we did here.

tokaplan (changes requested):

.pipelines/azure_pipeline_validation_appmonitoring.yaml
ramanujbhattacharjee93: ah yeah the name is wrong

appmonitoring/ts/src/CertificateManager.ts
ramanujbhattacharjee93: Yeah good point!

appmonitoring/ts/src/CertificateManager.ts
```diff
@@ -225,53 +258,65 @@ export class CertificateManager {
...
+        {
+            logger.info('WebhookCertData is null', operationId, this.requestMetadata);
+            await logger.SendEvent("CertificateFetchValueFailure", operationId, null, clusterArmId, clusterArmRegion, true);
+            return false;
+        }
+
         try {
             forge.pki.certificateFromPem(mwhcCaBundle);
             forge.pki.certificateFromPem(webhookCertData.caCert);
+            forge.pki.privateKeyFromPem(webhookCertData.caKey);
```
tokaplan, line 271: Let's cover the scenario where the certs are NOT regenerated (all good, nothing to do) with a unit test so that this can't be missed

tokaplan (changes requested):

tokaplan (approved):

PR #1411, opened 2025-04-29
Skip fluentd and telegraf in sidecar and rs when logs and events profile is enabled

suyadav1:


build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -21,46 +21,41 @@
 return if !@os_type.nil? && !@os_type.empty? && @os_type.strip.casecmp('windows').zero?
 return unless ENV['USING_AAD_MSI_AUTH']&.strip&.casecmp?('true')
```
wanlonghenry, line 23: could you share the test results here and that data won't be collected when logs and events profile is enabled

build/linux/installer/scripts/dcr-config-parser.rb
```diff
@@ -21,46 +21,41 @@
...
 
-if !@controllerType.nil? && !@controllerType.empty? && @controllerType.strip.casecmp("daemonset").zero? \
```
wanlonghenry, line 24: for daemonset, what is the behavior

build/linux/installer/scripts/livenessprobe.sh
```diff
@@ -17,18 +17,15 @@ syslogSetup() {
...
-    syslogSetup
-    CURRENT_LOGS_AND_EVENTS_ONLY=${LOGS_AND_EVENTS_ONLY}
-    ruby /opt/dcr-config-parser.rb > /dev/write-to-traces 2>&1
-    source /opt/dcr_env_var
-    if [ "${LOGS_AND_EVENTS_ONLY}" != "${CURRENT_LOGS_AND_EVENTS_ONLY}" ]; then
-      echo "dcr_env_var has been updated - dcr config changed" > /dev/termination-log
-      exit 1
-    fi
-  fi
+  syslogSetup
```
wanlonghenry, line 20: could you share detail more about always runs after syslogSetup when CONTROLLER_TYPE == DaemonSet.

PR #1415, opened 2025-05-02
Gangams/azure mdsd 1.35.1 update

ganga1980:
MDSD update to latest version. This version has ARM64 support for AMACA which is being used in high log scale mode.

Note > Increase in resource usage (primarily in memory) in ARM64 node is expected because before AMACA was not running on ARM64 node.

Memory usage with and without update

image

CPU usage with and without update
image

1. Verified all the data types ingested correctly.
2. No relevant errors and warnings (few unreleated warnings and errors which can be ignored)
3. validated on azure linux, ubuntu linux, x64, ARM 64 and FIPs nodes


zanejohnson-azure (dismissed):

kubernetes/linux/main.sh
```diff
@@ -27,10 +27,12 @@ startAMACoreAgent() {
          echo "export DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=$DOTNET_SYSTEM_GLOBALIZATION_INVARIANT"
          echo "export PA_CONFIG_PORT=$PA_CONFIG_PORT"
          echo "export CounterDataReportFrequencyInMinutes=$CounterDataReportFrequencyInMinutes"
+         echo "export AMACALogFilePath=$AMACALogFilePath"
+         echo "export AMACAConfigFilePath=$AMACAConfigFilePath"
       } >> ~/.bashrc
 
       source ~/.bashrc
-      /opt/microsoft/azure-mdsd/bin/amacoreagent -c $AMACAConfigFilePath --configport $PA_CONFIG_PORT --amacalog $AMACALogFilePath > /dev/null 2>&1 &
+      /opt/microsoft/azure-mdsd/bin/amacoreagent -c $AMACAConfigFilePath --configport $PA_CONFIG_PORT --amacalog $AMACALogFilePath --giglaport $PA_DATA_PORT > /dev/null 2>&1 &
```
suyadav1: FMI.. Where was PA_DATA_PORT used earlier? and why this change?

wanlonghenry: this is coming from https://github.com/microsoft/Docker-Provider/compare/ci_prod...thperapp/testCI-ARM64-MDSD

wanlonghenry (dismissed):

kubernetes/linux/main.sh
ganga1980: @suyadav1 , good question. earlier it was used from PA.json . with latest version, required as argument. cleaned up all the unused ones. PA_DATA_PORT  enviroment variable used to check whether the port is in listening state or not.

suyadav1 (approved):

PR #1424, opened 2025-05-06
longw/update windows 2019 build VM using 2022

wanlonghenry:
longw/update windows 2019 build using 2022 VM

### DIFF
--- a/.pipelines/azure_pipeline_mergedbranches.yaml
+++ b/.pipelines/azure_pipeline_mergedbranches.yaml
@@ -378,6 +378,8 @@ jobs:
   -  common
   pool:
     name: Azure-Pipelines-Windows-CI-Test-EO
+    image: ci-1es-managed-windows-2022
+    os: windows
   variables:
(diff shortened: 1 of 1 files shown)


.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -378,6 +378,8 @@ jobs:
   -  common
   pool:
     name: Azure-Pipelines-Windows-CI-Test-EO
+    image: ci-1es-managed-windows-2022
```
zanejohnson-azure, line 381: 3Q:
1. what ci-1es-managed-windows-2022?
2. why do you think it will fix the problem? is it because our 2022 windows build rarely fail?
3. If it can fix the 2019 problem, why is it safe to use 2022 for 2019? any risk or regression?

wanlonghenry: This VM is for build purpose only while base image is still 2019. Build failed because of 2019 VM issue.

PR #1438, opened 2025-05-12
Migrate esrp from corp to ame

wanlonghenry:
Migrate esrp from corp to ame: https://eng.ms/docs/microsoft-security/identity/trust-and-security-services/tss-high-security-environments/tss-esrp-fabric-and-platform-services/esrp-documentation/tsgs/sfi/tsg501-eliminate-access-to-codesigning-from-corp-tenants
Key Vault: ESRPContainerInsightsAME
MSI: ESRPServiceConnectionContainerInsights


ganga1980 (approved):

.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -69,6 +70,7 @@ jobs:
 
       echo "appRegistrationClientId is $APP_REGISTRATION_CLIENT_ID"
       echo "appRegistrationTenantId is $APP_REGISTRATION_TENANT_ID"
+      echo "esrpClientId is $ESRP_CLIENT_ID"
       echo "authAKVName is $AUTH_AKV_NAME"
       echo "authCertName is $AUTH_CERT_NAME"
       echo "authSignCertName is $AUTH_SIGN_CERT_NAME"
```
suyadav1, line 76: let's not print the clientId and tenantId in the logs due to security reasons. We should also mark these as a secret in the build pipeline params with an option to override.

wanlonghenry: As it does not contain secret, so it should be fine. Will also adjust params for build pipeline.

PR #1439, opened 2025-05-12
Mariner 3 upgrade

suyadav1:
### Resource comparison:
**Cluster with Arm64 and Amd64 nodes:**
CPU (3.1.27 vs Mariner 3):
image

Memory consumption (3.1.27 vs Mariner 3):
image

**Cluster with Arm64, Amd64 and FIPS - with custom agent config**
CPU (3.1.27 vs Mariner 3):
image

Memory consumption (3.1.27 vs Mariner 3):
image


ganga1980 (changes requested):
Following scenarios needs to be validated  for this change
1.  Cluster with Azure Linux (2.0 and 3.0) and Ubuntu  Nodes
2. ARM64 and x64 nodes
3. FIPs nodes
4. MSI auth and legacy auth

Please capture the resource usage with and without this change to verify if there are any perf regressions

kubernetes/linux/setup.sh
```diff
@@ -12,50 +12,28 @@ fi
...
 sudo update-ca-trust
 
-# sudo tdnf install ruby-3.1.3 -y
 if [ "$ARCH" == "arm64" ]; then
-    sudo tdnf install ruby-3.1.3-1.cm2.aarch64 -y
+    sudo tdnf install ruby-3.3.5-1.azl3.aarch64 -y
 else
     tdnf install -y gcc patch bzip2 openssl-devel libyaml-devel libffi-devel readline-devel zlib-devel gdbm-devel ncurses-devel
-    wget https://github.com/rbenv/ruby-build/archive/refs/tags/v20230330.tar.gz -O ruby-build.tar.gz
+    wget https://github.com/rbenv/ruby-build/archive/refs/tags/v20250409.tar.gz -O ruby-build.tar.gz
```
ganga1980, line 21: Please add comments  1. why we are not using tdnf install if installing tdnf doesnt work 2. why we have different versions for arm64 and x64.

kubernetes/linux/setup.sh
```diff
@@ -12,50 +12,28 @@ fi
...
     tar -xzf ruby-build.tar.gz
     PREFIX=/usr/local ./ruby-build-*/install.sh
-    ruby-build 3.1.3 /usr
+    ruby-build 3.3.8 /usr -v
+
+    rm ruby-build.tar.gz
```
ganga1980, line 26: is this not redudant with "rm -rf ruby-build-*" at line 28.

kubernetes/linux/setup.sh
```diff
@@ -12,50 +12,28 @@ fi
...
-gem uninstall cgi --version 0.3.5
-rm -f /usr/lib/ruby/gems/3.1.0/specifications/default/cgi-0.3.5.gemspec
-rm -rf /usr/lib/ruby/gems/3.1.0/gems/cgi-0.3.5
-gem uninstall net-imap --version 0.2.3
-gem uninstall time --version 0.2.0
-gem uninstall uri --version 0.11.0
-gem uninstall stringio --version 3.0.1
-gem uninstall rexml --version 3.2.5
-gem uninstall webrick --version 1.8.1
+rm /usr/lib/ruby/gems/3.3.0/specifications/default/openssl-3.2.0.gemspec
```
ganga1980, line 30: questions 1. are these helping to clean up default unused gems? 2. no side effects because of this, right?

kubernetes/linux/setup.sh
```diff
@@ -90,16 +68,16 @@ docker_cimprov_version=$(sudo tdnf list installed | grep docker-cimprov | awk '{
 echo "DOCKER_CIMPROV_VERSION=$docker_cimprov_version" >> packages_version.txt
 
 #install fluent-bit
-sudo tdnf install fluent-bit-3.0.6 -y
+sudo tdnf install fluent-bit-3.1.9 -y
 echo "$(fluent-bit --version)" >> packages_version.txt
 
 # install fluentd using the mariner package
-# sudo tdnf install rubygem-fluentd-1.14.6 -y
+# sudo tdnf install rubygem-fluentd -y
```
ganga1980: we can clean up these, right?

ganga1980 (changes requested):
left comments

kubernetes/linux/setup.sh
suyadav1: I don't see these gems at these locations, so yes, they are getting cleaned up.

kubernetes/linux/setup.sh
suyadav1: removed

kubernetes/linux/setup.sh
suyadav1: added

suyadav1:
> Following scenarios needs to be validated for this change
> 
> 1. Cluster with Azure Linux (2.0 and 3.0) and Ubuntu  Nodes
> 2. ARM64 and x64 nodes
> 3. FIPs nodes
> 4. MSI auth and legacy auth
> 
> Please capture the resource usage with and without this change to verify if there are any perf regressions

Tested all these scenarios. Don't see any issues in any of these.

ganga1980 (dismissed):

kubernetes/linux/setup.sh
```diff
@@ -12,50 +12,27 @@ fi
 sudo tdnf install ca-certificates-microsoft -y
 sudo update-ca-trust
 
-# sudo tdnf install ruby-3.1.3 -y
```
wanlonghenry, line 15: what is the plan for windows

suyadav1: Do you mean the plan to upgrade ruby for windows? I am not sure about this. @ganga1980 are we planning to do this?

suyadav1: Disussed this with @ganga1980, there is no dependency between upgrading the linux and windows version. We will upgrade windows version in upcoming sprints.

ganga1980 (approved):

PR #1460, opened 2025-06-02
TAF: Check errors in process files

suyadav1:
New tests are added as part of this PR to check if there are any errors present in any of the processes running.

### DIFF
--- a/test/testkube/testkube-test-crs.yaml
+++ b/test/testkube/testkube-test-crs.yaml
@@ -13,7 +13,7 @@ spec:
     repository:
       type: git
       uri: https://github.com/microsoft/Docker-Provider/
-      branch: suyadav/taf-newtest-fileerrorcheck
+      branch: ci_prod
       path: test/ginkgo-e2e
   executionRequest:
     variables:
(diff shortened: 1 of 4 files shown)


test/testkube/testkube-test-crs.yaml
```diff
@@ -16,6 +16,11 @@ spec:
       branch: ci_prod
       path: test/ginkgo-e2e
   executionRequest:
+    variables:
+      GENEVA_INTEGRATION:
+        name: GENEVA_INTEGRATION
+        value: "$GENEVA_INTEGRATION"
+        type: basic
```
wanlonghenry, line 23: do we want to set a default value here

suyadav1: The default value will be passed from the pipeline. I don't think we can set a default value here as the variable $GENEVA_INTEGRATION needs to be replaced with the value from the pipeline.

wanlonghenry (approved):

zanejohnson-azure:
besides checking errors of different process. we also want to check the container log error, something like below

kubectl logs -n $NAMESPACE $POD -c $CONTAINER

suyadav1:
> besides checking errors of different process. we also want to check the container log error, something like below
> 
> kubectl logs -n $NAMESPACE $POD -c $CONTAINER

This test is already added. Please check "The container logs should not contain errors" in containerstatus tests.

zanejohnson-azure (approved):

PR #1468, opened 2025-06-09
Longw/containerlogv2 fix telemetry bug

wanlonghenry:
Longw/containerlogv2 fix telemetry bug
When msi mode, the containerlogv2 telemetry will based on the extension fetch
https://msazure.visualstudio.com/InfrastructureInsights/_workitems/edit/33107568

### DIFF
--- a/source/plugins/go/input/lib/cadvisor.go
+++ b/source/plugins/go/input/lib/cadvisor.go
@@ -16,8 +16,6 @@ import (
 	"strings"
 	"time"
 
-	"Docker-Provider/source/plugins/go/src/extension"
-
 	"github.com/sirupsen/logrus"
 )
 
@@ -76,8 +74,6 @@ var (
 	winContainerPrevMetricRate              = make(map[string]float64)
 	Log                                     *logrus.Logger
 	osType                                  string
-	IsContainerLogV2DCR                     bool
-	ContainerType                           = os.Getenv("CONTAINER_TYPE")
 )
 
 func init() {
@@ -103,8 +99,6 @@ func init() {
 
 	Log.SetOutput(file)
 	Log.SetLevel(logrus.InfoLevel)
-
-	FLBLogger = CreateLogger(LogPath)
 }
 
 type metricDataItem map[string]interface{}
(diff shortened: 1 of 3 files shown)


source/plugins/go/input/lib/cadvisor.go
```diff
@@ -74,6 +76,8 @@ var (
 	winContainerPrevMetricRate              = make(map[string]float64)
 	Log                                     *logrus.Logger
 	osType                                  string
+	IsContainerLogV2DCR                     bool
+	ContainerType                           = os.Getenv("CONTAINER_TYPE")
```
zanejohnson-azure: they should not be capitalized as they are only used inside this file.

zanejohnson-azure:
can you add a work item for this PR so I can have more context what is the problem to fix.

zanejohnson-azure:
please also post tests in the PR for review.

source/plugins/go/input/lib/cadvisor.go
```diff
@@ -1067,6 +1077,13 @@ func getContainerCpuMetricItems(metricInfo map[string]interface{}, hostName, met
 						}
 						if len(os.Getenv("AZMON_CONTAINER_LOG_SCHEMA_VERSION")) > 0 {
 							telemetryProps["containerLogVer"] = os.Getenv("AZMON_CONTAINER_LOG_SCHEMA_VERSION")
+							if IsAADMSIAuthMode() {
+								if IsContainerLogV2DCR {
+									telemetryProps["containerLogVerDCR"] = "v2"
+								} else {
+									telemetryProps["containerLogVerDCR"] = "v1"
```
ganga1980: why do we need all these changes? cant we simply call in os.Setenv("AZMON_CONTAINER_LOG_SCHEMA_VERSION", "v2") in IsContainerLogV2 function or in oms.go should reflect the version info automatically, right?

ganga1980 (commented):
left comments

source/plugins/go/input/lib/cadvisor.go
wanlonghenry: AZMON_CONTAINER_LOG_SCHEMA_VERSION was defined from config map, so in this case, we could either call function in cadvisor or set a new ENV variable

PR #1469, opened 2025-06-10
Fix multiarch Linux support

tokaplan:


.pipelines/azure_pipeline_validation_appmonitoring.yaml
```diff
@@ -458,9 +458,14 @@ jobs:
...
-        docker tag $(testImageName) $(fullImageName)
-        docker push $(fullImageName)
+        # docker pull $(testImageName)
+        # docker tag $(testImageName) $(fullImageName)
+        # docker push $(fullImageName)
+        
+        echo "Retagging ${testImageName} as ${fullImageName} at ${{ variables.containerRegistry }}..."
+        imageName="${fullImageName#${{ variables.containerRegistry }}}"
+        echo "Image name: $imageName"
+        az acr import --name ${{ variables.containerRegistry }} --source $(testImageName) --image imageName #--force
```
tokaplan: @ramanujbhattacharjee93 can you think about this for a moment and tell me if we need --force here? Are we ever in a situation when the image would already exist at the destination and we'd want to overwrite? I'd rather not have it here unless you can think of a reason.

ramanujbhattacharjee93 (dismissed):

.pipelines/azure_pipeline_validation_appmonitoring.yaml
ramanujbhattacharjee93: Yes I would prefer not to have it. That will kinda create loopholes we dont want

ramanujbhattacharjee93 (dismissed):

ramanujbhattacharjee93 (approved):

PR #1478, opened 2025-06-12
Changed default pod chunk size to 10

suyadav1:
Verified the pod chunk size changing to 10 in Telemetry. 
No regression in data flow and CPU or memory usage.


build/common/installer/scripts/tomlparser-agent-config.rb
```diff
@@ -18,7 +18,7 @@
 # 250 Node items (15KB per node) account to approximately 4MB
 @nodesChunkSize = 250
 # 1000 pods (10KB per pod) account to approximately 10MB
-@podsChunkSize = 1000
+@podsChunkSize = 10
```
wanlonghenry, line 21: have we verified the chunk size change in all our ci cd clusters and cosmic clusters with large scale? and for this change, do we plan to rollout per AKS SDP

suyadav1:
we will take this change in the next release

PR #1479, opened 2025-06-12
Upgraded fluent-bit for windows with latest version

suyadav1:
Memory usage (3.1.27 vs new image):
image

I see some data increase in Prometheus metrics when I upgrade from current prod image but It's not related to the change - the metrics are coming differently for a linux node. When I downgrade from this change's image to current prod image, I see consistent numbers:
image
image

### DIFF
--- a/kubernetes/windows/setup.ps1
+++ b/kubernetes/windows/setup.ps1
@@ -26,7 +26,7 @@ Write-Host ('Creating folder structure')
 Write-Host ('Installing Fluent Bit');
 
     try {
-        $fluentBitUri='https://fluentbit.io/releases/3.0/fluent-bit-3.0.6-win64.zip'
+        $fluentBitUri='https://packages.fluentbit.io/windows/fluent-bit-4.0.3-win64.zip'
         Invoke-WebRequest -Uri $fluentBitUri -OutFile /installation/fluent-bit.zip
         Expand-Archive -Path /installation/fluent-bit.zip -Destination /installation/fluent-bit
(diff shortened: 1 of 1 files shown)


kubernetes/windows/setup.ps1
```diff
@@ -26,7 +26,7 @@ Write-Host ('Creating folder structure')
 Write-Host ('Installing Fluent Bit');
 
     try {
-        $fluentBitUri='https://fluentbit.io/releases/3.0/fluent-bit-3.0.6-win64.zip'
+        $fluentBitUri='https://packages.fluentbit.io/windows/fluent-bit-4.0.3-win64.zip'
```
ganga1980, line 29: @suyadav1 - can you please capture what scenarios being validated and also attach the resource usage with and without change? did you verify that, there are no errors in the fluent-bit log files?  and also, did you review there are no breaking changes between 3.0.6 and 4.0.3 ?

wanlonghenry, line 29: please share the approximate resource utilization change after major version bump for fluent-bit windows

suyadav1: I have updated the details of this PR with more info. I also checked that there are no errors in fluent-bit for windows agent.

suyadav1: added details of memory usage in the description.

ganga1980 (approved):

wanlonghenry (approved):

PR #1480, opened 2025-06-12
CVE 202543857: uninstall net-imap gem

suyadav1:
Memory usage (3.1.27 vs this change):
image

No issues in the data flow or resource consumption.

### DIFF
--- a/kubernetes/linux/setup.sh
+++ b/kubernetes/linux/setup.sh
@@ -34,9 +34,8 @@ rm -rf /usr/lib/ruby/gems/3.3.0/gems/find-0.2.0
 rm /usr/lib/ruby/gems/3.3.0/specifications/default/rdoc-6.6.3.1.gemspec
 rm -rf /usr/lib/ruby/gems/3.3.0/gems/rdoc-6.6.3.1
 
-# remove net-imap gem and strscan as it has a known CVE (CVE-2025-43857) and is not used by the agent
+# remove net-imap gem as it has a known CVE (CVE-2025-43857) and is not used by the agent
 gem uninstall net-imap --force
-gem uninstall strscan --force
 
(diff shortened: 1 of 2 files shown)


kubernetes/linux/setup.sh
```diff
@@ -34,6 +34,9 @@ rm -rf /usr/lib/ruby/gems/3.3.0/gems/find-0.2.0
 rm /usr/lib/ruby/gems/3.3.0/specifications/default/rdoc-6.6.3.1.gemspec
 rm -rf /usr/lib/ruby/gems/3.3.0/gems/rdoc-6.6.3.1
 
+# remove net-imap gem as it has a known CVE (CVE-2025-43857) and is not used by the agent
```
zanejohnson-azure, line 37: how to get the conclusion that it is not used by ci agent?

suyadav1: tested the changes on the local cluster and everything seems to be working without any issues

ganga1980: @suyadav1 , @zanejohnson-azure  was mentioning about the tool to check reverse dependencies. did you verify with that tool? Btw, unless we test all the possible scenarios and its hard to conclude if this is not being used.

ganga1980: as we discussed, lets double check if this is indirect dependency of the ruby code and also do the same for windows.

suyadav1: there is no dependency on net-imap:

PS C:\opt\amalogswindows\scripts\powershell> gem dependency net-imap --reverse-dependencies
Gem net-imap-0.2.3
  digest (>= 0)
  net-protocol (>= 0)
  strscan (>= 0)

zanejohnson-azure (approved):

PR #1488, opened 2025-06-17
Longw/high scale terraform

wanlonghenry:
Longw/high scale terraform
With high log scale and config map set:
image
image
without high log scale:
image


ganga1980 (commented):
Add the scenarios which are being validated.  Validate with and without high log scale mode to ensure both modes works fine.

ganga1980 (approved):

PR #1489, opened 2025-06-17
geneva identity

ganga1980:
geneva identity being used to load balance logs to different monikers.  This needs to be unique so that agent will load balances the correctly.

### DIFF
--- a/Documentation/Internal/ContainerLogV2-Linux.xml
+++ b/Documentation/Internal/ContainerLogV2-Linux.xml
@@ -8,9 +8,9 @@
       <IdentityComponent name="ClusterResourceId" envariable="AKS_RESOURCE_ID"/>
       <IdentityComponent name="ClusterRegion" envariable="AKS_REGION"/>
       <IdentityComponent name="Computer" envariable="HOSTNAME"/>
-      <IdentityComponent name="Tenant">TENANT_ID</IdentityComponent>
-      <IdentityComponent name="Role">ROLE_ID</IdentityComponent>
-      <IdentityComponent name="RoleInstance">ROLEINSTANCE_ID</IdentityComponent>
+      <IdentityComponent name="Tenant" envariable="AKS_RESOURCE_ID"/>
+      <IdentityComponent name="Role" envariable="AKS_REGION"/>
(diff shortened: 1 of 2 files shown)


Documentation/Internal/ContainerLogV2-Linux.xml
```diff
@@ -8,9 +8,9 @@
       <IdentityComponent name="ClusterResourceId" envariable="AKS_RESOURCE_ID"/>
       <IdentityComponent name="ClusterRegion" envariable="AKS_REGION"/>
       <IdentityComponent name="Computer" envariable="HOSTNAME"/>
-      <IdentityComponent name="Tenant">TENANT_ID</IdentityComponent>
-      <IdentityComponent name="Role">ROLE_ID</IdentityComponent>
-      <IdentityComponent name="RoleInstance">ROLEINSTANCE_ID</IdentityComponent>
+      <IdentityComponent name="Tenant" envariable="AKS_RESOURCE_ID"/>
```
suyadav1, line 11: FMI: what does this change mean? We will be using AKS_RESOURCE_ID as TenantId?

wanlonghenry, line 11: why this is using the resource_id?  it already has <IdentityComponent name="ClusterResourceId" envariable="AKS_RESOURCE_ID"/>  either we do not need Tenant or use the right tenant id from AAD

ganga1980: @suyadav1  - This is internal documentation (just for reference)  and not used in our code as is. This is Geneva config used by internal 1P teams on their Geneva accounts in Jarvis.  1P customers has to define this identity config on their Geneva logs account side in Jarvis,  and MDSD downloads this config in Geneva mode and uses load balances the Geneva logs across different storage account monikers. Please note that, this config change doesnt have any impact.

ganga1980: > why this is using the resource_id? it already has either we do not need Tenant or use the right tenant id from AAD

@wanlonghenry  - This can be any unique identifier and doesnt need to be AAD tenant id.

suyadav1 (approved):

PR #1518, opened 2025-08-25
Cert mount fix for AGC

suyadav1:
Cert mount fix for AGC for Mariner 3.
The OS host name is changed from "mariner" to "azurelinux" in Mariner 3. Since we already rolled out on the public cloud, we don't need the "mariner" check anymore.

Test notes:
The change only impacts AGC so we will toggle the test sub in AGC and validate the fix.

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -675,7 +675,7 @@ if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ] || [ "
     if [ -e "/etc/ama-logs-secret/PROXYCERT.crt" ]; then
       cp /etc/ama-logs-secret/PROXYCERT.crt /etc/pki/ca-trust/source/PROXYCERT.crt
     fi
-    update-ca-trust     
+    update-ca-trust
   else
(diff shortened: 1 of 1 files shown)


kubernetes/linux/main.sh
```diff
@@ -678,7 +678,7 @@ if [ $CLOUD_ENVIRONMENT == "usnat" ] || [ $CLOUD_ENVIRONMENT == "ussec" ] || [ "
     update-ca-trust
   else
     if [ $OS_ID != "ubuntu" ]; then
-      echo "Error: The ID in /etc/os-release is not ubuntu or mariner. Defaulting to ubuntu."
+      echo "Error: The ID in /etc/os-release is not ubuntu or azurelinux. Defaulting to ubuntu."
```
ganga1980: we should default to Mariner  instead of ubuntu (since we dont use ubuntu base image any more) incase if this is going to change in the future.

ganga1980 (commented):
left comment

kubernetes/linux/main.sh
suyadav1: done.

ganga1980 (approved):

PR #1527, opened 2025-09-23
Enabled OTel logs and traces support

suyadav1:
This pull request introduces support for OpenTelemetry logs in the Azure Monitor Agent (AMA) for Linux, alongside improvements to telemetry reporting. The changes span environment variable management, agent startup logic, liveness probe checks, Kubernetes manifest updates, and telemetry dimension reporting.

**OpenTelemetry Logs Support**

* Added logic in `main.sh` to detect if OpenTelemetry logs are enabled, set relevant environment variables (`IS_OPENTELEMETRY_LOGS_ENABLED`, `OPENTELEMETRY_LOGS_PORT`, `OPENTELEMETRY_LOGS_CONTAINER_PORT`), and start the core agent with appropriate configuration for OpenTelemetry logs. [[1]](diffhunk://#diff-16c72639c92748bb0587fea9d4bee32127bd5b252e2138fea83433a43c0a5f65R86-R105) [[2]](diffhunk://#diff-16c72639c92748bb0587fea9d4bee32127bd5b252e2138fea83433a43c0a5f65R533-R547) [[3]](diffhunk://#diff-16c72639c92748bb0587fea9d4bee32127bd5b252e2138fea83433a43c0a5f65R255-R262) [[4]](diffhunk://#diff-16c72639c92748bb0587fea9d4bee32127bd5b252e2138fea83433a43c0a5f65L1091-R1136)
* Updated Kubernetes manifest (`ama-logs.yaml`) to add environment variables and container ports for OpenTelemetry logs. [[1]](diffhunk://#diff-36d02cbde2b90a572c51c987289a1cfdf4726779cefbd512e2d1663ecf2c8dd4R457-R458) [[2]](diffhunk://#diff-36d02cbde2b90a572c51c987289a1cfdf4726779cefbd512e2d1663ecf2c8dd4R479-R482)
* Modified liveness probe script to check for agent presence when OpenTelemetry logs are enabled and exit with a specific message if not running.

**Telemetry and Monitoring Enhancements**

* Extended Go telemetry plugin to report OpenTelemetry logs mode and associated port/container port dimensions in metrics.

**Agent and Dependency Updates**

* Upgraded `azure-mdsd` package version from 1.35.7 to 1.37.0 in setup script for improved agent capabilities.

**Testing:**
Following scenarios were tested:
1. Count validation - matches the exact number of spans generated in the workspace.
2. Amd64 and Arm64 node support.
3. Supports only http/protobuf. Doesn't support http/json and gRPC.
4. Doesn't support compression.
5. Data sanity test. All the SpanIds and TraceIds are generating in proper format.
6. Data is getting populated in all tables.
7. AMACA is not throwing 1MB payload issue.
8. Doesn't crash in case of non-supported format/protocols.
9. Telemetry is emitting properly.
10. Supports upto 50k EPS with additional 250 Mi memory usage and 0.5 CPU. Below are the memory and CPU usage details per the OTLP load:
image
Scenario 1: ciprod:3.1.29 memory usage<br>
Scenario 2: Deployed OTLP image with otlp flag disabled<br>
Scenario 3: Deployed OTLP image with otlp flag enabled but no otlp traffic<br>
Scenario 4: Triggered 100 EPS traffic<br>
Scenario 5: Triggered 1k EPS<br>
Scenario 6: Triggered 10k EPS<br>
Scenario 7: Triggered 50k EPS<br>
Scenario 8: No otlp traffic but memory usage stays up<br>
Scenario 9: Triggered 50k EPS again<br>
Scenario 10: Triggered 85k EPS.

image


kubernetes/ama-logs.yaml
```diff
@@ -1093,10 +1099,6 @@ spec:
             #   value: "true"
             - name: APPMONITORING_AUTOINSTRUMENTATION_ENABLED
```
wanlonghenry, line 1100: is this still used in rs pods?

kubernetes/linux/main.sh
```diff
@@ -494,10 +524,25 @@ source common_agent_config_env_var
...
+      echo "export IS_OPENTELEMETRY_LOGS_ENABLED=$IS_OPENTELEMETRY_LOGS_ENABLED" >>~/.bashrc
+      if [ -n "${APPMONITORING_OPENTELEMETRYLOGS_PORT}" ]; then
+            export OPENTELEMETRY_LOGS_PORT="${APPMONITORING_OPENTELEMETRYLOGS_PORT}"
+            echo "export OPENTELEMETRY_LOGS_PORT=${OPENTELEMETRY_LOGS_PORT}" >>~/.bashrc
+      fi
+      if [ -n "${AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT}" ]; then
+            export OPENTELEMETRY_LOGS_CONTAINER_PORT="${AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT}"
+            echo "export OPENTELEMETRY_LOGS_CONTAINER_PORT=${OPENTELEMETRY_LOGS_CONTAINER_PORT}" >>~/.bashrc
+      fi
+      source ~/.bashrc
```
wanlonghenry: this step can be grouped

source/plugins/go/src/telemetry.go
```diff
@@ -292,6 +292,21 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 					telemetryDimensions["isHighLogScaleMode"] = isHighLogScaleMode
 				}
 
+				isOpenTelemetryLogsEnabled := os.Getenv("IS_OPENTELEMETRY_LOGS_ENABLED")
+				if isOpenTelemetryLogsEnabled != "" {
+					telemetryDimensions["isOpenTelemetryLogsEnabled"] = isOpenTelemetryLogsEnabled
```
ganga1980: since our telemetry volume  is huge and lets optimize as we add more telemetry,  instead of "isOpenTelemetryLogsEnabled", use "OtelLogs" : <true | false>. similarly openTelemetryLogsPort use "OtelLogsPort"

source/plugins/go/src/telemetry.go
```diff
@@ -292,6 +292,21 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
...
+					if strings.EqualFold(isOpenTelemetryLogsEnabled, "true") {
+						openTelemetryPort := os.Getenv("OPENTELEMETRY_LOGS_PORT")
+						if openTelemetryPort != "" {
+							telemetryDimensions["openTelemetryLogsPort"] = openTelemetryPort
+						}
+						openTelemetryContainerPort := os.Getenv("OPENTELEMETRY_LOGS_CONTAINER_PORT")
+						if openTelemetryContainerPort != "" {
+							telemetryDimensions["openTelemetryLogsContainerPort"] = openTelemetryContainerPort
+						}
```
ganga1980, line 308: since this is not configurable, I dont think we would need this.

source/plugins/go/src/telemetry.go
```diff
@@ -292,6 +292,21 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 					telemetryDimensions["isHighLogScaleMode"] = isHighLogScaleMode
 				}
 
+				isOpenTelemetryLogsEnabled := os.Getenv("IS_OPENTELEMETRY_LOGS_ENABLED")
+				if isOpenTelemetryLogsEnabled != "" {
+					telemetryDimensions["isOpenTelemetryLogsEnabled"] = isOpenTelemetryLogsEnabled
+					if strings.EqualFold(isOpenTelemetryLogsEnabled, "true") {
+						openTelemetryPort := os.Getenv("OPENTELEMETRY_LOGS_PORT")
+						if openTelemetryPort != "" {
```
ganga1980: 1. what's the openTelemetryPort  empty means? 2. we can collect telemetry if its not default port, right?

source/plugins/go/src/telemetry.go
```diff
@@ -292,6 +292,21 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 					telemetryDimensions["isHighLogScaleMode"] = isHighLogScaleMode
 				}
 
+				isOpenTelemetryLogsEnabled := os.Getenv("IS_OPENTELEMETRY_LOGS_ENABLED")
```
ganga1980: There are will be customers who directly ingests to our endpoint without app monitoring webhook, hence we should collect the collect telemetry about whether app monitoring webhook enabled or not. primarily, we would need telemetry when its not enabled.

kubernetes/linux/setup.sh
```diff
@@ -37,7 +37,7 @@ rm -rf /usr/lib/ruby/gems/3.3.0/gems/rdoc-6.6.3.1
 # remove net-imap gem as it has a known CVE (CVE-2025-43857) and is not used by the agent
 gem uninstall net-imap --force
 
-sudo tdnf install -y azure-mdsd-1.35.7
+sudo tdnf install -y azure-mdsd-1.37.0
```
ganga1980, line 40: Please make sure to update to official version and also review the changes between the current version and updated version to ensure there are no regressions.

kubernetes/linux/main.sh
```diff
@@ -79,7 +81,27 @@ startAMACoreAgent() {
          echo "export PA_CONFIG_PORT=$PA_CONFIG_PORT"
          echo "export CounterDataReportFrequencyInMinutes=$CounterDataReportFrequencyInMinutes"
          echo "export AMACALogFilePath=$AMACALogFilePath"
-      } >> ~/.bashrc
+      } >> ~/.bashrc  
+
+      if isOpenTelemetryLogsEnabled; then
+            echo "AMACoreAgent: Starting AMA Core Agent for Open Telemetry logs(and traces)"
```
ganga1980: lets merge this log message with "AMACoreAgent: Starting AMA Core Agent since High Log scale mode is enabled" since AMACA can be enabled either OpenTelemetry and/or highlogscale enabled

kubernetes/linux/main.sh
```diff
@@ -79,7 +81,27 @@ startAMACoreAgent() {
...
+            export PA_AMCS_PROTOCOL="HttpProtobuf"
+            export PA_AMCS_HOST="0.0.0.0"
+            export PA_AMCS_PORT=4319
+            if [ -n "${AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT}" ]; then
+                  # Convert AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT from string to int
+                  port_int=$((AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT + 0))
+                  if [ "$port_int" -gt 0 ] 2>/dev/null; then
+                        export PA_AMCS_PORT=$port_int
+                  fi
```
ganga1980, line 93: why do we need this logic? isnt that export PA_AMCS_PORT=$(AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT) ?

kubernetes/linux/main.sh
```diff
@@ -494,10 +524,25 @@ source common_agent_config_env_var
...
+      export IS_HIGH_LOG_SCALE_MODE=true
+      echo "export IS_HIGH_LOG_SCALE_MODE=$IS_HIGH_LOG_SCALE_MODE" >>~/.bashrc
+      source ~/.bashrc
+fi
+
+if isOpenTelemetryLogsEnabled; then
+      echo "Enabled OpenTelemetry Logs Mode"
+      export IS_OPENTELEMETRY_LOGS_ENABLED=true
+      echo "export IS_OPENTELEMETRY_LOGS_ENABLED=$IS_OPENTELEMETRY_LOGS_ENABLED" >>~/.bashrc
+      if [ -n "${APPMONITORING_OPENTELEMETRYLOGS_PORT}" ]; then
```
ganga1980: will be there any scenario where this will be empty? if yes, we should log error message.

kubernetes/ama-logs.yaml
```diff
@@ -454,6 +454,8 @@ spec:
               value: "false"
             - name: APPMONITORING_OPENTELEMETRYLOGS_PORT
               value: "28331"
+            - name: AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT
```
ganga1980, line 457: make the yaml changes consistent with AKS yaml with regards to open telemetry feature.

build/linux/installer/scripts/livenessprobe.sh
```diff
@@ -71,12 +71,21 @@ then
...
+  fi
+fi
+
+if [[ "${CONTROLLER_TYPE}" == "DaemonSet" && "${CONTAINER_TYPE}" != "PrometheusSidecar" && "${IS_OPENTELEMETRY_LOGS_ENABLED}" == "true" ]]; then
+  (ps -ef | grep "amacoreagent" | grep -v "grep")
+  if [ $? -ne 0 ]
+  then
+    echo "amacoreagent is not running when OpenTelemetry logs are enabled" > /dev/termination-log
+    exit 1
+  fi
```
ganga1980, line 79: Both can be merged. just you would need to add  || "${IS_OPENTELEMETRY_LOGS_ENABLED}" == "true" to the existing condition and enhance the log message echo "amacoreagent is not running when high log scale mode and/or opentelemetry is enabled"

kubernetes/linux/setup.sh
ganga1980, line 40: do we need to increase our linux daemon set resource limits to accommodate the OTEL feature?

ganga1980, line 40: can you list the scenarios which are being validated?  how many DCR-A's supported? will there any bottle necks if high scale, network flow logs and otel logs enabled?  For example, what happens if customer emits the logs and traces with compression or emits telemetry signals with different variant of http protocols for example http/1, http/1.1 and http/2  and http/json etc.

source/plugins/go/src/telemetry.go
ganga1980: 1. how do we troubleshoot if customer reports issue stating logs and traces are missing for certain namespaces ?
2.  does AMACA drops the logs/traces in any scenario for example payload higher than limits. for example, customer has client side batching or lot of trace/span attributes.
3.  will there be limit on number of apps in a node concurrently emitting telemetry signals?

kubernetes/linux/setup.sh
ganga1980: lets enumerate what scenarios are supported and what's scenarios not supported? does it work with ARM64, http proxy, private link, dual stack clusters etc.

ganga1980 (changes requested):
left comments

kubernetes/linux/main.sh
suyadav1: We have added AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT variable as a string in helm chart so converting it to int here.

kubernetes/ama-logs.yaml
suyadav1: Yes, the similar changes are applied here.

build/linux/installer/scripts/livenessprobe.sh
suyadav1: Done.

kubernetes/ama-logs.yaml
suyadav1: this is still part of helm chart. We can plan to remove it in the next iteration.

source/plugins/go/src/telemetry.go
suyadav1: Didn't get it. Which endpoint are we talking about here? As of now, the otel feature can be only be enabled via app monitoring agent
```
- name: APPMONITORING_OPENTELEMETRYLOGS_ENABLED
 value: "{{ $.Values.AppmonitoringAgent.isOpenTelemetryLogsEnabled | default false }}"
```

suyadav1: 1. We are tailing the amacoreagent logs file for any errors so we should be able to investigate if there are any failures/error in logs processing.
2. AMACA has a limit of 1MB payload as of now. It collects the logs/traces and creates a batch of less than 1MB payload. We saw this bug earlier where some logs were failing to process because of 1MB limit but now it's resolved. Are you saying each logs/traces payload can be more than 1MB? 
3. There is no limit on number of apps.

source/plugins/go/src/telemetry.go
suyadav1: done.

source/plugins/go/src/telemetry.go
suyadav1: openTelemtryPort is the node port

kubernetes/linux/main.sh
suyadav1: done.

kubernetes/linux/main.sh
suyadav1: This shouldn't be empty if the open telemetry logs support is enabled. Removed the check.

source/plugins/go/src/telemetry.go
suyadav1: Added logic to collect telemetry only when the used port is not the default one.

source/plugins/go/src/telemetry.go
suyadav1: Added app instrumentation flag in the telemetry only when it's disabled but otlp support is enabled

kubernetes/linux/setup.sh
suyadav1: will update in the description for testing completed. Pending tasks like max number of DCR-A support, high scale + network flow logs + otel logs support, http proxy, private link and dual stack clusters will be completed in the next iteration.

source/plugins/go/src/telemetry.go
suyadav1: As discussed, I will double check the behavior with the AMACA team in above scenarios.

kubernetes/linux/setup.sh
suyadav1: I have added the resource usage analysis in the description. We are planning to toggle the memory usage if needed for otlp users in the first iteration.

ganga1980: @suyadav1 - My ask is not add the internal link here. rather review the release notes and see if there are any potential regressions or the areas we would need to focus as part of the validation. lets remove the internal link since this code is public.

ganga1980 (approved):

kubernetes/linux/setup.sh
suyadav1: Missed adding that I have reviewed the release notes, and I don't see any areas to focus on other than the otel changes.

PR #1543, opened 2025-10-14
Gangams/workload identity geneva

ganga1980:
Add Microsoft Entra workload identity support for ama-logs-geneva pods which will be running in application logs k8s namespace which is owned and deployed by the customer. This is reference yaml changes for customer to follow.

Following scenarios validated .. 

-  E2E validation with both Microsoft Entra workload identity and existing managed identity
- validated the upgrade scenarios from workload identity to managed identity and vice versa.


charts/azuremonitor-containers-geneva/values.yaml
```diff
@@ -5,13 +5,15 @@
 replicaCount: 1
 
 genevaLogsConfig:
+ authidType: "AuthMSIToken"                                                                  # supported values AuthMSIToken or AuthWorkloadIdentity and recommended to use AuthWorkloadIdentity
```
wanlonghenry, line 8: do we have plan to migrate to default AuthWorkloadIdentity as it is recommended?

ganga1980: currently internal teams are using MSI auth and we would need them to migrate to workload identity. For workload identity, customers need to migrate their AKS clusters to AKS workload identity cluster - https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster.

Yes, at somepoint, we would make "AuthWorkloadIdentity" as default. Please note that, this is just reference implementation for internal customers to refer.

wanlonghenry (approved):

PR #1545, opened 2025-10-22
OTLP optimize telemetry

suyadav1:
This PR optimizes OTLP related telemetry by converting the logs to metrics for following scenarios:
1. Invalid config - compression is enabled.
2. Unsupported protocol
3. GIG LA token error
4. Logs Published and received EPS.

Metrics Output:
image

Verified that no otlp errors are getting logged in traces.

### DIFF
--- a/build/linux/installer/conf/fluent-bit-common.conf
+++ b/build/linux/installer/conf/fluent-bit-common.conf
@@ -103,12 +103,6 @@
     #  - "Will retry one more time in 5 seconds." retry notice (does not include AMACoreAgent keyword)
     Exclude log (OtlpTokenFetcher.*AMACoreAgent tenant not started|CreateSocket.*AMACoreAgent|ConnectToPipelineAgent.*AMACoreAgent|PAConnector::StartTenant.*AMACoreAgent|PAConnector::SendAmcsData.*AMACoreAgent|PAConnector::AmcsStopTenant.*AMACoreAgent|Will retry one more time in 5 seconds\.)
 
-[FILTER]
-    Name grep
-    Alias ama-logs_amaca_log_tail
-    Match oms.container.log.flbplugin.amaca.*
-    Exclude log Information
-
 [OUTPUT]
     Name                            oms
     Alias                           oms_output
--- a/source/plugins/go/src/telemetry.go
+++ b/source/plugins/go/src/telemetry.go
@@ -738,12 +738,17 @@ func PushToAppInsightsTraces(records []map[interface{}]interface{}, severityLeve
 			UpdateTracesErrorMetrics("AddonTokenAdapterErrorModifyingIptableRules")
 		} else if strings.Contains(logEntry, "Token last updated at") && strings.Contains(logEntry, "exiting the container") {
 			UpdateTracesErrorMetrics("AddonTokenAdapterExitContainerTokenNotUpdated")
-		} else if strings.Contains(logEntry, "Google.Protobuf") || strings.Contains(logEntry, "OtlpPipeline.HttpListenerConfigService") {
-			UpdateTracesErrorMetrics("OtlpInvalidProtocol")
-		} else if strings.Contains(logEntry, "GigOtlpDataOutput") && strings.Contains(logEntry, "Event:Log") {
-			UpdateTracesInfoMetrics("OtlpLogsEPS", logEntry)
-		} else if strings.Contains(logEntry, "GigOtlpDataOutput") && strings.Contains(logEntry, "Event:Span") {
-			UpdateTracesInfoMetrics("OtlpSpansEPS", logEntry)
+		} else if strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") {
+			if strings.Contains(logEntry, "Google.Protobuf") || strings.Contains(logEntry, "OtlpPipeline.HttpListenerConfigService") {
+				// This error occurs when the OTLP receiver receives data in an unsupported protocol format or compression is enabled.
+				UpdateTracesErrorMetrics("OtlpInvalidProtocol")
+			} else if strings.Contains(logEntry, "ContentType application/json not supported") {
+				UpdateTracesErrorMetrics("OtlpInvalidContentType")
+			} else if strings.Contains(logEntry, "GigOtlpDataOutput") && strings.Contains(logEntry, "Event:Log") {
+				UpdateTracesInfoMetrics("OtlpLogsEPS", logEntry)
+			} else if strings.Contains(logEntry, "GigOtlpDataOutput") && strings.Contains(logEntry, "Event:Span") {
+				UpdateTracesInfoMetrics("OtlpSpansEPS", logEntry)
+			}
 		} else {
 			if !strings.Contains(tag, "addon-token-adapter") && !(strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") && strings.Contains(logEntry, "Information")) {
 				logLines = append(logLines, logEntry)
--- a/source/plugins/go/src/telemetry.go
+++ b/source/plugins/go/src/telemetry.go
@@ -105,7 +105,7 @@ var (
 	// MultitenantNamespaceCount indicates the number of unique k8s namespaces enabled for multi-tenancy
 	MultitenantNamespaceCount int
 	// Regex to capture Received and Published EPS values from GigOtlpDataOutput log lines
-	otlpEPSRegex = regexp.MustCompile(`Received EPS:(\d+(?:\.\d+)?),\s*Published EPS:(\d+(?:\.\d+)?)`)
+	OtlpEPSRegex = regexp.MustCompile(`Received EPS:(\d+(?:\.\d+)?),\s*Published EPS:(\d+(?:\.\d+)?)`)
 )
 
 const (
@@ -456,19 +456,26 @@ func SendTracesAsMetrics(telemetryPushIntervalProperty string) {
 	TracesMetricsTicker = time.NewTicker(time.Second * time.Duration(telemetryPushInterval))
 
 	for ; true; <-TracesMetricsTicker.C {
+		// Capture and clear error metrics atomically
 		TracesErrorMetricsMutex.Lock()
-		for metricName, metricValue := range TracesErrorMetrics {
-			TelemetryClient.Track(appinsights.NewMetricTelemetry(metricName, metricValue))
-		}
+		errorMetricsSnapshot := TracesErrorMetrics
 		TracesErrorMetrics = map[string]float64{}
 		TracesErrorMetricsMutex.Unlock()
 
-		TracesInfoMetricsMutex.Lock()
-		for metricName, metricValue := range TracesInfoMetrics {
+		// Send metrics outside the lock
+		for metricName, metricValue := range errorMetricsSnapshot {
 			TelemetryClient.Track(appinsights.NewMetricTelemetry(metricName, metricValue))
 		}
+
+		// Same pattern for info metrics
+		TracesInfoMetricsMutex.Lock()
+		infoMetricsSnapshot := TracesInfoMetrics
 		TracesInfoMetrics = map[string]float64{}
 		TracesInfoMetricsMutex.Unlock()
+
+		for metricName, metricValue := range infoMetricsSnapshot {
+			TelemetryClient.Track(appinsights.NewMetricTelemetry(metricName, metricValue))
+		}
 	}
 }
 
@@ -686,7 +693,7 @@ func UpdateTracesInfoMetrics(key string, logEntry string) {
 
 // extractOtlpEPS returns Received and Published EPS values when present in a log entry
 func extractOtlpEPS(logEntry string) (float64, float64, bool) {
-	matches := otlpEPSRegex.FindStringSubmatch(logEntry)
+	matches := OtlpEPSRegex.FindStringSubmatch(logEntry)
 	if len(matches) != 3 {
 		return 0, 0, false
 	}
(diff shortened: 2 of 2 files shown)


source/plugins/go/src/telemetry.go
```diff
@@ -90,14 +92,20 @@ var (
...
+	//Metrics map for traces information
+	TracesInfoMetrics = map[string]float64{}
+	//Mutex for traces information metrics
+	TracesInfoMetricsMutex = &sync.Mutex{}
 	// ContainerLogV2ExtensionDCRCount indicates the number of ContainerLogV2 extension DCRs
 	ContainerLogV2ExtensionDCRCount int
 	// MultitenantNamespaceCount indicates the number of unique k8s namespaces enabled for multi-tenancy
 	MultitenantNamespaceCount int
+	// Regex to capture Received and Published EPS values from GigOtlpDataOutput log lines
+	otlpEPSRegex = regexp.MustCompile(`Received EPS:(\d+(?:\.\d+)?),\s*Published EPS:(\d+(?:\.\d+)?)`)
```
ganga1980: nit: use same naming convention as other variables to be consistent.

build/linux/installer/conf/fluent-bit-common.conf
```diff
@@ -103,12 +103,6 @@
     #  - "Will retry one more time in 5 seconds." retry notice (does not include AMACoreAgent keyword)
     Exclude log (OtlpTokenFetcher.*AMACoreAgent tenant not started|CreateSocket.*AMACoreAgent|ConnectToPipelineAgent.*AMACoreAgent|PAConnector::StartTenant.*AMACoreAgent|PAConnector::SendAmcsData.*AMACoreAgent|PAConnector::AmcsStopTenant.*AMACoreAgent|Will retry one more time in 5 seconds\.)
 
-[FILTER]
```
ganga1980, line 106: currently, we are ingesting all kinds of errors from amaca. with this change, there wont be any  error traces related to non-OTLP, right?

source/plugins/go/src/telemetry.go
```diff
@@ -684,16 +740,31 @@ func PushToAppInsightsTraces(records []map[interface{}]interface{}, severityLeve
 			UpdateTracesErrorMetrics("AddonTokenAdapterErrorModifyingIptableRules")
 		} else if strings.Contains(logEntry, "Token last updated at") && strings.Contains(logEntry, "exiting the container") {
 			UpdateTracesErrorMetrics("AddonTokenAdapterExitContainerTokenNotUpdated")
+		} else if strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") {
+			if strings.Contains(logEntry, "Google.Protobuf") || strings.Contains(logEntry, "OtlpPipeline.HttpListenerConfigService") {
+				// This error occurs when the OTLP receiver receives data in an unsupported protocol or compression is enabled.
+				UpdateTracesErrorMetrics("OtlpInvalidConfig")
+			} else if matched, _ := regexp.MatchString(`ContentType .* not supported`, logEntry); matched {
```
ganga1980, line 754: These conditions can be non OTLP also, right? But, we are ingesting as OtlpInvalidContentType and OtlpInvalidToken.

source/plugins/go/src/telemetry.go
```diff
@@ -684,16 +740,31 @@ func PushToAppInsightsTraces(records []map[interface{}]interface{}, severityLeve
...
+			} else if strings.Contains(logEntry, "GigLA Token not available") {
+				UpdateTracesErrorMetrics("OtlpInvalidToken")
+			} else if strings.Contains(logEntry, "GigOtlpDataOutput") && strings.Contains(logEntry, "Event:Log") {
+				UpdateTracesInfoMetrics("OtlpLogsEPS", logEntry)
+			} else if strings.Contains(logEntry, "GigOtlpDataOutput") && strings.Contains(logEntry, "Event:Span") {
+				UpdateTracesInfoMetrics("OtlpSpansEPS", logEntry)
+			}
 		} else {
-			if !strings.Contains(tag, "addon-token-adapter") {
+			if !strings.Contains(tag, "addon-token-adapter") && !(strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") && strings.Contains(logEntry, "Information")) {
```
ganga1980: why we would need this - !(strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") && strings.Contains(logEntry, "Information"))  ? is it for amaca and error logs?

source/plugins/go/src/telemetry.go
```diff
@@ -445,15 +453,22 @@ func SendTracesAsMetrics(telemetryPushIntervalProperty string) {
 		telemetryPushInterval = defaultTelemetryPushIntervalSeconds
 	}
 
-	TracesErrorMetricsTicker = time.NewTicker(time.Second * time.Duration(telemetryPushInterval))
+	TracesMetricsTicker = time.NewTicker(time.Second * time.Duration(telemetryPushInterval))
 
-	for ; true; <-TracesErrorMetricsTicker.C {
+	for ; true; <-TracesMetricsTicker.C {
```
ganga1980, line 458: This can be optimized to reduce the critical section - 

     for ; true; <-TracesMetricsTicker.C {
        // Capture and clear error metrics atomically
        TracesErrorMetricsMutex.Lock()
        errorMetricsSnapshot := TracesErrorMetrics
        TracesErrorMetrics = make(map[string]float64)
        TracesErrorMetricsMutex.Unlock()
    
        // Send metrics outside the lock
        for metricName, metricValue := range errorMetricsSnapshot {
            TelemetryClient.Track(appinsights.NewMetricTelemetry(metricName, metricValue))
        }
    
        // Same pattern for info metrics
        TracesInfoMetricsMutex.Lock()
        infoMetricsSnapshot := TracesInfoMetrics
        TracesInfoMetrics = make(map[string]float64)
        TracesInfoMetricsMutex.Unlock()
    
        for metricName, metricValue := range infoMetricsSnapshot {
            TelemetryClient.Track(appinsights.NewMetricTelemetry(metricName, metricValue))
        }
    }

source/plugins/go/src/telemetry.go
suyadav1: I don't think we will get similar errors in amaca for highLogScale or NetworkFlowLogs. If it's possible, we can change it.

build/linux/installer/conf/fluent-bit-common.conf
suyadav1: Didn't get it. This filter was removing info logs from AMACA file but we need it to parse the EPS. We are dropping them in the telemetry.go file if it's info logs and doesn't contain EPS logs.

source/plugins/go/src/telemetry.go
```diff
@@ -648,6 +663,47 @@ func UpdateTracesErrorMetrics(key string) {
...
+	if strings.Contains(key, "Otlp") && strings.Contains(key, "EPS") {
+		// 2025-09-18T17:23:19.0195013+00:00, amacoreagent, Information, Pid: 243334, Tid: 15] GigOtlpDataOutput: Identifier:dcr-**.gigl-dce-**, Event:Log, Total Received:10762, Total Published:10762, Total Failed:0, Received EPS:81.18, Published EPS:81.18.
+		// 2025-09-18T17:23:19.0246002+00:00, amacoreagent, Information, Pid: 243334, Tid: 15] GigOtlpDataOutput: Identifier:dcr-**.gigl-dce-**, Event:Span, Total Received:39080, Total Published:39080, Total Failed:0, Received EPS:264.30, Published EPS:264.30.
+		if received, published, ok := extractOtlpEPS(logEntry); ok {
+			if existingReceived, ok := TracesInfoMetrics[key+"Received"]; ok {
+				TracesInfoMetrics[key+"Received"] = math.Max(existingReceived, received)
+			} else {
+				TracesInfoMetrics[key+"Received"] = received
+			}
+			if existingPublished, ok := TracesInfoMetrics[key+"Published"]; ok {
```
ganga1980, line 684: 1. how about Failed? 2. dont we need EPS (Received and Published) instead of raw counters?

source/plugins/go/src/telemetry.go
suyadav1: This is to drop the info logs from amaca.

source/plugins/go/src/telemetry.go
suyadav1: We are logging EPS received and published for both logs and spans. If there is any difference between them, that would be the failed logs. We don't want to log total published/failed count as we only care about EPS.

source/plugins/go/src/telemetry.go
```diff
@@ -648,6 +663,47 @@ func UpdateTracesErrorMetrics(key string) {
 	TracesErrorMetricsMutex.Unlock()
 }
 
+func UpdateTracesInfoMetrics(key string, logEntry string) {
+	TracesInfoMetricsMutex.Lock()
+	if strings.Contains(key, "Otlp") && strings.Contains(key, "EPS") {
+		// 2025-09-18T17:23:19.0195013+00:00, amacoreagent, Information, Pid: 243334, Tid: 15] GigOtlpDataOutput: Identifier:dcr-**.gigl-dce-**, Event:Log, Total Received:10762, Total Published:10762, Total Failed:0, Received EPS:81.18, Published EPS:81.18.
+		// 2025-09-18T17:23:19.0246002+00:00, amacoreagent, Information, Pid: 243334, Tid: 15] GigOtlpDataOutput: Identifier:dcr-**.gigl-dce-**, Event:Span, Total Received:39080, Total Published:39080, Total Failed:0, Received EPS:264.30, Published EPS:264.30.
+		if received, published, ok := extractOtlpEPS(logEntry); ok {
```
ganga1980, line 678: Few questions 1. why are we not emitting separate metrics by Event (Log or Span)?

source/plugins/go/src/telemetry.go
```diff
@@ -648,6 +663,47 @@ func UpdateTracesErrorMetrics(key string) {
...
+			if existingReceived, ok := TracesInfoMetrics[key+"Received"]; ok {
```
ganga1980, line 679: 1. why are not using EPS instead of raw values? 2. why do we need max instead of sum here?  are the raw values are monotonically increasing values?

ganga1980 (commented):
left some comments

source/plugins/go/src/telemetry.go
suyadav1: We are.

source/plugins/go/src/telemetry.go
suyadav1: 1. the "key" is OtlpLogsEPS or OtlpSpansEPS so it's covered.
2. We want to log the max EPS for any debugging as sum or avg won't be helpful here.

source/plugins/go/src/telemetry.go
ganga1980: this code wont be hit  since you have **} else if strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") {**

source/plugins/go/src/telemetry.go
ganga1980: I am referring about generic errors - "ContentType .* not supported" and "GigLA Token not available" which are not specific to OTLP.

build/linux/installer/conf/fluent-bit-common.conf
ganga1980: without OTLP scenario, amaca being used for high log scale and network flow logs. If there are any  errors in amaca log file, then those errors will be ingested. This filter excluding Information log messages from AMACA but non Information log messages will be pass through this filter and which are ingested to our telemetry. since you have entire filter, those error traces (for non OTLP scenarios) will not be ingested.

source/plugins/go/src/telemetry.go
ganga1980: you should add this condition !strings.Contains(logEntry, "Information") at the end of the condition under  else if strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") {

build/linux/installer/conf/fluent-bit-common.conf
suyadav1: I have removed this filter which means all the logs will be passed through (including the info and error logs).

source/plugins/go/src/telemetry.go
suyadav1: Good catch, thanks! will update.

ganga1980: This addresses first comment on Filter plugin removal as well.

build/linux/installer/conf/fluent-bit-common.conf
ganga1980: if  you add the condition under } else if strings.Contains(tag, "microsoft.linuxmonagent.amaca.log") { and then this will be addressed as well. Essentially fixing other comment will address this as well.

source/plugins/go/src/telemetry.go
ganga1980: This may not be critical. But as FYI.

build/linux/installer/conf/fluent-bit-common.conf
suyadav1: Fixed it

source/plugins/go/src/telemetry.go
suyadav1: Fixed it.

source/plugins/go/src/telemetry.go
suyadav1: Fixed

ganga1980 (approved):

PR #1551, opened 2025-10-24
Adding change for networkflow logs new stream

wanlonghenry:
Summary
This PR introduces enhancements to the network flow log processing in the Fluent Bit plugin, focusing on improved stream tagging and extension handling for network flow counts.

Changes
Stream Tagging Improvements:

Added a new stream name constant: CONTAINER_NETWORK_LOGS.
Updated logic to first attempt using the new stream name when obtaining the stream tag; falls back to the legacy RETINA_NETWORK_FLOW_LOGS if unavailable.
Extension Field Handling:

Extracts ingress_flow_count, egress_flow_count, and unknown_direction_flow_count from the extensions field and maps them to top-level fields: IngressFlowCount, EgressFlowCount, and UnknownDirectionFlowCount.
Removes these flow count fields from the extensions map when serializing to AdditionalFlowData, ensuring they're not duplicated.
new stream:
image

old stream:
image

### DIFF
--- a/source/plugins/go/src/network_flow_logs.go
+++ b/source/plugins/go/src/network_flow_logs.go
@@ -14,8 +14,11 @@ import (
 	"github.com/tinylib/msgp/msgp"
 )
 
-// Stream name for retina networkflow logs
-const RetinaNetworkFlowLogsStreamName = "RETINA_NETWORK_FLOW_LOGS"
+// Stream names for network flow logs
+const (
+	ContainerNetworkLogsStreamName  = "CONTAINER_NETWORK_LOGS"   // New stream name
+	RetinaNetworkFlowLogsStreamName = "RETINA_NETWORK_FLOW_LOGS" // Legacy stream name
+)
 
 var (
 	// retina networkflow logs stream tag name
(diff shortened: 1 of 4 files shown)


source/plugins/go/src/network_flow_logs.go
```diff
@@ -286,12 +294,28 @@ func mapNetworkFlowLogsToDataMap(dataMap map[string]interface{}, record map[stri
...
 		dataMap["TraceObservationPoint"] = traceObservationPoint
 	}
-	// Packets and Bytes
-	if packetsSent, ok := flow["packets_sent"]; ok {
-		dataMap["PacketsSent"] = safeToInt(packetsSent)
-	}
-	if packetsReceived, ok := flow["packets_received"]; ok {
-		dataMap["PacketsReceived"] = safeToInt(packetsReceived)
+	// Flow counts and packets
+	if MdsdNetworkFlowLogsStreamTagName == getOutputStreamIdTag(ContainerNetworkLogsStreamName, "", &NetworkFlowTagRefreshTracker) {
```
ganga1980: instead of this, you can just use boolean flag based on stream obtained in https://github.com/microsoft/Docker-Provider/blob/1f251a8ae3b3a40a51567fb271b9256da8e3a903/source/plugins/go/src/network_flow_logs.go#L82 , right?

ganga1980 (commented):
left comment

source/plugins/go/src/network_flow_logs.go
wanlonghenry: from acns: "PacketsSent & PacketsReceived were never mapped to anything in the flow object. 
So now instead of them we have the new fields which still will not be mapped. So really only the names of columns are changing.
For AdditionalData in Extensions is it possible to keep that but exclude the three new columns? 
Extensions can contain other fields."
 so removed this check as those fields were not there before

source/plugins/go/src/network_flow_logs.go
```diff
@@ -286,12 +294,17 @@ func mapNetworkFlowLogsToDataMap(dataMap map[string]interface{}, record map[stri
 	if traceObservationPoint := extractString(flow, "trace_observation_point"); traceObservationPoint != "" {
 		dataMap["TraceObservationPoint"] = traceObservationPoint
 	}
-	// Packets and Bytes
-	if packetsSent, ok := flow["packets_sent"]; ok {
-		dataMap["PacketsSent"] = safeToInt(packetsSent)
-	}
-	if packetsReceived, ok := flow["packets_received"]; ok {
```
suyadav1, line 293: If we fall back to the legacy stream, won't we need these?

wanlonghenry: thanks for checking, i double checked with acns, those never populated

suyadav1 (approved):

PR #1589, opened 2026-01-06
Testkube workflow migration

suyadav1:
Testkube support for tests/testsuite/executors is [deprecated ](https://docs.testkube.io/articles/legacy-features) so need to migrate to test workflow. This PR contains following changes:

1. Migrated tests to test workflow.
2. Removed testsuite and handled test sequencing manually.
3. Removed executors since it's not supported anymore. Using the ginkgo image instead as recommended by Testkube.
4. Using the default images for all the dependent components since the policy doesn't block the third-party images anymore. This will keep us safe from CVEs.
5. Safe removal of old testkube installation.


test/testkube/testkube-test-crs.yaml
```diff
@@ -1,151 +1,154 @@
...
+        value: "{{config.GOTOOLCHAIN}}"
+    shell: ginkgo --label-filter "!(windows)" ./livenessprobe
+  pod:
+    nodeSelector:
+      kubernetes.io/os: linux
+      kubernetes.io/arch: amd64
 ---
-apiVersion: tests.testkube.io/v3
-kind: Test
+apiVersion: testworkflows.testkube.io/v1
```
gracewehner, line 109: Thanks Sunil, it looks like from the migration docs that the workflow can run multiple workflows in parallel like a testsuite running multiple tests, so that we don't have to manually in the pipeline:
- https://docs.testkube.io/articles/test-workflow-migration#test-suite-migration
- https://docs.testkube.io/articles/test-workflows-test-suites#execute-syntax

test/testkube/helm-testkube-values.yaml
```diff
@@ -405,22 +405,22 @@ testkube-api:
   # Test Workflows toolkit image parameters
   imageTwToolkit:
     # -- Test Workflows image registry
-    registry: mcr.microsoft.com
+    registry: docker.io
```
rashmichandrashekar, line 408: This will be blocked if the images are being pulled externally.. is there a reason why we can't add them to our mcr?

test/testkube/testkube-test-crs.yaml
suyadav1: Execute is not available in standalone setup:
https://docs.testkube.io/articles/test-workflows-test-suites#execute-syntax
image

test/testkube/helm-testkube-values.yaml
suyadav1: The image pull is not blocked anymore by azure policy. It just throws warning so we should be good. 
Pulling and pushing images to MCR adds extra maintenance since we have to do this every time there is a new version.

rashmichandrashekar: could we pls confirm this? Usually the first few pulls are allowed with a warning and then they are blocked.

test/testkube/testkube-test-crs.yaml
```diff
@@ -1,151 +1,154 @@
...
+kind: TestWorkflow
```
wanlonghenry, line 2: Does this require a specific minimum version of Testkube or Kubernetes to be installed in the cluster? If so, what are the requirements?

test/testkube/helm-testkube-values.yaml
suyadav1: Interesting! Let me try running few more instances.

test/testkube/helm-testkube-values.yaml
```diff
@@ -163,21 +163,22 @@ nats:
...
-  container:
-    image:
-      registry: mcr.microsoft.com
-      repository: azuremonitor/containerinsights/cidev
-      tag: nats_2.10.24-alpine
-      pullPolicy: 
+  # container:
+  #   image:
+  #     repository: nats
+  #     tag: 2.11.6-alpine
```
rashmichandrashekar, line 169: Why are these commented? are they part of the upgrade?

test/testkube/install-and-execute-testkube-tests.sh
```diff
@@ -18,6 +18,25 @@ done
 cluster="$(kubectl config current-context)"
 echo "Current cluster: $cluster"
 
+# Remove stale CRDs that block Helm ownership
+stale_crds=(
+    "testworkflowexecutions.testworkflows.testkube.io"
```
rashmichandrashekar, line 23: are these CRDs installed with every helm install of testkube?

suyadav1: Yes. These CRDs had ownership issues when we migrated Test to TestWorkflow so removed them in the script to handle it for every cluster. Helm install will re-install them as needed.

test/testkube/helm-testkube-values.yaml
suyadav1: Following the helm chart values here. The chart will pull the latest images as per testkube version.

test/testkube/testkube-test-crs.yaml
suyadav1: It's not dependent on Kubernetes version. The support was removed in release: https://docs.testkube.io/changelog#monthly-release-v1211-2025-12-10

test/testkube/testkube-test-crs.yaml
gracewehner: I see :( Thanks Sunil

test/testkube/helm-testkube-values.yaml
suyadav1: Triggered multiple runs and the policy doesn't throw the error anymore.

test/testkube/helm-testkube-values.yaml
suyadav1: As per the discussion, I have pinned the mongodb version. The update mongodb repo is same as what testkube tries to install by default.

rashmichandrashekar (approved):

PR #1590, opened 2026-01-06
Cosmic replicaset Kubernetes API client large scale issue

wanlonghenry:
Summary
This PR enhances the efficiency and reliability of Kubernetes resource fetching and processing. It adds support for streaming and parsing compressed (gzip) JSON payloads, improves continuation token management, and adds extra safeguards for custom metrics processing. The changes primarily affect Kubernetes API client handling, pod inventory processing, and utility checks for custom metrics enablement.

**source/plugins/ruby/in_kube_podinventory.rb**
Adds a check for the availability of custom metrics via CustomMetricsUtils before pushing mdmPodRecord items.
Prevents caching or pushing pod records for custom metrics when they are not enabled or available.

**source/plugins/ruby/KubernetesApiClient.rb**
Refactors getResourcesAndContinuationTokenV2 to stream and efficiently parse API responses, supporting both compressed (gzip) and uncompressed JSON payloads.
Adds logic to handle large API payloads by parsing them on-the-fly using Yajl for reduced memory usage, while smaller payloads are parsed fully in-memory.
Improves error handling, fallback mechanisms, resource usage accounting, and logging for API calls.
Cleanly manages continuation token extraction and propagates it in all parsing scenarios.
Adds new dependencies: yajl, zlib, and stringio for streaming and decompression support.

**source/plugins/ruby/CustomMetricsUtils.rb**
Enhances the check_custom_metrics_availability function: now disables custom metrics if the relevant environment variable is missing or set to an empty string, in addition to 'false'.

Scenaiors:
image
PATH 1: full_gzip
Condition: Response has Content-Encoding: gzip and Content-Length ≤ 256 KB.
image

PATH 2: full_plain
Condition: Response has no gzip (Content-Encoding: not set) and Content-Length ≤ 256 KB.
image

image
PATH 3: stream_gzip
Condition: Response has Content-Encoding: gzip and size is larger than 256 KB (or no Content-Length).
image

PATH 4: stream_plain
Condition: Response is plain (no gzip) and larger than 256 KB (or no Content-Length).
image


source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+              responseCode = response.code
+              unless responseCode == '200'
+                parse_mode = 'error'
+                @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: Non-success code #{responseCode} for #{uri}"
+                # Send telemetry for non-success response codes
+                @@K8sApiResponseTelemetryTimeTracker = ApplicationInsightsUtility.sendAPIResponseTelemetry(responseCode, uri, "K8sAPIStatus", @@K8sApiResponseCodeHash, @@K8sApiResponseTelemetryTimeTracker)
+                break
+              end
+
+              # Decide whether to stream or fallback to full parse based on Content-Length (if small, cheaper to full-parse)
```
zanejohnson-azure, line 908: code branching here, did we test all branches?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                content_length = Integer(response['Content-Length']) if response['Content-Length']
+              rescue; end
+              small_threshold = 256 * 1024 # 256KB
+
+              if content_length && content_length <= small_threshold
+                # Read whole (possibly compressed) body then use faster parser for small payloads
+                body_buf = +"" # mutable string
+                response.read_body { |c| body_buf << c }
+                total_compressed_bytes = body_buf.bytesize
+                if response['Content-Encoding'] == 'gzip'
```
zanejohnson-azure, line 920: another branch

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+              end # streaming path
+            end # http.request
+          end # Net::HTTP.start
+        rescue => inner_err
+          @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: streaming fetch/parse failed for #{uri}: #{inner_err}; falling back to legacy getKubeResourceInfoV2"
+          parse_mode = 'fallback'
+          begin
+            # Fallback to legacy path
+            fallbackResponseCode, resourceInfo = getKubeResourceInfoV2(uri, api_group: api_group)
+            responseCode = fallbackResponseCode if responseCode.nil?
```
zanejohnson-azure: getKubeResourceInfoV2 is used in this branch, why do we check responseCode which is from another branch in next lin?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+            @Log.info "KubernetesApiClient::getResourcesAndContinuationTokenV2(stream): Requesting #{uri} (api_group=#{api_group}) @ #{started_at.iso8601}"
+
+            http.request(kubeApiRequest) do |response|
+              responseCode = response.code
+              unless responseCode == '200'
+                parse_mode = 'error'
+                @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: Non-success code #{responseCode} for #{uri}"
+                # Send telemetry for non-success response codes
+                @@K8sApiResponseTelemetryTimeTracker = ApplicationInsightsUtility.sendAPIResponseTelemetry(responseCode, uri, "K8sAPIStatus", @@K8sApiResponseCodeHash, @@K8sApiResponseTelemetryTimeTracker)
+                break
```
zanejohnson-azure, line 905: why not fallback to legacy path?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                else
+                  parse_mode = 'full_plain'
+                end
+                total_uncompressed_bytes = body_buf.bytesize
+                resourceInventory = JSON.parse(body_buf)
+              else
+                # Streaming path - CRITICAL: Create parser ONCE outside the read_body loop
+                parse_mode = 'stream'
+                is_gzip = (response['Content-Encoding'] == 'gzip')
+                inflater = nil
```
zanejohnson-azure, line 936: why no rescue?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                    parse_mode = 'full_gzip'
+                  rescue => gzerr
+                    @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: gzip decompress(small) failed: #{gzerr}; using compressed body (parse will likely fail)"
+                  end
+                else
+                  parse_mode = 'full_plain'
+                end
+                total_uncompressed_bytes = body_buf.bytesize
+                resourceInventory = JSON.parse(body_buf)
+              else
```
zanejohnson-azure, line 932: if content_length is nil, this path will execute.

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                # Streaming path - CRITICAL: Create parser ONCE outside the read_body loop
+                parse_mode = 'stream'
+                is_gzip = (response['Content-Encoding'] == 'gzip')
+                inflater = nil
+                yajl_parser = nil
+                accumulated_buffer = +""  # Buffer for incomplete JSON chunks
```
zanejohnson-azure: unused var>?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                          serialized_items = JSON.generate(obj['items'])
+                          deep_copied_items = JSON.parse(serialized_items)
+                          parsed_items.concat(deep_copied_items)
+                        rescue => json_err
+                          @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: JSON round-trip failed: #{json_err}, using shallow copy"
+                          parsed_items.concat(obj['items'])
+                        end
+                      end
+                      if obj.key?('metadata') && obj['metadata'].is_a?(Hash)
+                        metadata_continue = obj['metadata']['continue']
```
zanejohnson-azure: why not check key 'contiue' exist or not before using it?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                        resource_version = obj['metadata']['resourceVersion'] if obj['metadata'].key?('resourceVersion')
+                      end
+                    end
+                  end
+
+                  # Stream and parse chunks
+                  chunk_count = 0
+                  response.read_body do |compressed_chunk|
```
zanejohnson-azure, line 971: do we assume reponse body is always compressed?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                    body_buf = Zlib::GzipReader.new(StringIO.new(body_buf)).read
+                    parse_mode = 'full_gzip'
+                  rescue => gzerr
+                    @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: gzip decompress(small) failed: #{gzerr}; using compressed body (parse will likely fail)"
+                  end
+                else
+                  parse_mode = 'full_plain'
+                end
+                total_uncompressed_bytes = body_buf.bytesize
+                resourceInventory = JSON.parse(body_buf)
```
zanejohnson-azure, line 931: json.parse could fail, why not rescue?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                parse_mode = 'stream'
+                is_gzip = (response['Content-Encoding'] == 'gzip')
+                inflater = nil
+                yajl_parser = nil
+                accumulated_buffer = +""  # Buffer for incomplete JSON chunks
+                begin
+                  if is_gzip
+                    # Use Inflate with gzip window bits for streaming
+                    inflater = Zlib::Inflate.new(Zlib::MAX_WBITS + 32)
+                    parse_mode = 'stream_gzip'
```
zanejohnson-azure, line 942: nit: inflater initialization is better placed where it is used at line 978 for readbility and maintainence.

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,190 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                  response.read_body do |compressed_chunk|
+                    chunk_count += 1
+                    total_compressed_bytes += compressed_chunk.bytesize
+
+                    decompressed = if is_gzip
+                      begin
+                        inflater.inflate(compressed_chunk)
+                      rescue Zlib::Error => zerr
+                        @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: gzip inflate failed at chunk #{chunk_count}: #{zerr}"
+                        raise
```
zanejohnson-azure, line 980: infalter will not be cleane up when the resuce happens here.

source/plugins/ruby/KubernetesApiClient.rb
zanejohnson-azure: what is def of an incomplete chunks, and how we know if the chunk is incomplete or not.

source/plugins/ruby/KubernetesApiClient.rb
rashmichandrashekar, line 908: Why do we need this? when we say it is cheaper - is it in terms of resource utilization? also does it affect how quickly the response is parsed? Can't we have stream processing irrespective of the context size?

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: We do fallback to legacy (getKubeResourceInfoV2) in case the streaming fetch/parse fails; see the rescue block around the streaming logic for that path.

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: small payload is the default one and handle gzip scenario; large payload is the steaming one

wanlonghenry: The logic for switching parse strategies is to optimize resource usage: for small payloads, full-parse is faster and uses less memory/CPU compared to setting up streaming. For large payloads, streaming avoids OOM and is more efficient.

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: That branch is exercised when the response body is small and gzip format

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: 'Content-Encoding' header for 'gzip' and only decompress if needed

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: rescue => inner_err will catch this

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: defined here: https://github.com/microsoft/Docker-Provider/blob/0ef85a247a746894619e9266a384fcf333b22604/source/plugins/ruby/KubernetesApiClient.rb#L899

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: this is intended behavior as when Content-Length is unavailable, the code safely defaults to streaming mode rather than attempting to load the entire response into memory, which prevents potential memory issues

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: The Yajl streaming parser maintains an internal buffer and incrementally parses removed unused

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: added

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: moved to ensure

source/plugins/ruby/KubernetesApiClient.rb
zanejohnson-azure: if we break here at line 905, fallback path that starts at line 1025 will not run, right?

source/plugins/ruby/KubernetesApiClient.rb
zanejohnson-azure: there are several scenarios, is_gzip (true of false), content-length (small, or large). because we have no unit test coverage, so we want to test all these branches.

source/plugins/ruby/KubernetesApiClient.rb
zanejohnson-azure: if content_length is nil, then something is wrong. should the code panic? what will be code behavior of streaming mode when content_length is nil? how can streaming mode safely handle this case?

source/plugins/ruby/KubernetesApiClient.rb
zanejohnson-azure: inside this loop, we check value of is_gzip true or false, which means that the compressed_chunk may not be compressed at all, right? the name here is confusing.

source/plugins/ruby/KubernetesApiClient.rb
zanejohnson-azure: if we are in this fall back path, it means something bad happened between 809-1024, the response.code is from the branch in 809-1024, why do we check whether that is nil or not?

since we are in fall back path which will get its own response.code when making kube api call, should it just uses its own. I do not understand why the fall back path need depend on a var between 809 and 1024.

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: The break is used specifically for handling HTTP cases where the response code is not "200". It is not meant to trigger the Ruby exception mechanism

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: the original intend was to log response code from stream first, changed to responseCode = fallbackResponseCode

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: for large payload size we already verified with cosmic; for small payload size we need to validate on ci cd; e2e is better because of dependency and resource leak check

zanejohnson-azure:
> Summary This PR enhances the efficiency and reliability of Kubernetes resource fetching and processing. It adds support for streaming and parsing compressed (gzip) JSON payloads, improves continuation token management, and adds extra safeguards for custom metrics processing. The changes primarily affect Kubernetes API client handling, pod inventory processing, and utility checks for custom metrics enablement.
> 
> **source/plugins/ruby/in_kube_podinventory.rb** Adds a check for the availability of custom metrics via CustomMetricsUtils before pushing mdmPodRecord items. Prevents caching or pushing pod records for custom metrics when they are not enabled or available.
> 
> **source/plugins/ruby/KubernetesApiClient.rb** Refactors getResourcesAndContinuationTokenV2 to stream and efficiently parse API responses, supporting both compressed (gzip) and uncompressed JSON payloads. Adds logic to handle large API payloads by parsing them on-the-fly using Yajl for reduced memory usage, while smaller payloads are parsed fully in-memory. Improves error handling, fallback mechanisms, resource usage accounting, and logging for API calls. Cleanly manages continuation token extraction and propagates it in all parsing scenarios. Adds new dependencies: yajl, zlib, and stringio for streaming and decompression support.
> 
> **source/plugins/ruby/CustomMetricsUtils.rb** Enhances the check_custom_metrics_availability function: now disables custom metrics if the relevant environment variable is missing or set to an empty string, in addition to 'false'.
> 
> Testing: image

could add description how was the test done, and what was verified in this test?

source/plugins/ruby/KubernetesApiClient.rb
```diff
@@ -864,18 +867,187 @@ def getResourcesAndContinuationTokenV2(uri, api_group: nil)
...
+                  if inflater
+                    inflater.finish rescue nil
+                    inflater.close rescue nil
+                  end
+                end
+              end # streaming path
+            end # http.request
+          end # Net::HTTP.start
+        rescue => inner_err
+          @Log.warn "KubernetesApiClient::getResourcesAndContinuationTokenV2: streaming fetch/parse failed for #{uri}: #{inner_err}; falling back to legacy getKubeResourceInfoV2"
```
rashmichandrashekar, line 1026: Hi Long, when we fallback have we made sure we dont send duplicate data?

zanejohnson-azure (approved):

source/plugins/ruby/KubernetesApiClient.rb
wanlonghenry: yes for the fallback 'resourceInventory = JSON.parse(resourceInfo.body)' overwrites anything from a failed stream and the continuationToken is derived from the final inventory making sure that no duplicate date sent.

rashmichandrashekar (approved):

PR #1592, opened 2026-01-06
Zane/fluentdissue

zanejohnson-azure:
cool.io is a dependency of fluentd, the cool.io 1.9.1 has high failure rate on arm64 when compiling it on amd64 using qemu simulator. 

sol: pin cool.io  to 1.8.0 that passed build successfully. note any cool.io version in below range works for fluentd 1.16 .3.

[cool.io](https://rubygems.org/gems/cool.io) >= 1.4.5, < 2.0.0

long-term solution: consider build arm version on arm64 host instead of using qemu on amd host.

### DIFF
--- a/kubernetes/linux/setup.sh
+++ b/kubernetes/linux/setup.sh
@@ -79,12 +79,16 @@ echo "$(fluent-bit --version)" >> packages_version.txt
 # install fluentd
 fluentd_version="1.16.3"
 
-# Pre-install cool.io to avoid ARM64 build issues (segfault during native extension compilation)
 if [ "$ARCH" == "arm64" ]; then
-    gem install cool.io -v "1.8.0" --no-document
+    # Pre-install a fixed version of cool.io to avoid ARM64 build issues.
+    # when upgrading fluentd in the future, check cool.io compatibility first.
+    # https://rubygems.org/gems/fluentd
+    gem install cool.io -v "1.9.0" --no-document
+    gem install fluentd -v $fluentd_version --no-document
+else
+    gem install fluentd -v $fluentd_version --no-document
 fi
 
-gem install fluentd -v $fluentd_version --no-document
 
 # remove the test directory from fluentd
 rm -rf /usr/lib/ruby/gems/3.3.0/gems/fluentd-$fluentd_version/test/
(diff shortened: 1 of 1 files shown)


kubernetes/linux/setup.sh
```diff
@@ -78,7 +78,18 @@ echo "$(fluent-bit --version)" >> packages_version.txt
...
+
+if [ "$ARCH" == "arm64" ]; then
+    # Pre-install a fixed version of cool.io to avoid ARM64 build issues. cool.io 1.9.1 has high failure rate.
+    # when upgrading fluentd in the future, check cool.io compatibility first.
+    # https://rubygems.org/gems/fluentd
+    # long-term solution: consider building arm version on arm host, currently qemu is used on amd64 host.
+    gem install cool.io -v "1.8.0" --no-document
+    gem install fluentd -v $fluentd_version --no-document
+else
+    gem install fluentd -v $fluentd_version --no-document
```
suyadav1, line 90: nit: this can be moved out of if-else block

kubernetes/linux/setup.sh
```diff
@@ -78,7 +78,18 @@ echo "$(fluent-bit --version)" >> packages_version.txt
...
 # install fluentd
 fluentd_version="1.16.3"
-gem install fluentd -v $fluentd_version --no-document
+
+if [ "$ARCH" == "arm64" ]; then
+    # Pre-install a fixed version of cool.io to avoid ARM64 build issues. cool.io 1.9.1 has high failure rate.
+    # when upgrading fluentd in the future, check cool.io compatibility first.
+    # https://rubygems.org/gems/fluentd
+    # long-term solution: consider building arm version on arm host, currently qemu is used on amd64 host.
+    gem install cool.io -v "1.8.0" --no-document
```
wanlonghenry: could you share why ping to this version and why 1.8.0 is stable for ARM?

kubernetes/linux/setup.sh
zanejohnson-azure: i did it literally for maintenance consideration, any engineer who upgrade fluentd in the future will see this note. by putting `gem install fluentd -v $fluentd_version --no-document` in a later section, the cool.io note may be ignored.

rashmichandrashekar (dismissed):

kubernetes/linux/setup.sh
zanejohnson-azure: we use fluentd 1.16.3 for a while, but cool.io version was upgrading overtime in the background whenever a newer cool.io version is available. I tested different version 1.8.0, 1.9.0, 1.9.1, and 1.8.0 is more stable.

PR #1593, opened 2026-01-07
bug fix

ganga1980:
This pull request makes a minor fix to the service records filtering logic in the `in_kube_podinventory` plugin. The change ensures that the correct key is used when checking whether to exclude a Kubernetes service record.

* Updated the call to `KubernetesApiClient.isExcludeResourceItem` to use the `Namespace` key instead of `namespace` when filtering service records in `in_kube_podinventory.rb`.

Fix has been verified.  DCR with include and "kube-system" namespace . 

image

### DIFF
--- a/source/plugins/ruby/in_kube_podinventory.rb
+++ b/source/plugins/ruby/in_kube_podinventory.rb
@@ -408,7 +408,7 @@ module Fluent::Plugin
         if continuationToken.nil? # sending kube services inventory records
           kubeServicesEventStream = Fluent::MultiEventStream.new
           serviceRecords.each do |kubeServiceRecord|
-            next unless !KubernetesApiClient.isExcludeResourceItem(kubeServiceRecord["ServiceName"], kubeServiceRecord["namespace"], @namespaceFilteringMode, @namespaces)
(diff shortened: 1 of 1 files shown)


zanejohnson-azure:
> This pull request makes a minor fix to the service records filtering logic in the `in_kube_podinventory` plugin. The change ensures that the correct key is used when checking whether to exclude a Kubernetes service record.
> 
> * Updated the call to `KubernetesApiClient.isExcludeResourceItem` to use the `Namespace` key instead of `namespace` when filtering service records in `in_kube_podinventory.rb`.
> 
> Fix has been verified.

how was it verified? please share results if you have.

wanlonghenry (approved):

ganga1980:
> > This pull request makes a minor fix to the service records filtering logic in the `in_kube_podinventory` plugin. The change ensures that the correct key is used when checking whether to exclude a Kubernetes service record.
> > 
> > * Updated the call to `KubernetesApiClient.isExcludeResourceItem` to use the `Namespace` key instead of `namespace` when filtering service records in `in_kube_podinventory.rb`.
> > 
> > Fix has been verified.
> 
> how was it verified? please share results if you have.

updated the screenshot before and after fix.

zanejohnson-azure (approved):

PR #1598, opened 2026-01-21
fix: bleu cloud name is azurebleucloud

gracewehner:
The cloud environment coming from AKS is `azurebleucloud` and is also the name mdsd expects. Contract with AKS and mdsd for the name changed at some point. The endpoints are correct however, just the customEnvironment name wasn't valid to mdsd.

```
2026-01-21T07:19:30.2315920Z: Skipped IMDS metadata lookup, will use configured region "bleufrancecentral", environment "bleu"
...
2026-01-21T07:19:30.2326740Z: Error: mdsd is terminated with exception: Unknown cloud environment "bleu", cannot construct MCS or MSI token resource endpoints
2026-01-21T07:19:30.2329030Z: FATAL: mdsd killed by direct call, no signal involved
```

mdsd codebase:
```
const std::string BLEU_CLOUD_NAME = "azurebleucloud";
```

### DIFF
--- a/kubernetes/linux/main.sh
+++ b/kubernetes/linux/main.sh
@@ -6,7 +6,7 @@ startTime=$(date +%s)
 echo "startup script start @ $(date +'%Y-%m-%dT%H:%M:%S')"
 
 # Supported cloud environments
-SUPPORTED_CLOUDS=("azurepubliccloud" "azurechinacloud" "azureusgovernmentcloud" "usnat" "ussec" "bleu")
+SUPPORTED_CLOUDS=("azurepubliccloud" "azurechinacloud" "azureusgovernmentcloud" "usnat" "ussec" "azurebleucloud")
 
 getDomainFromSecret() {
       # Read the domain from the AMA logs secret and convert to lowercase
@@ -51,7 +51,7 @@ getClusterCloudEnvironment() {
                   echo "ussec"
                   ;;
             "opinsights.sovcloud-api.fr")
-                  echo "bleu"
+                  echo "azurebleucloud"
                   ;;
             ""|*)
                   echo "unknown"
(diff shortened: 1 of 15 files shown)


kubernetes/windows/main.ps1
```diff
@@ -327,7 +327,7 @@ function Get-McsGlobalEndpoint{
                 "azureusgovernmentcloud"  { $mcs_globalendpoint = "https://global.handler.control.monitor.azure.us" }
                 "usnat"                   { $mcs_globalendpoint = "https://global.handler.control.monitor.azure.eaglex.ic.gov" }
                 "ussec"                   { $mcs_globalendpoint = "https://global.handler.control.monitor.azure.microsoft.scloud" }
-                "bleu"                    { $mcs_globalendpoint = "https://global.handler.control.monitor.sovcloud-api.fr" }
+                "azurebleucloud"                    { $mcs_globalendpoint = "https://global.handler.control.monitor.sovcloud-api.fr" }
```
wanlonghenry: alignment

kubernetes/linux/main.sh
```diff
@@ -6,7 +6,7 @@ startTime=$(date +%s)
 echo "startup script start @ $(date +'%Y-%m-%dT%H:%M:%S')"
 
 # Supported cloud environments
-SUPPORTED_CLOUDS=("azurepubliccloud" "azurechinacloud" "azureusgovernmentcloud" "usnat" "ussec" "bleu")
```
wanlonghenry, line 9: In RP we have tested cosmic sub under /home/aks-rp/toggles/global/sigs/containerinsights/omsagent-image-tag-linux.yaml  does this change need to be test as well?

gracewehner: Thanks Long, yes we need to toggle cosmic sub for bleu with this change

ganga1980 (approved):

kubernetes/linux/main.sh
suyadav1: how is this working for cosmic as of now with "bleu" env name?

gracewehner: @suyadav1 from @ganga1980:

> I have forwarded email thread about contract between MDSD and AKS, and from our side. Looks like this got changed in AKS RP side as part of this PR - [Pull request 13088295: BLEU (kube-control-plane chart): Use AzureStackCloud and Mount CA Cert from …](https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/13088295?_a=files&path=/ccp/control-plane-core/charts/kube-control-plane/charts/kube-addons/templates/addons/partner-addons/_omsagent.yaml) and MDSD team implemented using different cloud name (for GCS, they have used "Bleu" and for azure monitor, they have implemented "azurebleucloud").  Because of the workspace dependency (in link access check) on bootstrap  scenario was not supported for CI and  E+D cosmic deploying with HELM chart which doesn't use  MSI token adapter and likely MDSD using the domain name.

My understanding is before it was using the non-msi auth helm chart path which accepts the bleu environment name but the msi auth path doesn't

PR #1609, opened 2026-02-27
Add nodes/pods based on k8s version

rashmichandrashekar:
Add nodes/pods based on k8s version

### DIFF
--- a/charts/azuremonitor-containers/templates/ama-logs-rbac.yaml
+++ b/charts/azuremonitor-containers/templates/ama-logs-rbac.yaml
@@ -23,8 +23,17 @@ metadata:
     heritage: {{ .Release.Service }}
 rules:
 - apiGroups: [""]
-  resources: ["pods", "events", "nodes", "nodes/stats", "nodes/metrics", "nodes/spec", "nodes/proxy","namespaces", "services", "persistentvolumes"]
+  resources: ["pods", "events", "nodes", "nodes/stats", "nodes/metrics", "nodes/spec", "namespaces", "services", "persistentvolumes"]
   verbs: ["list", "get", "watch"]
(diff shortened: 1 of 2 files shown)


charts/azuremonitor-containers/templates/ama-logs-rbac.yaml
```diff
@@ -23,8 +23,17 @@ metadata:
     heritage: {{ .Release.Service }}
 rules:
 - apiGroups: [""]
-  resources: ["pods", "events", "nodes", "nodes/stats", "nodes/metrics", "nodes/spec", "nodes/proxy","namespaces", "services", "persistentvolumes"]
+  resources: ["pods", "events", "nodes", "nodes/stats", "nodes/metrics", "nodes/spec", "namespaces", "services", "persistentvolumes"]
   verbs: ["list", "get", "watch"]
+{{- if semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version }}
```
suyadav1, line 28: typo here? it should be 1.33.0?

suyadav1: Also, .

> Capabilities.KubeVersion.Version is populated from the Kubernetes API server’s /version endpoint, returning a semantic version string like v1.28.3

We need to remove 'v' prefix before comparing

wanlonghenry (approved):

wanlonghenry (changes requested):
lgtm also typo in version

charts/azuremonitor-containers/templates/ama-logs-rbac.yaml
rashmichandrashekar: discussed offline - since both are correct

suyadav1 (approved):

wanlonghenry (approved):

PR #1622, opened 2026-03-13
Nicchambers/daemonsetvaluechanges

NicAtMS:
Updated the logs helm chart to support daemonset t-shirt sizing and introduced the new cluster kind which will be replacing isAutomaticSKU. These changes have been tested with fixtures and produce the expected results when t-shirt sizing is enabled and disabled


charts/azuremonitor-containerinsights/templates/_helpers.tpl
```diff
@@ -103,4 +103,136 @@ HOST CA CERTIFICATE MOUNTING SECTION (from AKS chart)
...
+
+{{/* CPU millicores */}}
+{{- else if eq $suffix "m" -}}
+  {{- $result = add (mul $intPart 1) (div $fracPart $fracDivisor) -}}
+
+{{/* No suffix - treat as base unit */}}
+{{- else if eq $suffix "" -}}
+  {{- if contains "." $number -}}
+    {{/* Decimal number without suffix - assume CPU cores, convert to millicores */}}
+    {{- $result = add (mul $intPart 1000) (div (mul $fracPart 1000) $fracDivisor) -}}
```
wanlonghenry, line 210: will this treat cpu 1 as 1 core or 1m? cpu and memory should be considered differently

charts/azuremonitor-containerinsights/templates/_helpers.tpl
```diff
@@ -103,4 +103,136 @@ HOST CA CERTIFICATE MOUNTING SECTION (from AKS chart)
...
+{{/* Handle zero/empty */}}
+{{- if or (eq $quantity "") (eq $quantity "0") -}}
+0
+{{- else -}}
+
+{{/* Extract number and suffix */}}
+{{- $number := "" -}}
+{{- $suffix := "" -}}
+{{- $i := 0 -}}
+{{- range $char := (split "" $quantity) -}}
```
wanlonghenry: split "" is not a robust/portable way to iterate characters in Helm/Sprig across versions; it can behave unexpectedly consider using use a safer parsing approach with regex (regexFind, regexReplaceAll

charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
```diff
@@ -35,24 +35,22 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: ama-logs
+  name: ama-logs{{- if gt $index 0 }}-{{ $size.name }}{{- end }}
```
wanlonghenry, line 42: could you share how can we check the nodeAffinity logic is mutually exclusive per tier

charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
```diff
@@ -647,19 +641,19 @@ spec:
...
           imagePullPolicy: Always
         {{- else }}
           imagePullPolicy: IfNotPresent
         {{- end }}
           resources:
             limits:
               {{- $containerResources := index $size.containers "ama-logs-prometheus" }}
-              cpu: {{ $containerResources.cpuLimit }}
-              memory: {{ $containerResources.memoryLimit }}
+              cpu: {{ include "maxResourceValue" (list $.Values.OmsAgent.omsAgentPrometheusSidecarCPULimit $containerResources.cpuLimit) }}
```
wanlonghenry, line 659: for test we would want to make sure that both cpu and memory be covered with different set of values; such as cpu with 750m, 1, 6   memory with 750mi, 1Gi 6Gi

charts/azuremonitor-containerinsights/templates/_helpers.tpl
NicAtMS: this will treat it as 1 core. This behavior mimics what is done for addons right now. From one of the snapshots where we set the value to 3 via the variable set (essentially toggles) we get:

                  resources:
                    limits:
                        cpu: 3
                        memory: 3Gi
                    requests:
                        cpu: 229m
                        memory: 1058Mi

charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
NicAtMS: the way the node affinity is constructed forces mutual exclusivity. we are essentially doing:

maxCPU: { valueOfCurrentTshirtSize }
MinCPU: { valueOfPreviousTshirtSize }

This means that, for example, the large size daemonset will have bounds of:

maxCPU { largeCPUSize }
minCPU { mediumCPUSize + 1}

this ensures that even if the large size is misconfigured for some reason, as in, it's set to 5 and the medium size is also set to 5, the node afinity for large would be anything less than 5 and larger than 6, which is not valid, meaning it would essentially be omitted. There is no way for these values to conflict.

charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
NicAtMS: in my testing with fixtures, I tested the following values:

CPU: 500m, 2, 3
Memory: 1Gi, 2Gi, 3Gi

Please let me know if these values are acceptable tests

charts/azuremonitor-containerinsights/templates/_helpers.tpl
NicAtMS: updated to use regexFind

PR #1635, opened 2026-04-06
add test clusters

zanejohnson-azure:


rashmichandrashekar (dismissed):

.pipelines/azure_pipeline_mergedbranches.yaml
suyadav1, line 945: when are we planning to remove these test clusters?

zanejohnson-azure: remove after we are done with migration of the existing clusters from test pipeline to build pipeline. I will need my own clusters as reference when there are issues during migration.

rashmichandrashekar (dismissed):

suyadav1 (approved):

PR #1636, opened 2026-04-08
AMCS change in extension chart

suyadav1:


charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
```diff
@@ -425,6 +425,10 @@ spec:
             - name: CLUSTER_CLOUD_ENVIRONMENT
               value: "{{ $.Values.global.commonGlobals.CloudEnvironment | lower }}"
 {{- end }}
+            - name: AMCS_CLIENT_INSTALL_ID_OVERRIDE
```
wanlonghenry, line 428: I think this also needed to be added to ama-logs-deployment-multitenancy.yaml

suyadav1: Added

PR #1653, opened 2026-04-21
Support for delos 

rashmichandrashekar:
- Delos support for logs addon


charts/azuremonitor-containerinsights-for-prod-clusters/templates/ama-logs.yaml
```diff
@@ -15,6 +15,7 @@
 {{- $aksclustername := last $resourceParts -}}
 {{- $aksResourceGroup := index $resourceParts 4 -}}
 {{- $region := .Values.global.commonGlobals.Region -}}
+{{- $cloudEnv := lower $.Values.global.commonGlobals.CloudEnvironment }}
```
zanejohnson-azure, line 18: the helm chart in this azuremonitor-containerinsights-for-prod-clusters is only used from backdoor deployment for two model clusters in release pipeline and test clusters in build pipeline, and they will not be in production. this folder will be gone when merged chart is in ci_prod branch. that being said, any helm chart need  to be in long's merged chart branch, which is the ground of truth.

rashmichandrashekar: yes, we need to come up with a better way to do this.. For everyone to do this is not ideal.. ideally it should have been in the same folder as in main so that we can simply do a rebase from main to that branch

rashmichandrashekar: i think a better way to do this is to keep a copy for extension's current state in the main branch as well so that we can make all changes going in to both folders, that way it will be easier to merge back from main to Long's branch

rashmichandrashekar: i have added that folder here as well so that we can make all the future changes here and merge back

kubernetes/linux/main.sh
```diff
@@ -154,6 +157,13 @@ setCloudSpecificApplicationInsightsConfig() {
             echo "export APPLICATIONINSIGHTS_ENDPOINT=$APPLICATIONINSIGHTS_ENDPOINT" >>~/.bashrc
             source ~/.bashrc
             ;;
+      #    "azuredeloscloud")
```
zanejohnson-azure, line 160: when will this be commented out? should we add a "TODO:" comment so this will not be forgotten.

test/unit-tests/test_cases/Test-GetClusterCloudEnvironment.ps1
```diff
@@ -11,7 +11,8 @@ function Test-ValidEnvironmentVariables {
         "azureusgovernmentcloud",
         "usnat",
         "ussec",
-        "azurebleucloud"
```
zanejohnson-azure, line 14: unrelated to this PR: this unit-tests never run at all, right?

kubernetes/linux/main.sh
rashmichandrashekar: this will be uncommented when AI is supported in delos, we are tracking this for delos completion, and will not be forgotten

test/unit-tests/test_cases/Test-GetClusterCloudEnvironment.ps1
rashmichandrashekar: I dont know :) may be @suyadav1 can answer

zanejohnson-azure (approved):

PR #1662, opened 2026-04-24
Zane/fix fluentd procstat pattern

zanejohnson-azure:
# problem 1
both ruby and fluentd has the keyword "ruby" and "fluentd" in its cmdline, so using pattern can't tell who is who. also using exe as filter does not work either, because fluentd is a gem under ruby, so the exe of fluentd process is also ruby. 

fix: use procstat.filter doc: https://docs.influxdata.com/telegraf/v1/input-plugins/procstat/#:~:text=This%20plugin%20allows%20to%20monitor,service%20that%20started%20the%20process.

tested, now values of fluentd and ruby matches what is showing inside container

image

manual check
fluentd (supervisor): 47.9 MB
ruby: 82.5 MB 

# problem 2:
main telegraf and the process metrics collection telegraf both use telegraf.exe so exe can be differentiate whether the telegraf process is main or process-metrics. 

fix: The only difference is the telegraf config files, so we can use pattern to differentiate between them.

# problem 3:
add a ProcessName field to make query easier.

# test
image

### DIFF
--- a/build/linux/installer/conf/telegraf-ama-logs-process-metrics.conf
+++ b/build/linux/installer/conf/telegraf-ama-logs-process-metrics.conf
@@ -109,7 +109,7 @@
 
 [[inputs.procstat]]
   name_prefix = "t.azm.ms/"
-  pattern = "fluentd(?!.*under-supervisor)"
+  pattern = "ruby /usr/bin/fluentd"
   interval = "60s"
   pid_finder = "native"
   pid_tag = true
--- a/build/linux/installer/conf/telegraf-ama-logs-process-metrics.conf
+++ b/build/linux/installer/conf/telegraf-ama-logs-process-metrics.conf
@@ -31,6 +31,7 @@
     AgentVersion = "$AGENT_VERSION"
     ControllerType = "$CONTROLLER_TYPE"
     AksResourceId = "$AKS_RESOURCE_ID"
+    ProcessName = "mdsd"
 
 [[inputs.procstat]]
   name_prefix = "t.azm.ms/"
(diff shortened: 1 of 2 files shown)


build/linux/installer/conf/telegraf-ama-logs-process-metrics.conf
```diff
@@ -109,7 +111,6 @@
 
 [[inputs.procstat]]
   name_prefix = "t.azm.ms/"
-  pattern = "fluentd"
```
suyadav1, line 112: we can also use "--under-supervisor" pattern to figure out ruby vs fluentd. Though both these processes are ruby/fluentd itself. One is supervisor process and the other is the worker process.

zanejohnson-azure: yes, that is another way. my 1st commit is using that approach https://github.com/microsoft/Docker-Provider/pull/1662/changes/98996df68dcce9671fe4abd078aaab93733de9f8.

current way is easier to maintain and read i believe.

suyadav1: The problem with this is that now you have to write different queries for different processes i.e. for mdsd, you use customDimensions.exe but for fluentd, you would use customDimensions.filter. Can we push the process names into a single column and use that to query?

zanejohnson-azure: added a new field to make query easier. also include telegraf changes in this PR so all changes are here.

suyadav1 (approved):

PR #1665, opened 2026-05-01
chore: Add automated Go version upgrade workflow

suyadav1:
- Add GitHub Actions workflow (.github/workflows/go-upgrade.yml) that runs daily to detect new Go versions from MCR, update go.mod files and dependencies, and open a PR
- Add env/go-version.txt as the single source of truth for GOLANG_BASE_IMAGE
- Update ADO pipeline to read GOLANG_BASE_IMAGE from env/go-version.txt

### DIFF
--- a/.pipelines/azure_pipeline_mergedbranches.yaml
+++ b/.pipelines/azure_pipeline_mergedbranches.yaml
@@ -12,6 +12,7 @@ variables:
   subscription: '9b96ebbd-c57a-42d1-bbe9-b69296e4c7fb'
   containerRegistry: 'containerinsightsprod'
   repoImageName: '${{ variables.containerRegistry }}.azurecr.io/public/azuremonitor/containerinsights/cidev'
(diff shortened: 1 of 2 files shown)


.pipelines/azure_pipeline_mergedbranches.yaml
```diff
@@ -162,6 +162,15 @@ extends:
             targetPath: '$(Build.ArtifactStagingDirectory)'
             artifactName: linux-drop
         steps:
+        - bash: |
+            # Read Go version from checked-in version file and set as pipeline variable
```
rashmichandrashekar: @suyadav1 - does this cause any issues when telegraf is in a different version but our pipeline is in a different version, will that be fine since we dont build tgf ourselves?

suyadav1: Yes, that should be fine. The approach remains the same, earlier we were reading directly from the pipeline variable manually updated by us whereas this will read from a file updated by this automation and set it as the pipeline variable.

rashmichandrashekar (approved):

PR #1212, opened 2024-03-21
move LOGS_AND_EVENTS_ONLY telemetry to DS

pfrcks:
This pull request includes changes primarily aimed at improving telemetry and data collection. The changes involve the addition of a new environment variable `LOGS_AND_EVENTS_ONLY` in the `telemetry.go` file and the removal of the same variable from the `in_kube_podinventory.rb` file.

Key changes include:

Telemetry improvements:

* [`source/plugins/go/src/telemetry.go`](diffhunk://#diff-89812be185b3a7815ab0ee0f75c129ac501d03e6dd56009344ad4db5d88b73a6R253-R256): Added a new environment variable `LOGS_AND_EVENTS_ONLY` to the `SendContainerLogPluginMetrics` function. This variable is used to add a new dimension to the telemetry data if it is set.

Data Collection Profile changes:

* [`source/plugins/ruby/in_kube_podinventory.rb`](diffhunk://#diff-601d8e1229fe058c89f6416233d1630f63c6f7f88ff949ad31dd8d3538475eb6L253-L255): Removed the environment variable `LOGS_AND_EVENTS_ONLY` from the `enumerate` function. This variable was previously used to set the `logsOrEventsOnlyProfile` property in the telemetry data when the `dataCollectionProfile` is being set.

These changes seem to be aimed at refining the data collection and telemetry reporting mechanisms by moving the `LOGS_AND_EVENTS_ONLY` environment variable from one part of the codebase to another.

PR #1672, opened 2026-05-06
Emit multiline telemetry from DaemonSet heartbeat

suyadav1:
## Problem

Multiline configuration telemetry (`multilineEnabled`, `multilineLanguages`) is silently dropped when the `LOGS_AND_EVENTS_ONLY` profile is enabled, so we have no visibility into multiline usage for clusters running that profile.

## Root cause

The multiline dimensions are emitted from two places today, and neither runs in `LOGS_AND_EVENTS_ONLY` mode:

| Emitter | Plugin / channel | Why it doesn't fire in LOGS_AND_EVENTS_ONLY |
| --- | --- | --- |
| `source/plugins/ruby/CAdvisorMetricsAPIClient.rb` (lines 306-311) | fluentd cAdvisor heartbeat → `customMetrics` | fluentd is skipped entirely in this profile (`kubernetes/linux/main.sh` ~lines 1202-1211) |
| `source/plugins/go/input/lib/cadvisor.go` (lines 1068-1070) | fluent-bit `perf` input → `customEvents` (cAdvisor heartbeat) | only loaded when `AZMON_RESOURCE_OPTIMIZATION_ENABLED=true` (gated in `build/common/installer/scripts/fluent-bit-conf-customizer.rb` line 75) |

The DaemonSet heartbeat (`ContainerLogDaemonSetHeartbeatEvent`) emitted from `source/plugins/go/src/telemetry.go` is the one telemetry path that always runs in the DS, including under `LOGS_AND_EVENTS_ONLY`. It already carries the `logsAndEventsOnly` dimension but did not include the multiline fields.

## Fix

Add `multilineEnabled` and `multilineLanguages` to the DaemonSet heartbeat in `telemetry.go`, right next to the existing `logsAndEventsOnly` dimension.

```go
azmonMultilineEnabled := os.Getenv("AZMON_MULTILINE_ENABLED")
if azmonMultilineEnabled != "" {
    telemetryDimensions["multilineEnabled"] = azmonMultilineEnabled
    azmonMultilineLanguages := os.Getenv("AZMON_MULTILINE_LANGUAGES")
    if azmonMultilineLanguages != "" {
        telemetryDimensions["multilineLanguages"] = azmonMultilineLanguages
    }
}
```

## Precedent

Commit `aa3098b14` (PR #1212, "move LOGS_AND_EVENTS_ONLY telemetry to DS") used the same approach to surface the `logsAndEventsOnly` dimension. The duplication between the cAdvisor heartbeat and the DS heartbeat also matches the existing `nflEnabled` pattern (emitted in both `cadvisor.go` and `telemetry.go`).

## Why we keep the existing emit sites

- They publish to a different App Insights table (`customMetrics` via the cAdvisor heartbeat vs `customEvents` via the DS heartbeat), so they do not double-count.
- The Ruby and Go cAdvisor sites are mutually exclusive (one runs with fluentd, the other only when resource optimization is enabled).
- Removing them would be a behavior change for the non-`LOGS_AND_EVENTS_ONLY` paths and is out of scope for this fix.

## Validation

- `go build ./...` from `source/plugins/go/src` is clean.
- No behavior change for non-`LOGS_AND_EVENTS_ONLY` modes; the multiline dimensions will simply also be present on the DS heartbeat.

## Follow-up (not in this PR)

A broader audit identified other log-related fields with the same gap in `LOGS_AND_EVENTS_ONLY` (e.g. `clusterstdoutlogs`, `clusterstderrlogs`, `containerLogsRoute`, `metadataEnabled`, etc.). Happy to fix those in a separate PR if desired.


source/plugins/go/src/telemetry.go
```diff
@@ -295,6 +295,15 @@ func SendContainerLogPluginMetrics(telemetryPushIntervalProperty string) {
 					telemetryDimensions["logsAndEventsOnly"] = logsAndEventsOnly
 				}
 
+				azmonMultilineEnabled := os.Getenv("AZMON_MULTILINE_ENABLED")
+				if azmonMultilineEnabled != "" {
+					telemetryDimensions["multilineEnabled"] = azmonMultilineEnabled
+					azmonMultilineLanguages := os.Getenv("AZMON_MULTILINE_LANGUAGES")
+					if azmonMultilineLanguages != "" {
```
rashmichandrashekar, line 302: can this be special characters and does ingestion work for all characters? any validation needeD?

suyadav1: we already have validation in place to accept only the allowed languages: https://github.com/microsoft/Docker-Provider/blob/74ec72e9c14058af7b0dce5a24067c567ac4b23a/build/common/installer/scripts/tomlparser.rb#L379-L382

rashmichandrashekar (approved):

PR #1677, opened 2026-05-11
add e2e tests to all nodes clusters

zanejohnson-azure:
tested using ginko on this cluster.

### DIFF
--- a/test/scenario/log-gen-all-nodes-daemonset.yaml
+++ b/test/scenario/log-gen-all-nodes-daemonset.yaml
@@ -81,3 +81,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        resources:
+          requests:
+            cpu: 50m
+            memory: 256Mi
+          limits:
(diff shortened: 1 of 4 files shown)


test/scenario/log-gen-all-nodes-daemonset.yaml
```diff
@@ -0,0 +1,83 @@
...
+    app: log-gen
+    os: windows
+spec:
+  selector:
+    matchLabels:
+      app: log-gen
+      os: windows
+  template:
+    metadata:
+      labels:
```
rashmichandrashekar, line 64: pls set resource req and limits

rashmichandrashekar (dismissed):

test/scenario/log-gen-all-nodes-daemonset.yaml
zanejohnson-azure: good catch. added.

rashmichandrashekar (approved):

PR #1678, opened 2026-05-11
remove deployment verification

zanejohnson-azure:
this verification introduces more noise than benefits due to flaky image checking during helm chart rollout.
ama-metrics does not do deployment verification and works fine. so remove it from ama-logs too.


suyadav1:
I like this validation in place to make sure that the image deployed is the expected one. Instead of getting the image from kubectl get pods, can we read it from the DS/deployment spec which shouldn't be flaky?

zanejohnson-azure:
> I like this validation in place to make sure that the image deployed is the expected one. Instead of getting the image from kubectl get pods, can we read it from the DS/deployment spec which shouldn't be flaky?

helm chart crud will fail if deployment, so adding deployment spec check is unnecessary.

suyadav1 (approved):

PR #1711, opened 2026-06-08
OTLP GRPC support

suyadav1:
This pull request updates the configuration for the AMACoreAgent's OpenTelemetry logging and upgrades the `azure-mdsd` package version in the Linux Kubernetes setup. The most important changes are grouped below:

OpenTelemetry configuration improvements:

* Replaces legacy `PA_AMCS_*` environment variables with new `OTLP_HTTP_PROTOBUF_LOGS_TRACES_*` and `OTLP_GRPC_LOGS_TRACES_*` variables for both host and port, providing clearer separation between HTTP/Protobuf and gRPC endpoints.
* Adds support for overriding both HTTP/Protobuf and gRPC ports via `AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT` and `AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT_GRPC` environment variables.
* Updates `.bashrc` to persist the new OTLP environment variables, and disables OTLP ports by setting them to `-1` when OpenTelemetry logs are not enabled.

Dependency update:

* Upgrades the `azure-mdsd` package from version `1.40.3` to `1.42.0` in the `setup.sh` script to ensure the agent uses theversion which supports both http and grpc protocols in parallel.

Resource usage:

image

image

S.N. L2R | Image | Load | CPU usage | Mem usage
-- | -- | -- | -- | --
1 | ciprod | 50k EPS HTTP only | 0.85 core | 458 MiB
2 | cidev | 25k EPS HTTP + 25k EPS GRPC | 1 core | 512 MiB
3 | cidev | 50k EPS HTTP only | 0.87 core | 462 MiB
4 | cidev | 50k EPS GRPC only | 0.95 core | 500 MiB


kubernetes/linux/main.sh
```diff
@@ -84,22 +84,40 @@ startAMACoreAgent() {
          echo "export AMACALogFilePath=$AMACALogFilePath"
       } >> ~/.bashrc
 
-      if isOpenTelemetryLogsEnabled; then
-            export PA_AMCS_PROTOCOL="HttpProtobuf"
```
rashmichandrashekar, line 88: @suyadav1 - as a part of this we are validating both HTTP and GRPC right?

suyadav1: yes, tested individually and both in parallel.

kubernetes/linux/main.sh
```diff
@@ -84,22 +84,40 @@ startAMACoreAgent() {
...
-            export PA_AMCS_HOST="0.0.0.0"
-            export PA_AMCS_PORT=4319
+      if isOpenTelemetryLogsEnabled; then         
+            export OTLP_HTTP_PROTOBUF_LOGS_TRACES_HOST="0.0.0.0"
+            export OTLP_GRPC_LOGS_TRACES_HOST="0.0.0.0"   
+            export OTLP_HTTP_PROTOBUF_LOGS_TRACES_PORT=4319
+            export OTLP_GRPC_LOGS_TRACES_PORT=4320
             if [ -n "${AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT}" ]; then
```
zanejohnson-azure, line 92: some comments about data validation and error handling.

1. AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT: this should be validated before  using it.

2. if [ "$port_int" -gt 0 ]
we need handle the false case. 

do 1 will eliminate if clause in 2.

suyadav1: AZMON_OPENTELEMETRYLOGS_CONTAINER_PORT will only be used when it's set and > 0. Any other validation needed? 
If the port_int <= 0, we won't use it and go with the default value.

zanejohnson-azure (approved):

PR #1717, opened 2026-06-17
Recover ama-logs workspace key from extension protected secret + checksum restart

jhmadhav:
## Summary

Recover the Container Insights workspace key from the extension manager's
protected-parameters secret when it is not re-delivered to the chart, and roll
the agent pods (Linux **and** Windows) when the effective secret changes.

This prevents a class of failures where `ama-logs-secret` gets overwritten with
an empty workspace `KEY`, which silently breaks log ingestion ~10-14 days later
when the previously-cached credentials expire.

## Background / root cause

For workspace-key (non-AAD) clusters, the workspace key is delivered to the
chart as a **protected parameter**. The config agent persists it on-cluster in
`protected-ext-parameters-<release>` and passes it to Helm as
`OmsAgent.workspaceKey` (or `amalogs.secret.key` for Arc).

During an extension auto-update the extension manager can stop re-delivering the
protected parameters. When that happens, `.Values.OmsAgent.workspaceKey` renders
**empty**, and the chart overwrites the live `ama-logs-secret` `KEY` with an
empty value. The agent keeps working until its cached credentials expire, then
fails — making the root cause hard to correlate with the triggering update.

Importantly, the agent does **not** clear `protected-ext-parameters-<release>`
when it drops the CR reference, so that secret remains a valid source of truth
for the key.

### 1. Conditional workspace-key fallback (`ama-logs-secret.yaml`)
When the incoming workspace key is empty/placeholder, fall back to the existing
on-cluster `protected-ext-parameters-<release>` secret (data key
`OmsAgent.workspaceKey` for AKS, `amalogs.secret.key` for Arc) via Helm `lookup`.

Semantics (intentionally **conditional**, not unconditional):
- Incoming **real** key always wins → key rotation is preserved.
- Incoming **empty** key → use whatever is in the existing secret.

Only the **KEY** is recovered. The workspace **ID** (WSID) is a *non-protected*
parameter that continues to be delivered, so it is taken from values as-is.

### 2. `checksum/secret` on the AKS pod templates (Linux + Windows)
The Arc path already had `checksum/secret`; the AKS (non-Arc) path only had a
`WSID` annotation, which does **not** change when the workspace **KEY** changes.
Added `checksum/secret` to all three AKS workloads — `ama-logs-daemonset.yaml`,
`ama-logs-daemonset-windows.yaml`, and `ama-logs-deployment.yaml` — so pods roll
when the effective secret actually changes.

## Important nuance: AAD/MSI clusters

For `OmsAgent.isUsingAADAuth: "true"` clusters the workspace key is **empty by
design** (auth uses a managed-identity token, not a workspace key). The fix does
**not** special-case this and does **not** need to:

- On an AAD cluster the `protected-ext-parameters-<release>` secret is itself
  empty, so the fallback resolves to empty — exactly the value we want sent to
  the chart. No guard on `isUsingAADAuth` is required.

So the KEY-fallback is a **no-op for AAD clusters** and a **fix for legacy
workspace-key clusters that dropped protected settings**. The `checksum/secret`
change benefits both.

## Behavior matrix

| Cluster | Incoming key | Existing protected secret | Rendered KEY | Correct |
|---|---|---|---|---|
| AAD auth (healthy) | empty (intended) | empty | empty | ✅ |
| Key auth (healthy) | real key | real key | real key (incoming wins) | ✅ |
| Key auth (dropped protected params) | empty | retained key | recovered key | ✅ |
| Key auth (rotation) | new key | old key | new key (incoming wins) | ✅ |

## Testing

Validated against a **live AKS cluster running the real azure-monitor-logs
extension** (chart 3.3.0, `isUsingAADAuth: true`) using
`helm template --dry-run=server` (executes `lookup` server-side):

- Empty incoming key + key present in `protected-ext-parameters-*` → **recovered** ✅
- Real incoming key → **takes precedence** (rotation safe) ✅
- Current published rendering (no server lookup) + empty key → `KEY: ""`
  (reproduces the bug) ✅
- `helm lint` passes; Arc empty-key path still fails loudly via `required`.

## Notes

- `lookup` returns empty during `helm template`/client-side dry-run/first
  install; on a real `helm upgrade` (extension manager flow) it returns the live
  secret. On first install the incoming key is populated, so the fallback is a
  no-op there.
- `lookup` requires `get` on Secrets in the release namespace; the extension's
  Helm operator already has this.

### DIFF
--- a/charts/azuremonitor-containerinsights/templates/ama-logs-daemonset-windows.yaml
+++ b/charts/azuremonitor-containerinsights/templates/ama-logs-daemonset-windows.yaml
@@ -78,6 +78,7 @@ spec:
         agentVersion: "47.7.1"
         dockerProviderVersion: "18.0.1-0"
         schema-versions: "v1"
+        checksum/secret: {{ include (print $.Template.BasePath "/ama-logs-secret.yaml") . | sha256sum }}
         WSID: {{ .Values.OmsAgent.workspaceID | b64enc | quote }}
         kubernetes.azure.com/no-http-proxy-vars: "true"
 {{- end }}
--- a/charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
+++ b/charts/azuremonitor-containerinsights/templates/ama-logs-daemonset.yaml
@@ -108,6 +108,7 @@ spec:
         agentVersion: "azure-mdsd-1.40.3"
         dockerProviderVersion: "18.0.1-0"
         schema-versions: "v1"
+        checksum/secret: {{ include (print $.Template.BasePath "/ama-logs-secret.yaml") $ | sha256sum }}
         WSID: {{ $.Values.OmsAgent.workspaceID | b64enc | quote }}
         kubernetes.azure.com/no-http-proxy-vars: "true"
 {{- end }}
--- a/charts/azuremonitor-containerinsights/templates/ama-logs-deployment.yaml
+++ b/charts/azuremonitor-containerinsights/templates/ama-logs-deployment.yaml
@@ -74,6 +74,7 @@ spec:
         cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
         dockerProviderVersion: "18.0.1-0"
         schema-versions: "v1"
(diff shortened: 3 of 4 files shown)


suyadav1:
Closing this since the change is already pushed in another PR.

PR #1765, opened 2026-08-18
Fix extension chart for FF & MC

suyadav1:
This pull request updates the way cloud environment selection is handled for domain configuration in the Azure Monitor Container Insights Helm chart. Instead of using boolean flags for specific environments, it now uses a single variable to determine the cloud environment, simplifying configuration and reducing potential errors.

**Cloud environment configuration improvements:**

* Replaced the `isMoonCake` and `isFairfax` boolean flags in `values-template.yaml` with a single `$cloudEnv` variable to determine the cloud environment.
* Updated conditional logic in `ama-logs-secret.yaml` to use `$cloudEnv` for selecting the correct domain for China (`azurechinacloud`) and US Government (`azureusgovernmentcloud`) environments, instead of the removed boolean flags.


charts/azuremonitor-containerinsights/values-template.yaml
```diff
@@ -34,9 +34,6 @@ OmsAgent:
   isUsingAADAuth: "true"
   workspaceID: "<your_workspace_id>"
   accessTokenSecretName: "ama-logs-secret"
-  # Cloud environment
-  isMoonCake: false
```
rashmichandrashekar, line 38: we are sure these are not used anywhere right?

suyadav1: yes, it's not used anywhere else in the repo.

rashmichandrashekar (approved):

PR #1785, opened 2026-09-17
Upgrade Windows Telegraf to 1.40.0

zanejohnson-azure:
## Purpose

Upgrade the **official Windows Telegraf package from 1.24.2 to 1.40.0** for [discovery-token file rereading on a failed-poll retry](https://github.com/influxdata/telegraf/blob/e9017dc3266369d6fa185e0e130af1d1d4021ce9/plugins/inputs/prometheus/kubernetes.go#L94-L109). No Telegraf source patch or new production dependency.

## Why the remaining changes are needed

| File / area | Reason and upstream reference |
|---|---|
| `kubernetes/windows/setup.ps1` | Download the official 1.40.0 archive; pin its SHA256 and fail on installation errors. |
| `build/windows/installer/conf/telegraf.conf` | Preserve the **overall 15-second scrape timeout**. [1.24.2 assigned `response_timeout` to `http.Client.Timeout`](https://github.com/influxdata/telegraf/blob/9550e7a533dd00632e14435e87ed3eb4b04832c6/plugins/inputs/prometheus/prometheus.go#L254-L261); [1.40 documentation says it now covers headers only](https://github.com/influxdata/telegraf/blob/e9017dc3266369d6fa185e0e130af1d1d4021ce9/plugins/inputs/prometheus/README.md#L152-L157). Therefore use `timeout = "15s"`. |
| `build/common/installer/scripts/tomlparser-prom-customconfig.rb` | Remove the old Windows-only `response_timeout` override for generated namespace plugins. Linux already used `timeout`; **Linux output, filter names/values and public ConfigMap keys are unchanged**. |
| `build/windows/installer/conf/telegraf-ama-logs-process-metrics.conf` | The optional service shares the upgraded executable. Replace removed `pid_tag` with the [documented `tag_with = ["pid"]`](https://github.com/influxdata/telegraf/blob/e9017dc3266369d6fa185e0e130af1d1d4021ce9/plugins/inputs/procstat/README.md#L55-L76). [1.40's option schema](https://github.com/influxdata/telegraf/blob/e9017dc3266369d6fa185e0e130af1d1d4021ce9/plugins/inputs/procstat/procstat.go#L32-L50) no longer has `pid_tag`. CPU/RSS filters, intervals and output stay unchanged. |
| `main.ps1` + `telegraf-windows-service.rb` | Stock 1.40 [uses native service detection](https://github.com/influxdata/telegraf/blob/e9017dc3266369d6fa185e0e130af1d1d4021ce9/cmd/telegraf/telegraf_windows.go#L85-L94) that [requires a `services.exe` parent in session 0](https://github.com/golang/sys/blob/v0.47.0/windows/svc/security.go#L69-L120); the observed container uses session 5 and failed SCM startup. Reuse the bundled Ruby service dispatcher to run the unchanged executable with `--console`, handling both service roles, child exit and bounded logging/shutdown. |
| `telegraf-windows-console.rb` | Provide native stop signaling and [Windows job containment](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects). Enroll the host before spawning so children cannot escape during startup. This is a helper library, not another service. |
| Tests, runners, docs and DevSkim comments | Cover package/config preservation and service lifecycle; document these compatibility reasons and narrowly suppress public-checksum/test-loopback false positives. |

## Latest six-hour results

**Commit `df6e8d1e`, [build 126849](https://github-private.visualstudio.com/microsoft/_build/results?buildId=126849):** WS2022 observation on September 22, 00:03-06:03 UTC, with **Prometheus and process-metrics collection both enabled**.

- **Process metrics working:** both services started normally. Application Insights received **360 CPU + 360 RSS measurements each** for both Telegraf children and MonAgentCore, with **72/72 five-minute bins populated per series**.
- **Data delivery working:** **ContainerLogV2: 21,413 rows; InsightsMetrics: 216,000 rows**, each with **72/72 nonempty bins**. No sustained delivery gap was observed.
- **No earlier rapid handle-growth pattern observed:** Prometheus Telegraf **263 -> 282**; process-metrics Telegraf **449 -> 403** (range **401-494**). All recorded process identities were unchanged.


suyadav1:
Did we test the fix for WS2022 and WS2025 both? We should merge https://github.com/microsoft/Docker-Provider/pull/1782 and verify there are no issues for ws2025?

suyadav1:
Can we add a check to kill Telegraf / restart the container if the number of handlers are increasing so it doesn't impact other processes?

zanejohnson-azure:
> Can we add a check to kill Telegraf / restart the container if the number of handlers are increasing so it doesn't impact other processes?

i would not prefer to swallow it when it happens. if we hit handle increasing issue in the future, i would prefer to let it fail and then we need do the actual fix.

zanejohnson-azure:
> Did we test the fix for WS2022 and WS2025 both? We should merge #1782 and verify there are no issues for ws2025?

synced offline, i will do the win25 fix after this PR is merged. will rebase win25 fix PR and validate then.

suyadav1 (approved):

